PR #1, opened 2021-05-19
Add split selection

craffel:
Not sure if this is the right way to do it, but it appears to work.


stephenbach:
Awesome, thanks!

PR #2, opened 2021-05-19
Dataset configs / Jinja Tpls / Feature schemas

srush:
This change merges some new features to the prompt viewer. This is what it currently looks like: 

image

Main changes are: 

* Add dataset caching and dataset configuration for models like GLUE
* Adds splits options  
* Mode to allow users to use Jinja templates for very quick template writing.  (and to make it easier to see which are written already)
* Display the Datasets feature schema to see the class types defined by the model
* Display the description of the dataset. 
* Some aesthetic changes to the sidebar


stephenbach:
Beautiful!

PR #8, opened 2021-05-20
update session_example_index after change in split selection

shanyas10:
the index was not getting updated after split and resulted in an exception in cases where it was greater than the length of dataset


stephenbach:
Thanks for the PR! I think we need something a little more complicated to fix this issue. Right now, session_state is used to remember what example we're looking at, so that if someone saves an edit to a template, it doesn't change the example they're working on. This PR just gets around the issue by regenerating the random index on each page load.

So maybe it's better to add a slider or something to select the example and get rid of session_state altogether?

stephenbach:
Thank you!

Issue #7, opened 2021-05-20
basic quality code checks


VictorSanh:
It would be nice to set up some code quality checks (black, isort for instance) to ensure consistency in coding quality & style

PR #11, opened 2021-05-20
build: add Makefile and set dependencies for coding style

tianjianjiang:
close: bigscience-workshop/promptsource#7

# What changed
- Add a Makefile that imitates the procedures of HuggingFace repositories
- Change Python version to 3.8 for Apple Silicon
- Update requirements.txt based on what Poetry suggests, which is compatible with the PR bigscience-workshop/promptsource#5

# Notes
- I use [Angular's commit message format](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit)
- Not sure if we want to have pre-commit hook, Poetry, GitHub Action, etc.


stephenbach:
Thanks Mike!

Do you think it's necessary to hardcode the exact dependencies in requirements.txt? We now have `pyyaml>=5` and `streamlit>=0.82.0`. I've run into issues where being too strict with versions leads to unresolvable constraints among transitive dependencies.

tianjianjiang:
> Thanks Mike!
> 
> Do you think it's necessary to hardcode the exact dependencies in requirements.txt? We now have `pyyaml>=5` and `streamlit>=0.82.0`. I've run into issues where being too strict with versions leads to unresolvable constraints among transitive dependencies.

@stephenbach Hey, you're welcome and sorry for the late reply.
I will relax the conditions of the version numbers soon.

In the meantime, I'm pretty sure that you are fully aware of what I'm about to say, just in case:
- We probably don't need exact version numbers for requirements.txt, those were just for reproducibility, I think.
    - In my experience, unresolvable constraints happened when certain packages had inappropriate dependencies (in their own setup.py or requirements.txt) that we have no control over them. Sometimes inexact version constrains led to conflicts, too. Other than that, `pip-tools` (hence Poetry) almost always does a good job.
- If we don't want to enforce dependency and/or virtual environment management tools, it might be nice to have a full requirements.txt instead of one having only top-level dependencies. In the former case, version numbers better be exact, for ruling out any element of surprise.

promptsource/promptsource.py
```diff
@@ -157,12 +157,14 @@ def save_data(message="Done!"):
     st.sidebar.subheader("Dataset Schema")
     st.sidebar.write(render_features(dataset.features))
 
-    
-    st.sidebar.subheader("Select Example")
-    example_index = st.sidebar.slider("Select the example index", 0, len(dataset)-1)
-    
-    example = dataset[example_index]
-    st.sidebar.write(example)
+    with st.form("example_form"):
```
stephenbach, line 160: This also got reverted.  It should be a slider now.

promptsource/promptsource.py
```diff
@@ -196,15 +204,14 @@ def save_data(message="Done!"):
...
                 index = 0
-            template_name = st.selectbox('', template_list, key='template_select',
-                                          index=index, help='Select the template to work on.')
+            template_name = st.selectbox(
+                "", template_list, key="template_select", index=index, help="Select the template to work on."
+            )
 
             if st.button("Delete Template", key="delete_template"):
-                templates.remove_template(template_key, template_name)
+                templates.remove_template(template_key, template.get_name())
```
stephenbach, line 212: Looks like this got reverted back on this branch. It should indeed be `template_name` because the object doesn't need to be instantiated.

stephenbach (changes requested):
Thanks! Only issue is that it looks like `promptsource/promptsource.py` isn't the latest version. Maybe can take version on `main` and apply formatting? I'll try to hold other PRs until this one gets done so there aren't any more changes to merge.

promptsource/promptsource.py
tianjianjiang: Oops, I probably chose wrong side when rebase-force-push. Let me try it again. Apologies for the trouble.

tianjianjiang: Seems fixed.

promptsource/promptsource.py
tianjianjiang: Seems fixed.

stephenbach (approved):
Awesome, thank you!

Issue #3, opened 2021-05-19
Surface datasets without prompts


srush:
It would be nice to visually indicate which datasets people should work on next, and which have a lot of prompts already.

PR #15, opened 2021-05-21
Display number of existing templates/(dataset,config) and add checkbox to filter priority datasets

manandey:
PR to display current number of templates created for each dataset/config


stephenbach:
Awesome! I wonder if there's a way to display this info without loading the dataset? So as described in #3 , users could know what they should prioritize.

shanyas10:
I was wondering if it would also be beneficial to sort the dataset_list in the (reversed) order of the number of templates present

manandey:
> Awesome! I wonder if there's a way to display this info without loading the dataset? So as described in #3 , users could know what they should prioritize.

Sure! I will have a look at it.

manandey:
Hi @stephenbach! The current PR is updated to accommodate #3 as well.

shanyas10:
@stephenbach we were wondering if the number of templates should be displayed in the dropdown as well (since the current implementation would give us an idea of the number post-selection only)? Or we can have custom filters for datasets for filtering on basis of the number of existing templates ( in addition to others like language, as suggested in #22  )

templates.yaml
```diff
@@ -4,44 +4,77 @@ ag_news:
...
     name: test1
+    output_fn: ''
+    prompt_fn: ''
     reference: example template_1
   basic: !Template
     jinja: "{{text}} \nIs this text an example of news about world politics, sports,\
       \ business, or technology? ||| \n{{[\"World politics\", \"Sport\", \"Business\"\
       , \"Technology\"][label] }}"
     name: basic
+    output_fn: ''
```
stephenbach, line 15: Looks like this reverted to an old version of `templates.yaml`. There shouldn't be any *_fn fields anymore.

promptsource/templates.py
```diff
@@ -67,7 +69,7 @@ def remove_template(self, dataset, template_name):
             raise ValueError(f"No template with name {template_name} " + f"for dataset {dataset} exists.")
 
         del self.templates[dataset][template_name]
-
+        self.templates[dataset]["count"] = num_templates
```
stephenbach, line 72: Is it necessary to maintain an explicit count of the templates? Would just calling `len(self.templates[dataset])` have the same effect and avoid the headache of having to keep track of the counts?

stephenbach:
Thanks @manandey and @shanyas10 ! Left a couple comments.

Re: displaying number of templates in the dropdown, that would be nice! (Also need to make it work for subsets.)

manandey (commented):
@stephenbach variable names updated in this.

templates.yaml
manandey: The changes have been updated!

stephenbach (approved):
LGTM!

PR #16, opened 2021-05-21
remove max_size restriction

shanyas10:
removing the restriction on max_size (100 MB) of data that can be downloaded


stephenbach:
Thank you!

Issue #13, opened 2021-05-20
 FileNotFoundError while loading a dataset


manandey:
Loading datasets like Salesforce/QAConv, wmt/wmt17, yluisfern/PBU etc. failing due to` FileNotFoundError: Couldn't find file locally at Salesforce/QAConv\QAConv.py, or remotely at https://huggingface.co/datasets/Salesforce/QAConv/resolve/main/QAConv.py. Please provide a valid dataset name`

arnaudstiegler:
Looked into that, it's basically coming from the community datasets. Some of them just don't work. 
We could remove those community datasets  by doing `datasets.list_datasets(with_community_datasets=False)` 

I've seen a bunch of issues with those community datasets, like dummy datasets (like `'smallv0221/my-test`) and duplicate datasets (for instance `'lhoestq/squad'`). Not that familiar with that part of the `datasets` hub so not sure how many of the community datasets are valuable.
@stephenbach WDYT?

stephenbach:
Thanks both! Yes, I think focusing on the datasets without community ones is fine for now.

arnaudstiegler:
Sounds good, will open a PR for that

PR #17, opened 2021-05-21
Removing community datasets from dataset list

arnaudstiegler:
Removing community datasets from the datasets list. Most of them don't work, and there are also duplicates from the non-community datasets. We go from 884 datasets to 663 datasets.


arnaudstiegler:
Fix for #13

stephenbach:
Thank you!

Issue #10, opened 2021-05-20
define data augmentation API | Streamlit → produce variations of prompt and/or input


tianjianjiang:
# Job Story

## When
Using Streamlit for generating templates

## I want to
Have data augmentation API defined
- [ ] Input arguments
    - Instance or template? Template might be tricky for preserving slots.
    - Prompt and/or input?
- [ ] Output type
    - Just one string or a list of strings?

## So I can
Produce variations of prompt and/or input

# Notes
I'm gonna do it during the weekend (2021-05-23/24)

stephenbach:
Closing due to inactivity. Feel free to reopen if you want to revisit this!

PR #19, opened 2021-05-21
feat: draft augmentation API with back-translation strategy #10

tianjianjiang:
close: #10 

As per @VictorSanh's request, this PR provides a draft of back-translation based data augmentation. Its API is too simple to be really useful, such that the PR itself is also a draft.

As the docstring says:

```
This simple and stupid draft is meant to be a pot of stone soup.
Example:
    `augment("Business")` returns `"Enterprises"`
Note:
    The back-translation strategy uses two different models for increasing the diversity.
    The risk of distortion is as the above example shown.
Todo:
    * Speed and memory-footprint
    * Refactoring for a better implementation of Strategy pattern
    * API for batch processing
    * Special tokens that indicate augmentation invariants
```

Any suggestion/patch is welcome!

---
~@tianjianjiang 's first question is: suppose we want template-instance-augmentation being 1-to-many-to-many, shall we have separate yaml files for instances and augmentations?~
Without further ado, this PR only demonstrates what will happen to the prompt being augmented by back-translation. For example:
```
Is this a piece of news regarding world politics, sports, business, or science and technology?
```
becomes
```
Is this news about global politics, sports, business, or science and technology? 
```
Please kindly note that we probably don't want `world politics` being altered as `global politics`. See #49 and #55.

### DIFF
--- a/promptsource/augmentation.py
+++ b/promptsource/augmentation.py
@@ -1,0 +1,38 @@
+"""A Draft of data augmentation for prompt/output instance.
+
+This simple and stupid draft is meant to be a pot of stone soup.
+
+Example:
+    `augment("Business")` returns `"Enterprises"`
+
+Note:
+    The back-translation strategy uses two different models for increasing the diversity.
+    The risk of distortion is as the above example shown.
+
+Todo:
+    * Speed and memory-footprint
+    * Refactoring for a better implementation of Strategy pattern
+    * API for batch processing
+    * Special tokens that indicate augmentation invariants
+"""
+from transformers import AutoModelForSeq2SeqLM, AutoTokenizer, pipeline
+
+
+en_fr_translator = pipeline("translation_en_to_fr")
+fr_en_model_name = "Helsinki-NLP/opus-mt-ROMANCE-en"
+fr_en_translator = pipeline(
+    model=AutoModelForSeq2SeqLM.from_pretrained(fr_en_model_name),
+    tokenizer=AutoTokenizer.from_pretrained(fr_en_model_name),
+    task="translation_fr_to_en",
+)
+
+
+def _back_translation_strategy(text: str) -> str:
+    return fr_en_translator(en_fr_translator(text)[0]["translation_text"])[0]["translation_text"]
+
+
+def _noop_strategy(text: str) -> str:
+    return text
+
+
+augment = _back_translation_strategy
--- a/promptsource/augmentation.py
+++ b/promptsource/augmentation.py
@@ -15,9 +15,14 @@ Todo:
     * API for batch processing
     * Special tokens that indicate augmentation invariants
 """
+import re
+from typing import Iterable, Mapping
+
+from jinja2 import Environment
 from transformers import AutoModelForSeq2SeqLM, AutoTokenizer, pipeline
 
 
+jinja_env = Environment()
 en_fr_translator = pipeline("translation_en_to_fr")
 fr_en_model_name = "Helsinki-NLP/opus-mt-ROMANCE-en"
 fr_en_translator = pipeline(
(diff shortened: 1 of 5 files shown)


promptsource/augmentation.py
```diff
@@ -0,0 +1,38 @@
...
+
+def _back_translation_strategy(text: str) -> str:
+    return fr_en_translator(en_fr_translator(text)[0]["translation_text"])[0]["translation_text"]
+
+
+def _noop_strategy(text: str) -> str:
+    return text
+
+
+augment = _back_translation_strategy
```
tianjianjiang, line 38: I'm debating with myself whether and when to use class-based Strategy pattern.
For now this is likely unnecessary.

promptsource/augmentation.py
```diff
@@ -0,0 +1,38 @@
...
+    task="translation_fr_to_en",
+)
+
+
+def _back_translation_strategy(text: str) -> str:
+    return fr_en_translator(en_fr_translator(text)[0]["translation_text"])[0]["translation_text"]
+
+
+def _noop_strategy(text: str) -> str:
+    return text
```
tianjianjiang, line 35: Next time we may change the interface to
```python
def _some_strategy(
    prompt_template_str: str,
    prompt_parsed_template: jinja2.nodes.Template = None,
    **kwargs
)
```

promptsource/promptsource.py
```diff
@@ -414,7 +415,9 @@ def show_text(t, width=WIDTH):
                         if len(prompt) > 1:
                             st.write("Y")
                             show_text(prompt[1], width=40)
-
+                        st.write("Augmented Prompt (demo)")
+                        aug_subject = template.jinja.split("|||")[0].replace("{{text}}", "").strip()
```
VictorSanh, line 419: the trickiest part will be somehow here:
1/ parsing the jinja template should be easy (as you've seen, we decided to use jinja to escape protected text `{{"this text should not be changed"}}`) 
2/ augment and reinsert the jinja variables. You mentioned you had some strategy to re-insert at the correct place. is that correct?

promptsource/augmentation.py
tianjianjiang: Not doing this for now.

promptsource/promptsource.py
tianjianjiang: My apologies for the late reply and wasn't being clear about the changes.
The previous revision was assuming that #49 and related discussions are not done yet, but now I see #49 is closed so the implementation is updated accordingly.

test/promptsource/test_augmentation.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def test_back_translation_strategy():
+    aug_tmpl = aug.back_translation_strategy(CASE_AG_NEWS_1)
+    assert aug_tmpl == CASE_AG_NEWS_1.replace(
+        "a piece of news regarding",
+        "a new one about"
+    ).replace(
+        ", or",
+        " or"
```
tianjianjiang, line 18: Empirically, the reason why this augmentation is less than ideal ("a new one" and the missing comma) is probably due to the multiple constants and their long replacements (e.g., `x_cnst_1_0`).

promptsource/augmentation.py
```diff
@@ -0,0 +1,86 @@
...
+fr_en_translator = pipeline(
+    model=AutoModelForSeq2SeqLM.from_pretrained(fr_en_model_name),
+    tokenizer=AutoTokenizer.from_pretrained(fr_en_model_name),
+    task="translation_fr_to_en",
+)
+
+CONST_REGEX = re.compile(r"(?P<cnst>{{\"[^\"]*\"}})")
+
+
+def _decorate_constants(tmpl_rows: Iterable[str]) -> (Iterable[str], Mapping[str, str]):
```
tianjianjiang, line 37: I failed to find a way of controlling Jinja constants so this is based on regex...

tianjianjiang:
Probably not going to happen any time soon.

Issue #20, opened 2021-05-23
Example keys with dashes are inaccessible


stephenbach:
Seems to be a quirk of the way we're using Jinja. See, e.g., TREC. A template like
```
{{text}}

Is this asking about a description, an entity, an abbreviation, a person, a quantity, or a location?
|||
{{ ["Description", "Entity", "Abbreviation", "Person", "Quantity", "Location"] [label-coarse] }}
```
doesn't render correctly because `label-coarse` is parsed as `label`. See, e.g., https://stackoverflow.com/questions/28652373/jinja2-in-variable-bug

Can probably just fix by replacing dashes with underscores in keys of the example dictionary when rendering. Might be good to change the displayed information in the sidebar on the fly to match.

Urmish:
I tried pushing a fix to this, but got the following error -
remote: Permission to bigscience-workshop/promptsource.git denied to Urmish

The fix I did was simple. When the template is being saved, I replace all "-" with "_".

![image](https://user-images.githubusercontent.com/9425893/119277829-75047d00-bbe7-11eb-844e-bdc6b20fb6fe.png)

stephenbach:
Thanks for your help with this! I think we might want a different approach, because (at least for me?) this creates a valid Jinja template, but it doesn't actually render the template correctly because there's still a key of "label-coarse" in the dictionary coming from the dataset library. (It renders part of the template, but it doesn't show the answer to the question.) So instead of modifying the templates, I think we need to modify the examples.

Urmish:
Ah! Yes, see if this works?
https://github.com/bigscience-workshop/promptsource/pull/21

Changes both examples and template. I can see the render happening without the label error and the label being displayed -

After the changes -
![image](https://user-images.githubusercontent.com/9425893/119279575-33c59a80-bbf2-11eb-9bbb-15c6af3f015a.png)

Before the changes - 
![image](https://user-images.githubusercontent.com/9425893/119279606-5eafee80-bbf2-11eb-92c3-f1d00b93b98b.png)

PR #21, opened 2021-05-23
Fix for https://github.com/bigscience-workshop/promptsource/issues/20

Urmish:
1. Changes the template.jinja before its saved
2. Changes the keys in examples after its loaded


stephenbach:
Great, thanks!

A couple of thoughts:

1) We should also update the dataset schema with the same cleaning. Maybe we can make a function and just call it on both dictionaries?

2) I'm leaning towards not doing the replacement in line 242, because it might affect other things besides the variable names. For example, someone might want to use a dash in the text part of their prompt. If we change the schema and example, that should be sufficient.

What do you think?

Urmish:
Creating a utility function and (2) makes sense. Can you expand a bit on the dataset schema part. I am not sure if I follow which dictionary you are referring to.

stephenbach:
Hi,
I'm referring to the thing in the sidebar above "Select Example" that's called "Dataset Schema". It's rendered with the line
```
st.sidebar.write(render_features(dataset.features))
```
Thanks!

Urmish:
I see. Let me check how to do that. A quick way dint work as I am not sure if the dataset.features field is mutable. Running into 
"AttributeError: Can't set attribute"

Urmish:
@stephenbach - Made the requested changes. But that will still not solve the problem as the text we enter for the template -
"{{text}}

```
Is this asking about a description, an entity, an abbreviation, a person, a quantity, or a location?
|||
{{ ["Description", "Entity", "Abbreviation", "Person", "Quantity", "Location"] [label-coarse] }}"

```
Has label-coarse and does lead to the error when the template is applied -

```
File "/Users/urmisht/Documents/bigScience/promptsource/promptsource/promptsource.py", line 294, in <module>
    prompt = template.apply(example)
```

So the replace statement for template.jinja will still be needed to render it correctly

Urmish:
I see. Thanks for the clarification. Then this branch has the required fixes? Do you want to give it a try @stephenbach ?

stephenbach:
Right, then I think it's good to go. I will merge it in, thanks!

stephenbach:
@Urmish looks like it's failing code quality checks. Can you add a commit running
```
make style
make quality
```

Urmish:
@stephenbach - Done

Urmish:
Sorry, I dint realize we had code quality checks setup until you mentioned it :)

stephenbach:
Having some weird issue with black (#39). Will just merge in for now.

Issue #24, opened 2021-05-24
🐛`UnboundLocalError: local variable 'd_to_b' referenced before assignment`


VictorSanh:
When loading a dataset (apparently, datasets with configs), this error pops up:

> UnboundLocalError: local variable 'd_to_b' referenced before assignment
Traceback:
File "/home/hf/promptsource/.venv/lib/python3.7/site-packages/streamlit/script_runner.py", line 338, in _run_script
    exec(code, module.__dict__)
File "/home/hf/promptsource/promptsource/promptsource.py", line 252, in <module>
    state.sync()
File "promptsource/session.py", line 70, in sync
    self._state["hash"] = d_to_b

PR #25, opened 2021-05-24
Update session.py (fix for :  UnboundLocalError: local variable 'd_to_b' referenced before assignment)

shanyas10:
fix for #24


stephenbach:
Thanks!

PR #30, opened 2021-05-26
Retain priority dataset in list and add max templates field

### LINKED ISSUE #27: Priority dataset disappears from dataset list after adding a third template
1. Select `Filter Priority Dataset`
2. Select a dataset with 2 templates (e.g.: ag_news)
3. Type in a new template name, press `Create`
4. The form refreshes and the current dataset no longer appears in the template list (because of the `count_dict[d]>2` condition)

One possible solution would be to save the name of the current working dataset and include it in the
templates list in `list_datasets`, something like:
`        dataset_list = list(set(dataset_list) - 
        set(list(d for d in count_dict if count_dict[d]>2 and d!=state.current_dataset_name))) `

When should it be cleared? Maybe as soon as the new template has been saved?

elsanns:
Fixes #27

This PR 
1. Adds a field to configure the maximum number of templates applied with 'Filter Priority Datasets' selected
2. Prevents the current working dataset from getting filtered away after adding new templates (above the configured max). The dataset is kept in the list until the user has changed the dataset or switched the 'Filter Priority Datasets' off.

I ran `make style` and `make quality` so there are more changes to the file than just the fix (hope that's ok).

### DIFF
--- a/promptsource/promptsource.py
+++ b/promptsource/promptsource.py
@@ -1,7 +1,6 @@
 import datasets
 import requests
 import streamlit as st
-
 from session import _get_state
 from templates import Template, TemplateCollection
 from utils import _ADDITIONAL_ENGLISH_DATSETS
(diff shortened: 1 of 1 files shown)


stephenbach:
LGTM! Thanks!

PR #35, opened 2021-05-28
Prompt `movie_rationales` and `xnli/en`

VictorSanh:
Prompting two datasets:
- XNLI (subset en)
- movie_rationales --> good example of how you can create different "skills" by selecting different subsets of the columns. For instance, it can be a classification task or a span extraction task. Also, we can do much more than I initially thought using for loops and if conditions in jinja (I'll add to the contribution guide once we start writing it)


stephenbach:
Ah this is so cool!!

LGTM!

stephenbach (approved):

Issue #29, opened 2021-05-26
Creating unique identifier in the template.yaml


arnaudstiegler:
For now, it looks like we can sort of uniquely identify each template using a combination of template name and dataset name, but I'm expecting potential collisions when a lot of people start contributing. Besides, naming each template might not be useful (like if we end up with names like `template1` `template2` etc...), and it would help contributors if they don't have to add a name/check conflicts on the naming part before merging their template.yaml.

I was thinking that we could add an ID to each entry by getting the hash of timestamp + dataset + string of prompt python function or jinja template? That should be more than enough to prevent collisions

shanyas10:
As an extension, I think we can also keep a user from saving an empty template (currently we can save a template right after entering the name). However, @arnaudstiegler if we remove the template name, do you think it can make it difficult for one to trace back to the template they created? As an alternative, I think we can have the author's name saved in the template so that it's easier to filter but I'm sure there are other better ways.

arnaudstiegler:
For sure keeping the template name is better to retrieve some templates. 
Adding the author name is also useful, and as you mentioned probably a better way to filter out templates.

That being said, I believe the template name should not be used as a database key here:
- so that we don't have to force providing a name (no need for uninformative names, contributors don't have to add them)
- so that we don't have to deal with collisions in names (i.e 2 templates with same name will have different uuid anyway)

stephenbach:
I'll get to it tonight if no one claims it before then!

arnaudstiegler:
Yes, if one person is assigned to one dataset, there's a lot less risks of conflict and we don't really have to deal with that at all.
@stephenbach Happy to work on that if you want

stephenbach:
Sure, help is always appreciated! I'll tag you too. I was thinking we'd just modify `TemplateCollection` to read and write from the directory. We could add some code to handle modifications to avoid rewriting files that haven't been modified. Maybe read everything once and only write/reread files that are modified? Not sure if that's doable with the `@st.cache` annotation.

arnaudstiegler:
yes, that makes sense. 
I'm not super familiar with the repo yet, but it sounds like we could create a helper class, one per dataset that will link to the corresponding template.yaml. So that we would read/write with the class corresponding to the dataset
`TemplateCollection` would be reading all files / init those helper classes

stephenbach:
[...]

PR #43, opened 2021-05-31
Add uuid to each template | allow for name updates

arnaudstiegler:
This PR introduces:
- a unique identifier for each template entry
- we can now update the name of a template without having to create a new entry in the yaml file
- reformatted the current yaml files to have the newly added ids. Nothing has been changed in the names/templates/references

Even if we use uuids as keys for the templates, we still use the names to index them in the UI. So there's a mapping between name and ids to allow to pass a name to update a template, and there's a sync function that ensures any updates in self.templates is applied to the mapping as well

NB: all git diffs for the templates are solely due to some re-ordering based on the new ids

### DIFF
--- a/promptsource/promptsource.py
+++ b/promptsource/promptsource.py
@@ -215,7 +215,7 @@ if dataset_key is not None:
 
     dataset_templates = template_collection.get_dataset(dataset_key, conf_option.name if conf_option else None)
 
-    template_list = dataset_templates.all_names
+    template_list = dataset_templates.all_template_names
     num_templates = len(template_list)
     st.sidebar.write(
         "No of Templates created for "
@@ -253,7 +253,7 @@ if dataset_key is not None:
                 )
                 new_template_submitted = st.form_submit_button("Create")
                 if new_template_submitted:
-                    if new_template_name in dataset_templates.all_names:
+                    if new_template_name in dataset_templates.all_template_names:
                         st.error(
                             f"A template with the name {new_template_name} already exists "
                             f"for dataset {state.templates_key}."
@@ -272,7 +272,7 @@ if dataset_key is not None:
                     state.new_template_name = None
 
             dataset_templates = template_collection.get_dataset(*state.templates_key)
-            template_list = dataset_templates.all_names
+            template_list = dataset_templates.all_template_names
             if state.template_name:
                 index = template_list.index(state.template_name)
             else:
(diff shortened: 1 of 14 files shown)


arnaudstiegler:
#29

stephenbach:
Thank you! I submitted some templates for TREC right before you submitted this one. Would you mind pulling and updating the format for those too to make sure they're correct?

stephenbach:
I think there's still an issue (unrelated to trec). When I merge this PR into main locally I get the following error on the app:
```
AttributeError: 'Template' object has no attribute 'id'
Traceback:
File "/Users/bach/anaconda3/envs/bigscience/lib/python3.7/site-packages/streamlit/script_runner.py", line 338, in _run_script
    exec(code, module.__dict__)
File "/Users/bach/promptsource/promptsource/promptsource.py", line 84, in <module>
    template_collection = TemplateCollection()
File "promptsource/templates.py", line 27, in __init__
    self.datasets_templates: Dict[(str, Optional[str]), DatasetTemplates] = self._collect_dataset()
File "promptsource/templates.py", line 49, in _collect_dataset
    output[(dataset, filename)] = DatasetTemplates(dataset_name=dataset, subset_name=filename)
File "promptsource/templates.py", line 100, in __init__
    self.sync_mapping()
File "promptsource/templates.py", line 106, in sync_mapping
    self.name_to_id_mapping = {template.name: template.id for template in self.templates.values()}
File "promptsource/templates.py", line 106, in <dictcomp>
    self.name_to_id_mapping = {template.name: template.id for template in self.templates.values()}
```
It looks like your branch is behind main. Would you first pull main into your branch, test, and let us know when it's ready? Thank you!

arnaudstiegler:
Indeed, I was missing the winogrande folder in my branch which had a template.yaml with the old format (i.e no template.id).
Merged main into my branch, tested locally, all seems to work.

As an aside, the sorting of the templates within the `templates.yaml` is now based on the random uuid unfortunately... So when you add a new template, it's randomly inserted in the file. I don't see an easy fix for that for now, so I'm returning the sorted list of template names so that at least in the UI, the templates are sorted alphabetically.

stephenbach:
Beautiful!! Sorry to add more for you to do, but @VictorSanh just fixed the code quality checks. Would you pull main one more time, run `make quality` and `make style` and then do one more commit? Thank you!

arnaudstiegler:
Done!

stephenbach:
Awesome, thank you!

PR #46, opened 2021-06-01
High level metrics

VictorSanh:
Add a mode into `promptviz.py` (which i renamed `promptapp.py`)
- Mode `Helicopter view`: high-level metrics to track the progress of prompt collection. Will be useful once we start the collection at scale to hit our target
- Mode `Prompted dataset viewer`: same as previous (scrolling through a prompted dataset)

Capture d’écran 2021-05-31 à 10 42 05 PM

The real changes are between L41 and L72 (the rest is just indentation into an if condition)

Ultimately, we can have a third mode (the main app for sourcing)


stephenbach (approved):
Cool! LGTM!

PR #47, opened 2021-06-01
correct command for promptviz in README

manandey:


VictorSanh:
good catch, thanks @manandey !

Issue #26, opened 2021-05-24
Write Template Writing Guide


stephenbach:
There should be some guidelines for users. For now, we can assume they are BigScience participants, so the technical details of using the system are less important. It's more about what kind of information should be included in prompts, pitfalls to avoid, and other tips (i.e., how to map labels to strings in a way consistent with few-shot evaluation techniques).

VictorSanh:
+ add a step before submitting for code quality/style:

> **4. Format your code.** Run black and isort so that your newly added files look nice with the following command:
> ```bash
> make style
> make quality
> ```

PR #51, opened 2021-06-01
Template writing guide

stephenbach:


stephenbach:
Fixes #26 .

CONTRIBUTING.md
```diff
@@ -0,0 +1,129 @@
...
+
+Finally, remember that a template must produce two strings: a prompt and an output.
+To separate these two pieces, use three vertical bars `|||`.
+So, a complete template for AG News could be:
+```jinja2
+{{text}} 
+Is this a piece of news regarding {{"world politics"}}, {{"sports"}}, {{"business"}}, {{"science and technology"}}? ||| 
+{{ ["World politics", "Sports", "Business", "Science and technology"][label] }}
+```
+
```
srush, line 102: @VictorSanh and I were thinking that blank templates should just get filtered automatically.  So for Trec you could write: 

{% if label_coarse == 0 %}
Is this question asking for {{"a definition, a description, a manner of action, or a reason"}}? {{text}}
||| ... 
{% endif %}

Think that is nicer than having all the templates in one.

CONTRIBUTING.md
```diff
@@ -0,0 +1,129 @@
...
+directory in the repo will be modified. To upload it, following these steps:
+1. Run `make style` and `make quality`.
+2. Commit the modified template files (anything under `templates`) to git.
+3. Push to your fork on GitHub.
+4. Open a pull request against `main` on the PromptSource repo.
+5. When the PR is merged into main, mark the dataset in green on the spreadsheet. 
+
+
+## Jinja Cookbook
+TODO
```
srush, line 129: Maybe say something like 

"Jinja includes lots of complex features but for most instances you should only need to use ... "

We can have people ping the channel if they want to do crazy things.

srush (approved):

CONTRIBUTING.md
VictorSanh: Concrete example: https://github.com/bigscience-workshop/promptsource/blob/main/templates/trec/templates.yaml#L37 would be broken down into 4 templates which can optionally return "blank" in which case, we should ignore them when prompting the dataset

CONTRIBUTING.md
```diff
@@ -0,0 +1,138 @@
...
+out background information on the dataset by reading the information in the
+app. The dataset is a collection of examples, and each example is a Python
+dictionary. The sidebar will tell you the schema that each example has.
+1. **Start a new template**. Enter a name for your first template and hit "Create."
+You can always update the name later. If you want to cancel the template, select
+"Delete Template."
+1. **Write the template**. In the box labeled "Template," enter a Jinja expression.
+See the [getting started guide](#getting-started-using-jinja-to-write-templates)
+and [cookbook](#jinja-cookbook) for details on how to write templates.
+1. **Add a reference.** If your template was inspired by a paper, note the
```
craffel, line 36: What should the reference format be? Copy-paste bibtext from google scholar?

CONTRIBUTING.md
```diff
@@ -0,0 +1,138 @@
...
+what your template does.
+1. **Save the template**. Hit the "Save" button. The output of the template
+applied to the current example will appear in the right sidebar.
+1. **Verify the template**. Check that you didn't missed any case by scrolling through a handful of examples of the prompted dataset using the "Prompted dataset viewer" mode.
+1. **Upload the template(s).** Submit a PR using the instructions
+[here](#uploading-templates).
+1. **Duplicate the template(s).** If the dataset you have chosen bear the same
+format as other datasets (for instance `MNLI` and `SNLI` have identical format),
+you can simply claim these datasets and duplicate the templates you have written to these additional datasets.
+## Getting Started Using Jinja to Write Templates
```
craffel, line 47: One thing I don't see included here is using Python code. Not sure what the support is there or if you plan to put that in "cookbook". I could see it being necessary for datasets like SQuAD where examples aren't just "dict of features" https://huggingface.co/datasets/squad

craffel: SQuAD may actually be a nice example for "advanced usage" since I don't really know how we'd put that in jinja format OTOH..

CONTRIBUTING.md
stephenbach: Ah yes, this is an elegant solution to #4 . 👍

CONTRIBUTING.md
stephenbach: I'm not picky. I think Google Scholar bibtex is such a mess usually, I was just formatting in a human readable way, like
```
Making Pre-trained Language Models Better Few-shot Learners. Tianyu Gao, Adam Fisch, and Danqi Chen. ArXiv 2020.
```
But if wanted to create an automatic bibliography, that could be cool. In that case, we might need a bibtex specific field as well because we're already mixing notes and things in the "reference."

A solution could be: only bibtex in reference, any other description should go in the name?

CONTRIBUTING.md
stephenbach: The most recent discussion I recall is that things that can't be captured in Jinja will be expressed in python classes that subclass Template. Those won't be entered through the app but just programmed and committed. (Still need to write the code that merges everything.)

craffel: That seems reasonable to me.

CONTRIBUTING.md
```diff
@@ -113,16 +116,22 @@ correctness.
...
 ## Uploading Templates
 
 Once you save or modify a template, the corresponding file inside the `templates`
 directory in the repo will be modified. To upload it, following these steps:
-1. Run `make style` and `make quality`.
-2. Commit the modified template files (anything under `templates`) to git.
-3. Push to your fork on GitHub.
-4. Open a pull request against `main` on the PromptSource repo.
-5. When the PR is merged into main, mark the dataset in green on the spreadsheet. 
+1.
```
stephenbach, line 129: @VictorSanh Was there a step you wanted to add here?

CONTRIBUTING.md
stephenbach: Ehh... I'm already cooling on this idea. I would prefer to just leave it unconstrained unless someone feels differently.

CONTRIBUTING.md
VictorSanh: oh no, i just messed up, corrected it

CONTRIBUTING.md
```diff
@@ -0,0 +1,129 @@
...
+"Delete Template."
+1. **Write the template**. In the box labeled "Template," enter a Jinja expression.
+See the [getting started guide](#getting-started-using-jinja-to-write-templates)
+and [cookbook](#jinja-cookbook) for details on how to write templates.
+1. **Add a reference.** If your template was inspired by a paper, note the
+reference in the "Template Reference" section. You can also add a description of
+what your template does.
+1. **Save the template**. Hit the "Save" button. The output of the template
+applied to the current example will appear in the right sidebar.
+1. **Upload the template(s).** Submit a PR using the instructions
```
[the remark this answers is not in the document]
stephenbach: Sorry, what's repeated?

CONTRIBUTING.md
VictorSanh: I don't have a strong opinion on this
as long as it's easy to find the reference, it should do the trick
at least: link to pdf + table number or link to pdf + permalink to the line of code

VictorSanh (approved):

VictorSanh:
I like the current version, I think we can merge it and iterate further if necessary @stephenbach

PR #52, opened 2021-06-02
Fixed XNLI prompts entailment direction

awebson:
Currently there are two XNLI prompts that ask "does Sentence 2 entail Sentence 1". Note that for entailment, we can only prompt “does S1 entail S2” (does the premise entail the hypothesis) not the other way around. For contradiction, we could maybe either prompt "does S1 contradict S2" or “does S2 contradict S1”, but I sticked to the former to be consistent. Would be an interesting experiment to see if it generalizes to the latter though. 

I also fixed some typos and standardized line breaks to match Figure G7 (ANLI) of the GPT-3 paper. If they look good, I will write more prompts from other papers and copy and paste them to other NLI datasets.

### DIFF
--- a/templates/xnli/en/templates.yaml
+++ b/templates/xnli/en/templates.yaml
@@ -3,29 +3,29 @@ subset: en
 templates:
   4e122d26-7e79-49c0-961b-cf8ee134759e: !Template
     id: 4e122d26-7e79-49c0-961b-cf8ee134759e
-    jinja: "Sentence 1: {{premise}}\nSentence 2: {{hypothesis}}\n\nDoes Sentence 1\
-      \ contradict Sentence 2? Yes, No or Neutral? |||\n{% if label == 0 %} \nNo\n\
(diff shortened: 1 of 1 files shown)


VictorSanh:
oh yes, I messed up (I always talk about entailment but never use the verb entail though...) thank you @awebson !

I think these choices on line breaks, including "question" or not, including a comma or not are arbitrary and we are looking for this diversity. Ultimately, a model shouldn't be sensitive to how the prompt+x is formulated

I'll merge your pr

awebson:
Thanks! Yeah sometimes I also have to pause for a second and think, wait, which one is the premise and which one is the hypothesis again? I forgot which paper started doing it but they renamed P and H to just “context” and “question”, which are much less confusing. 

I 100% agree with you that we’d like the model to be robust to line breaks and stuff. GPT-3 is still weirdly sensitive to them (e.g., bottom left paragraph at p. 6 of [Reynolds & McDonell 2021](https://arxiv.org/pdf/2102.07350.pdf)), and you’re right that having diverse punctuations in the training set may improve invariance.

PR #54, opened 2021-06-02
Templates of IMDB

Urmish:


templates.yaml
```diff
@@ -45,3 +45,44 @@ ag_news:
       {{["Not-Related", "Related"][label]}}'
     name: Test1
     reference: ''
+imdb:
+  imdb_1: !Template
+    jinja: '{{text}}
+
+      What is the sentiment expressed in this text?
+
+      {{ ["neg", "pos"] [label] }}'
```
srush, line 54: Use words or english phrases as opposed to abbreviations "negative" "positive" etc.

srush (changes requested):

templates.yaml
Urmish: Done

PR #56, opened 2021-06-03
paws and paws-x template added

sbmaruf:
Special Notes:
1. There are some labels that are -1. According to the documentation, they are "missing" or "wrong". So I have ignored them while writing the prompt.
2. The prompts for subset `labeled_final`, `labeled_swap` and `unlabeled_final` are same. Initially I have created prompts for `labeled_final` then copy-pasted the `templates/paws/labeled_final/templates.yaml` to `templates/paws/labeled_swap/templates.yaml` and `templates/paws/unlabeled_final/templates.yaml`. [I also changed the `subset` field in each of the `templates.yaml` file](https://github.com/bigscience-workshop/promptsource/pull/56/commits/e5129b82a87c3983fd1a72ed6bc34a42a87132a2). Similar way I created prompt for `paws-x` `en` subset prompts.  But I see that each time I create a new prompt, a hash is generated by the app in the  `templates.yaml`. I wonder if there will be an issue because for copy-pasting the hash remains the same.

### DIFF
--- a/templates/paws-x/en/templates.yaml
+++ b/templates/paws-x/en/templates.yaml
@@ -1,4 +1,4 @@
-dataset: paws
+dataset: paws-x
 subset: en
 templates:
   2ff509a8-2712-4406-97bd-ce86824cdc18: !Template
--- a/templates/paws-x/en/templates.yaml
+++ b/templates/paws-x/en/templates.yaml
@@ -1,0 +1,66 @@
+dataset: paws
+subset: labeled_final
+templates:
+  2ff509a8-2712-4406-97bd-ce86824cdc18: !Template
+    id: 2ff509a8-2712-4406-97bd-ce86824cdc18
+    jinja: "{% if label == 0 or label == 1 %} \n{{sentence1}} Question: {{sentence2}}\
+      \ paraphrase or not?\nSentence 1 paraphrase Sentence 2? Yes or No? \n{% endif\
+      \ %}\n||| \n{% if label == 0 %} \nFalse\n{% elif label == 1 %}\nTrue\n{% endif\
+      \ %}"
+    name: PAWS-ANLI GPT3-no-label
+    reference: ANLI prompt format from Table G7 in the GPT3 paper. Additionally added
+      task information without any label..
+  32c0f3d5-a6c7-4ab2-a922-3a9d66db309f: !Template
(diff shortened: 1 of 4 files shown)


VictorSanh:
thanks @sbmaruf ! looking at it right now

@arnaudstiegler i agree that duplicating prompts for different subsets (or datasets) by copy-pasting is the most straightforward way to do it. it messes up with the uniqueness of the uuids... Will that be a problem?
cc @stephenbach

stephenbach:
I can't spot any bugs that we'd hit if templates in different subsets have the same uuid. They're all siloed by subset. So I think as long as we update the documentation to say they're only unique within a `DatasetTemplates` object, which is 1:1 with .yaml files, I think we're ok.

Otherwise LGTM. I think it's fine not to escape the "Yes" and "No" options., although we don't actually address that in the current guidelines.

arnaudstiegler:
Yes, we are only using the ids within each subset so there's no problem with having duplicates. And I don't see why we would need an absolute id across datasets at this time.

One concern related to copy/paste in general would be that if you copy/paste directly form the templates.yaml, it means you might not have actually tested it on the dataset. But in this case, if subsets have same format, makes total sense to do it

VictorSanh:
> I can't spot any bugs that we'd hit if templates in different subsets have the same uuid. They're all siloed by subset. So I think as long as we update the documentation to say they're only unique within a DatasetTemplates object, which is 1:1 with .yaml files, I think we're ok.

Ok! I'll do that now

> I think it's fine not to escape the "Yes" and "No" options., although we don't actually address that in the current guidelines.

Question for brainstorm: if the question is "True or False?" do we authorize "Yes/No" as labels? I think that for yes/true, no/false it's fine but I can imagine more complicated setups where this is like shooting ourselves in the foot (and thus escaping the label in the prompt)

Otherwise, LGTM @sbmaruf !

VictorSanh:
> Question for brainstorm: if the question is "True or False?" do we authorize "Yes/No" as labels? I think that for yes/true, no/false it's fine but I can imagine more complicated setups where this is like shooting ourselves in the foot (and thus escaping the label in the prompt)

I fixed the examples we already have by protecting/escaping any label in the prompt that is not a binary Yes/No/True/False/Positive/Negative.

@sbmaruf I'm merging your pull request!

PR #57, opened 2021-06-03
Fix: Paws and paws-x 

sbmaruf:
This pull 
- fixes a previous bug in prompt.
- fixes a typo on ' ' and '.'
- adds four additional prompts suggested by "Iz Beltagy"
- adds two additional prompts. The idea is to create a generative task from the dataset. The dataset asks the question, given two-sentence, if they are paraphrase or not. What I did is, 
```
{% if label == 1 %} 
Paraphrase the sentence: {{sentence1}} 
{% endif %}
||| 
{{sentence2}} 

```
I added prompt for both directions.

### DIFF
--- a/templates/paws-x/en/templates.yaml
+++ b/templates/paws-x/en/templates.yaml
@@ -1,6 +1,12 @@
 dataset: paws-x
 subset: en
 templates:
+  289448c3-63fc-4904-8387-86d232424243: !Template
+    id: 289448c3-63fc-4904-8387-86d232424243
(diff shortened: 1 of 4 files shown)


stephenbach (changes requested):
Thanks for the PR! The only issue I see is that the relation names "paraphrase" and "non-paraphrase" should be escaped in the various templates, since they're not exactly yes/no or true/false, as per discussion in #56.

sbmaruf:
@stephenbach Can you take a look now?

stephenbach:
Awesome, thank you!

PR #58, opened 2021-06-04
Adds automatic tests for templates

### LINKED ISSUE #48: Add some dummy checks to the jinja template
Add a suite of sanity checks on the jinja template. for instance,
- are all variables called in the template actual keys in the example dictionary?
- template content is a duplicated with another one

Please add ideas!

+ can duplicate these tests into the "common pitfalls" section of the contribution guide (#26)

stephenbach:
Resolves #48.

### DIFF
--- a/test/validate_templates.py
+++ b/test/validate_templates.py
@@ -1,0 +1,58 @@
+from jinja2 import BaseLoader, Environment, meta
+from templates import TemplateCollection
+from utils import get_dataset_builder
+
+#
+# This script validates all the templates in the repository with simple syntactic
+# checks:
+#
+# 0. Are all templates parsable YAML?
+# 1. Are all referenced variables in the dataset schema?
+# 2. Does the template contain a prompt/output separator "|||" ?
+# 3. Are all names and templates within a data (sub)set unique?
+#
+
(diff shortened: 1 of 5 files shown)


test/validate_templates.py
```diff
@@ -0,0 +1,63 @@
...
+# 2. Does the template contain a prompt/output separator "|||" ?
+# 3. Are all names and templates within a data (sub)set unique?
+#
+
+# Sets up Jinja environment
+env = Environment(loader=BaseLoader)
+
+# Loads templates and iterates over each data (sub)set
+template_collection = TemplateCollection()
+for (dataset_name, subset_name) in template_collection.keys:
```
VictorSanh, line 20: you can also separate a test call per dataset. Otherwise the for loop will exit at the first error you encounter rather than throwing raising for all data (sub) sets.

An example here: https://github.com/huggingface/datasets/blob/master/tests/test_dataset_cards.py#L74

VictorSanh (approved):

VictorSanh:
Modulo my minor comment on separating the test per dataset, this lgtm
Thank you Steve!

PR #59, opened 2021-06-04
Prompts for S/M/X/ANLI, RTE, CB, SciTail, HANS

awebson:
Now there are at least 5 templates for each NLI task. Roughly, the three-way NLI tasks (S/M/X/ANLI and SuperGLUE CB) share the same prompt templates, while the binary NLI tasks (RTE, SciTail, and HANS) share the same templates. But there are multiple exceptions. For example, SciTail only has neutral labels without any contradiction labels, and it sounds weird to say, e.g., based on the premise the hypothesis is False. 

I manually checked each task and read out loud each prompt with some examples to make sure that they all sound reasonably natural. I erred on the slightly more conservative side because I find some prompts in the literature, e.g., [here](https://arxiv.org/pdf/2105.11447.pdf) on p. 20 "Premise… Hypothesis… Therefore the answer is yes", reads quite unnatural. (Maybe it's fine that they somehow work in evaluation, but we might want higher-standard prompts for training.) 

@VictorSanh wrote some prompts for XNLI that ask the model to predict the contraposition. I didn't adopt those for the binary NLI tasks because it sounded weird to ask "does P contradict H", when the label is only "non-entailment". (We could in theory ask "does P not entail H", but I have zero faith in neural nets today understanding any negation.)

Anyway. NLI annotations are not meant to be strictly logical/truth-conditional entailment, and (for humans at least) the exact phrasing of prompts really have meaningful pragmatic effects. I'm sure the ones I wrote today include some mistakes, so scrutiny is obviously welcome.

### DIFF
--- a/templates/anli/templates.yaml
+++ b/templates/anli/templates.yaml
@@ -6,12 +6,12 @@ templates:
       \ 1 contradict Sentence 2? Yes, No, or Neutral? |||\n{% if label == 0 %} \n\
       No\n{% elif label == 1 %}\nNeutral\n{% else %}\nYes\n{% endif %}"
     name: does S1 contradict S2?
-    reference: Copied from Victor's prompt for XNLI.
+    reference: Copied from Victor's prompts for XNLI.
   5459237b-97de-4340-bf7b-2939c3f7ca19: !Template
     id: 5459237b-97de-4340-bf7b-2939c3f7ca19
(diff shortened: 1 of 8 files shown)


stephenbach:

awebson:
Done!

VictorSanh:
Looks good to me @awebson, didn't find any obvious error but I can be wrong
merging it now, please update the spreadsheet

PR #60, opened 2021-06-04
Update CONTRIBUTING.md

ibeltagy:
Following a conversation on slack, updating the instructions to include examples of two predefined formats.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,8 +42,9 @@ applied to the current example will appear in the right sidebar.
 through a handful of examples of the prompted dataset using the
 "Prompted dataset viewer" mode.
 1. **Write between 5 and 10 templates**. Repeat the steps 4 to 8 to create between 5
-and 10 (more if you want!) templates per dataset/subset. Feel free to introduce some diversity
-both in the format and the formulation.
+and 10 (more if you want!) templates per dataset/subset. Feel free to introduce
+a mix of formats, some that follow the templates listed in the [best practices](#best-practices)
+and some that are more diverse in the format and the formulation. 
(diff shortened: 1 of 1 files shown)


sbmaruf:
Hi! I recently adopted your suggestion in paws/paws-x dataset.

### multiple-choice
```
Sentence 1: {{sentence1}}
Sentence 2: {{sentence2}}
Select the relation between Sentence 1 and Sentence 2?
A) Paraphrase B) Not-paraphrase
|||
{% if label == 0 %} 
Not-paraphrase
{% elif label == 1 %}
Paraphrase
{% endif %}
```
 
### task_description-input

```
Determine the relation between the following two sentences. The relations are Paraphrase, Not-paraphrase
{{sentence1}}
{{sentence2}}
||| 
{% if label == 0 %} 
Not-paraphrase
{% elif label == 1 %}
Paraphrase
{% endif %}
```

Do you see any problem with the prompt style? Like in `multiple-choice`, Should we include `Sentence 1` and `Sentence 2`? Also for adding labels, should we add them with `A)`, `B)`, `C)` opinion or only label is ok?

VictorSanh:
Looks great, thank you @ibeltagy !

VictorSanh:
> Do you see any problem with the prompt style? Like in `multiple-choice`, Should we include `Sentence 1` and `Sentence 2`? Also for adding labels, should we add them with `A)`, `B)`, `C)` opinion or only label is ok?

I'll Iz answer more precisely, I think it's mostly a question of conventions so for the sake of consistency across datasets, maybe it's safer to remove these A/B/C/D in the prompt since that's how we have it in the guide

PR #62, opened 2021-06-04
Add billsum templates

cccntu:
Note: the text and summary are long.
Image from paper: https://arxiv.org/abs/1910.00523
image


stephenbach:

stephenbach:
LGTM!

PR #64, opened 2021-06-04
prompts for amazon_reviews_multi and math_qa

manandey:


stephenbach:

manandey:
> Thank you for the PR! Would you mind pulling `main` before we merge? We've just added some automatic checks for templates that we'd like to run.

Updated @stephenbach

stephenbach (changes requested):
Thanks! You don't need to escape words like "stars" and "review." Escaping is only for words that affect the correctness of the output, which is usually class names. So, for example, you could replace "product category" with "type" and it would still be a sensible prompt.

manandey:
> Thanks! You don't need to escape words like "stars" and "review." Escaping is only for words that affect the correctness of the output, which is usually class names. So, for example, you could replace "product category" with "type" and it would still be a sensible prompt.

Updated the changes, @stephenbach

stephenbach (approved):
Awesome, thank you! Please go ahead and update the spreadsheet!

PR #70, opened 2021-06-04
Add GLUE templates

craffel:
I wrote 5 templates for MNLI and copied them into MNLI-matched and -mismatched, since latter two subsets are validation/test-only. All the other entailment datasets use different sets of prompts.

### DIFF
--- a/templates/glue/cola/templates.yaml
+++ b/templates/glue/cola/templates.yaml
@@ -1,9 +1,66 @@
 dataset: glue
 subset: cola
 templates:
+  1d3f5f15-8128-4445-8de5-92365b7e54a8: !Template
+    id: 1d3f5f15-8128-4445-8de5-92365b7e54a8
+    jinja: 'Does the following sentence make sense and use correct English? Please
+      answer {{"yes"}} or {{"no"}}.
+
+      {{sentence}}
+
+      |||
+
+      {{ ["no", "yes"][label] }}'
+    name: Make sense yes no
+    reference: ''
   39a701ff-bb4b-48ac-8c0a-8c61bf0d4b8d: !Template
     id: 39a701ff-bb4b-48ac-8c0a-8c61bf0d4b8d
-    jinja: '{{sentence}} \nIs this example grammatically correct? ||| {{ ["No", "Yes"][label]
-      }}'
+    jinja: '{{sentence}}
+
+
+      Is this example grammatically correct and sensible?
+
+      |||
+
+      {{ ["No", "Yes"][label] }}'
     name: jinja_example
     reference: A sample glue template
+  6f49b860-9145-4fcb-b632-9faea39e254e: !Template
+    id: 6f49b860-9145-4fcb-b632-9faea39e254e
+    jinja: 'I''m copy-editing a story for publication. It has the following sentence
+      in it:
+
+      {{sentence}}
+
+      Does that sentence make sense and is it grammatically correct? Please answer
+      yes or no.
+
+      |||
+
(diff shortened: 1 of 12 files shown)


templates/glue/rte/templates.yaml
```diff
@@ -0,0 +1,66 @@
...
+      {{["yes", "no"][label]}}'
+    name: imply
+    reference: ''
+  f56ffced-9b16-431a-8a17-501e63cddf73: !Template
+    id: f56ffced-9b16-431a-8a17-501e63cddf73
+    jinja: '{{sentence2}}
+
+      Does this imply
+
+      {{sentence1}}
```
VictorSanh, line 58: Does this imply {{sentence1}} *?*

templates/glue/stsb/templates.yaml
```diff
@@ -0,0 +1,72 @@
+dataset: glue
+subset: stsb
+templates:
+  50e3a541-108c-4b26-a423-956562d9b3af: !Template
+    id: 50e3a541-108c-4b26-a423-956562d9b3af
+    jinja: Rank on a scale from {{"0.0"}} to {{"5.0"}} how similar the sentences "{{sentence1}}"
+      and "{{sentence2}}" are. ||| {{label}}
```
VictorSanh, line 7: i wonder if we should not have some round-up on the float... wdyt?
It's not really informative beyond the 2 first decimals and I am not sure it's fair to expect such a precise generation from a model

templates/glue/cola/templates.yaml
```diff
@@ -1,9 +1,66 @@
...
+      Is this example grammatically correct and sensible?
+
+      |||
+
+      {{ ["No", "Yes"][label] }}'
     name: jinja_example
     reference: A sample glue template
+  6f49b860-9145-4fcb-b632-9faea39e254e: !Template
+    id: 6f49b860-9145-4fcb-b632-9faea39e254e
+    jinja: 'I''m copy-editing a story for publication. It has the following sentence
```
VictorSanh, line 30: love this creativity haha

VictorSanh (approved):
Looks good to me!

Love the creativity!

See my comment on stsb

templates/glue/stsb/templates.yaml
craffel: Most STS-B labels are quantized to the nearest 0.2, so for T5 during training we just quantized the training labels to the nearest 0.2. https://github.com/google-research/text-to-text-transfer-transformer/blob/master/t5/data/preprocessors.py#L816
Still got SoTA. Can we do something like `round(label*5)/5` or even `round(label*10)/10` (if we want it to be more generic) in jinja? I think if the model only sees this precision during training, it will not try to generate silly precision during inference.

craffel: Looks like I can do ``{{label|round(1)}}`` instead of ``{{label}}``. Do you think we should go all-in for the 0.2 rounding, or just the 1-decimal-place?

VictorSanh: oh awesome!
yes 0.2 rounding looks good since it worked well enough for t5!

templates/glue/rte/templates.yaml
craffel: This would make the formatting weird because RTE sentences can have terminal punctuation marks themselves, e.g. it would look like
```
Weapons of Mass Destruction Found in Iraq.
Does this imply
No Weapons of Mass Destruction Found in Iraq Yet. ?
Please answer A) yes or B) no.

PR #73, opened 2021-06-04
add Zest

VictorSanh:
Note: this is an example where all the splits don't have all the same features. The test split doesn't have the "domain" feature (always set to `null`). Jinja replaces it with `None`.
I still left the "domain" feature because it's secondary information and doesn't disrupt the flow.

### DIFF
--- a/templates/zest/templates.yaml
+++ b/templates/zest/templates.yaml
@@ -1,0 +1,140 @@
+dataset: zest
+templates:
+  223c6226-5f2d-4dd8-9710-4657ffb54f13: !Template
+    id: 223c6226-5f2d-4dd8-9710-4657ffb54f13
+    jinja: '{{context}}
+
+      {{question}}|||
+
+      {% if answer[0] == "n/a" %}
+
+      I don''t know
+
+      {% else %}
+
+      answer[0]
+
+      {% endif %}'
+    name: concatenate
+    reference: Concatenate question and question
+  2283cebf-988e-4bff-96bf-982a09963e49: !Template
+    id: 2283cebf-988e-4bff-96bf-982a09963e49
+    jinja: 'Decide whether the question "{{question}}" is answerable solely based
+      on this passage: {{context}}|||
+
+      {% if answer[0] == "n/a" %}
+
+      No
+
+      {% else %}
+
+      Yes
+
+      {% endif %}'
+    name: answerable_or_not
+    reference: Decide whether this question is answerable
+  6f694e45-1d17-4067-a1f6-7dae89c148db: !Template
+    id: 6f694e45-1d17-4067-a1f6-7dae89c148db
+    jinja: 'My daughter is asking me a question about {{domain}}: {{question}}
+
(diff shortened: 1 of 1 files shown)


srush:
Do we allow different templates for splits? I thought it was just the sub-groups.

VictorSanh:
> Do we allow different templates for splits? I thought it was just the sub-groups.

No, right now we don't have such a logic (different templates for different split).
For the sake of simplicity, I think we should check before applying the jinja whether all the fields are present. I can imagine that for many datasets, the test split just doesn't have the fields we use as labels.

templates/zest/templates.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+      {{question}}|||
+
+      {% if answer[0] == "n/a" %}
+
+      Can''t answer
+
+      {% else %}
+
+      answer[0]
```
srush, line 72: Needs to be in a {{ }}

templates/zest/templates.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+      {% endif %}'
+    name: gpt3instruct_format
+    reference: Template format from GPT3 instruct
+  846cc8ff-0527-4b2f-8da4-46613e915ff5: !Template
+    id: 846cc8ff-0527-4b2f-8da4-46613e915ff5
+    jinja: '{% if answer[0] != "n/a" %}{{context}}
+
+      Based on the previous passage, generate a question which has for answer {{answer[0]}}.
```
srush, line 81: has "the following passage for an answer"

templates/zest/templates.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+      {{question}}|||
+
+      {% if answer[0] == "n/a" %}
+
+      Can''t answer
+
+      {% else %}
+
+      answer[0]
```
srush, line 119: Needs to be {{ }}

templates/zest/templates.yaml
```diff
@@ -0,0 +1,140 @@
...
+      |||
+
+      {{question}}
+
+      {% endif %}'
+    name: generate_the_question_with_domain
+    reference: Generate the question with domain
+  bdaf4f8a-2344-4e46-a52b-2045a080a4b2: !Template
+    id: bdaf4f8a-2344-4e46-a52b-2045a080a4b2
+    jinja: 'Answer this question about {{domain}} based on the context. If the question
```
srush, line 106: domain is of the form "national_parks" I think you could replace the "_" with a space?

templates/zest/templates.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+      What should be their answer based on this context: {{context}}|||
+
+      {% if answer[0] == "n/a" %}
+
+      I don''t know
+
+      {% else %}
+
+      answer[0]
```
srush, line 136: Needs to be {{ }}

srush (changes requested):

srush (approved):

PR #74, opened 2021-06-04
Adapted Perez et al. 2021’s prompts for SuperGLUE

awebson:
Per @VictorSanh’s request, I added Ethan Perez’s prompts for SuperGLUE (from his repo [here](https://github.com/ethanjperez/true_few_shot/tree/main/templates.super_glue)). 

Many of them are near duplicates of each other, differing only in whether the model is supposed to answer "Yes" or "No" vs. "True" or "False". This is probably something we can address/augment programmatically for all binary classification tasks, so for now I wrote all templates with "Yes/No".

### DIFF
--- a/templates/anli/templates.yaml
+++ b/templates/anli/templates.yaml
@@ -30,7 +30,7 @@ templates:
     jinja: '{{premise}} Based on the previous passage, is it true that {{hypothesis}}
       Yes, no, or maybe? ||| {{ ["Yes", "Maybe", "No"][label] }}'
     name: based on the previous passage
-    reference: "Adopted from the BoolQ prompts in Schick & Sch\xFCtze 2021."
+    reference: "Adapted from the BoolQ prompts in Schick & Sch\xFCtze 2021."
   cf55f09b-bc68-439d-a9b2-781263509f99: !Template
     id: cf55f09b-bc68-439d-a9b2-781263509f99
     jinja: "Sentence 1: {{premise}}\nSentence 2: {{hypothesis}}\nQuestion: Does Sentence\
--- a/templates/hans/templates.yaml
+++ b/templates/hans/templates.yaml
@@ -18,7 +18,7 @@ templates:
       \ 1 entail Sentence 2? Yes or no? |||\n{% if label == 0 %} \nYes\n{% else %}\n\
       No\n{% endif %}"
     name: does S1 entail S2?
-    reference: Adopted from Victor's prompts for XNLI.
+    reference: Adapted from Victor's prompts for XNLI.
   9666d06a-63eb-4992-a1f5-3e582ffe39a6: !Template
     id: 9666d06a-63eb-4992-a1f5-3e582ffe39a6
     jinja: Given that {{premise}} Does it follow that {{hypothesis}} Yes or no? |||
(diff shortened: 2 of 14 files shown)


stephenbach:
Thanks! Looks great! A couple of questions:
- Why is `templates/scitail/snli_format/templates.yaml` deleted?
- Are there other templates for superglue that we should add at the same time?

awebson:
1. The `test1` prompt in `scitail/snli_format` asked "Are these sentences {{"neutral or entailment"}} to one another?" But NLI entailment is always annotated when the premise entails the hypothesis, never the the other way around or bidirectional as in "entailment to one another". So I deleted this prompt. Its correct unidirectional version is already included under the `tsv_format` split. (All splits have the same data in SciTail.)

2. Good question! I read most of the papers linked in our literature review google doc and I think most of the manually written prompts built up on either the GPT-3 paper or the Schick & Schütze paper, and I included those prompts in this PR (with the caveat as mentioned above that I deduplicated the highly similar ones). I'm aware that aside from RTE and CB we still have less than 5 templates per SuperGLUE task. I will write more next week, but also obviously others are welcome to contribute as well.

stephenbach:
Ok, thanks, let's do 2 PRs then, one for each dataset. For SciTail, if the dataset has the same format for different subsets, let's copy the templates so that each subset is covered. If you can make those changes, we'll merge in these for superglue. We can leave it marked as a WIP on the spreadsheet.

awebson:
Ah sorry maybe I wasn't clear. For SciTail, it has 4 "subsets": `snli_format`, `tsv_format`, `dgem_format`, and `predictor_format`. They all have the same data*, but `tsv_format` is the closest to all other NLI datasets. So in #59 I wrote prompts for `tsv_format`, and this PR deletes an incorrect prompt under `snli_format`.

*They all have the same premises, hypotheses, and labels. Although different formats have different extra data (e.g., syntatic parse or graph structure for each sentence) which would require extra special prompts. 

Would you still like me to make 2 PRs? (One for SuperGLUE and one for SciTail?)

stephenbach:
Sure, let's separate these so that we can finish this part of superglue. I'm not clear on do you think there's value in writing templates for the other subsets?

awebson:
Oh I was thinking the `tsv_format` for SciTail is enough and there is no need to write prompts for the other SciTail format "subsets" in the near future. Certainly for the purposes of our experiments, we’re not studying whether adding explicit parse trees would help. Which is why I thought we could just squeeze in this minor SciTail fix into this one PR?

stephenbach:
There are multiple tasks that can be created from the different annotations. For example, in the predictor format, there is a question answering task.

stephenbach:
Awesome, thanks!

awebson:
Oh I see. You’re right. I knew SciTail was built from several QA sets but I didn't know that the predictor format actually recasted them as such. Also, I just noticed that some part of SciTail is actually its own QA set (e.g., [SciQ](https://huggingface.co/datasets/sciq)) that are also in the Hugging Face `datasets` library.

I reverted any changes to SciTail in this PR (except the typo fixes), and I will mark both SciTail and SuperGLUE as WIP in our spreadsheet.

PR #75, opened 2021-06-04
OpenBookQA prompts

arnaudstiegler:
It looks like the `additional` subset for this dataset is identical to the `main` subset. I'm not 100% sure so I duplicated the templates for the two subsets.

Otherwise, I did not create prompts without providing choices because some of the text frames the question as "Which of these is the right answer" which makes no sense without having the choices.
For instance:
text: `a lack of illumination would result in which of these?`
Choices:
```
- the restroom overflowing with water
- a principal's door being locked
- a student unable to view what he is
writing
- a teacher being unable to transport to
school
```
In that case, it would be non-sensical not to provide the choices.


stephenbach:
Looks great! Only one optional question/suggestion:
Would it make sense to replace
`Is the right answer {{'', ''.join(["A", "B", "C", "D"])}} ?`
with
`Is the right answer {{''A, B, C, or D"}} ?`
Right now, there's no `or`, which maybe is helpful to indicate it's an either-or choice?

arnaudstiegler:
Agreed! I updated the prompt

stephenbach:
Perfect! And thanks for catching the subset name fix too. Good to know to keep an eye for it when copying between subsets!

PR #78, opened 2021-06-05
cnn_dailynews

arunraja-hub:
This is my first contribution to hackaprompt


stephenbach:
Looks great! The creativity is awesome!

stephenbach:
You can mark it as completed on the spreadsheet. I don't think there's a need to do the earlier versions unless there's differences in the data and/or annotations.

PR #81, opened 2021-06-05
Add prompts for `piqa` dataset

cccntu:


templates/piqa/templates.yaml
```diff
@@ -0,0 +1,121 @@
...
+      |||
+
+      {{["Yes", "No"][label]}}'
+    name: Does this solution make sense? sol1
+    reference: ''
+  f42cd457-a14b-465a-a139-d7d2407a3bac: !Template
+    id: f42cd457-a14b-465a-a139-d7d2407a3bac
+    jinja: 'Given a sentence, correct it if it doesn''t make sense.
+
+      Input: {{goal}} {{sol1}}
```
srush, line 111: This phrasing is a bit awkward with capitalization. Maybe you could make it more explicit? It looks like this:

```
Given a sentence, correct it if it doesn't make sense.
Input: When boiling butter, when it's ready, you can Pour it onto a plate
Output:```

or maybe just use {{sol1[0].lower() + sol[1:]}}

File: templates/piqa/templates.yaml [hunk 2]
```diff
@@ -0,0 +1,121 @@
...
+    reference: ''
+  99565244-4eaf-4004-a28b-4362ba5bcac3: !Template
+    id: 99565244-4eaf-4004-a28b-4362ba5bcac3
+    jinja: '{{goal}} {{sol2}}
+
+      Does this solution make sense?
+
+      |||
+
+      {{["No", "Yes"][label]}}'
```
srush, line 52:
```
When boiling butter, when it's ready, you can Pour it onto a plate
Does this solution make sense?
```

"Solution" reads strangely to me. Maybe just say "does this phrase make sense?" or alternatively say "Given the problem {{prob}} is {{sol}} as solution to that problem?"

srush (changes requested):
small comments above

templates/piqa/templates.yaml
cccntu: Not sure if we should use `sol1[0].lower() + sol1[1:]}` everywhere, some solutions do need capitalization. But I guess lower is better. I updated most of the templates to use this.

srush (approved):

PR #82, opened 2021-06-05
Viewer -> color the outputs based on whether they are variables or text

srush:
Small tweak to make the viewer show which text came from data and which came from a prompt. 

image

### DIFF
--- a/promptsource/promptsource.py
+++ b/promptsource/promptsource.py
@@ -53,14 +53,13 @@ WIDTH = 80
 def show_jinja(t, width=WIDTH):
     wrap = textwrap.fill(t, width=width, replace_whitespace=False)
     out = highlight(wrap, DjangoLexer(), HtmlFormatter())
-    # st.write(out.replace("\n\n", "\n"), unsafe_allow_html=True)
     st.write(out, unsafe_allow_html=True)
 
 
 def show_text(t, width=WIDTH):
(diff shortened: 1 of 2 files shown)


promptsource/promptsource.py
```diff
@@ -53,14 +53,13 @@ def reset_template_state():
...
     out = highlight(wrap, DjangoLexer(), HtmlFormatter())
-    # st.write(out.replace("\n\n", "\n"), unsafe_allow_html=True)
     st.write(out, unsafe_allow_html=True)
 
 
 def show_text(t, width=WIDTH):
     wrap = [textwrap.fill(subt, width=width, replace_whitespace=False) for subt in t.split("\n")]
     wrap = "\n".join(wrap)
-    st.text(wrap)
+    st.write(wrap, unsafe_allow_html=True)
```
VictorSanh, line 62: the only minor thing that bothers me with `st.text` vs `st.write` is that `write` tends to mess up with the line breaks 
 and "===" sequences in the template ( i am being very picky right now, feel free to ignore)

text:
Capture d’écran 2021-06-05 à 10 30 32 AM

write:
Capture d’écran 2021-06-05 à 10 30 44 AM

VictorSanh:
that's awesome, thanks!
looks good to me modulo the code quality checks

promptsource/promptsource.py
srush: Oh yeah it must think it is markdown. I'll see if I can fix.

PR #87, opened 2021-06-07
prompts for `liar`

manandey:


templates/liar/templates.yaml
```diff
@@ -0,0 +1,78 @@
+dataset: liar
+templates:
+  331fec6f-bc64-4a81-b895-b1c5d5a61e7e: !Template
+    id: 331fec6f-bc64-4a81-b895-b1c5d5a61e7e
+    jinja: 'What was the party affiliation of the speaker of the following statement?
```
craffel, line 5: It might be hard for the model to free-form guess a party affiliation - can you create a variant of this prompt where you list out the possible choices for the party affiliation? Or is this free-form, i.e. there is no finite set of party affiliations to choose from?

templates/liar/templates.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+      ===
```
craffel, line 7: Can you include some prompts without this `===` or remove it from some of the prompts? I don't think this is a common feature of prompts, and it might teach the model to associate "===" with this dataset.

templates/liar/templates.yaml
```diff
@@ -0,0 +1,78 @@
...
+  d5b1af62-fc94-4031-9e9f-8ea6532926e0: !Template
+    id: d5b1af62-fc94-4031-9e9f-8ea6532926e0
+    jinja: 'Guess the context in which the statement was stated by speaker {{speaker}},
+      who is a {{job_title}}
+
+      ===
+
+      {{statement}} |||
+
+      {{context}}'
```
craffel, line 64: Same comment about providing choices for the context, if they are available.

craffel (changes requested):
Thanks! A few comments/questions.

templates/liar/templates.yaml
manandey: Actually, there is no finite set of party affiliations to choose from. So it is fine to have it the way it is?

templates/liar/templates.yaml
manandey: Sure, I will remove `===` from some of the prompts.

templates/liar/templates.yaml
manandey: There are no options for providing choices for the context here as well. So should we keep the prompt the way it is?

templates/liar/templates.yaml
craffel: Yep, that's fine.

templates/liar/templates.yaml
craffel: Sounds good.

templates/liar/templates.yaml
manandey: Updated! @craffel

templates/liar/templates.yaml
```diff
@@ -0,0 +1,75 @@
...
+    id: c235a60e-5604-4fad-b020-df7f333bdac5
+    jinja: 'Given the following statement:
+
+
+      {{statement}}
+
+
+      To what subject would you categorize it? |||
+
+      {{subject}}'
```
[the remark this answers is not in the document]
manandey: By changing it to `{{subject | split(",") | choice}}`, I am getting `TemplateAssertionError: no filter named 'split'`

manandey:
Hi @craffel! I have opened a new PR #218 where I have updated the code according to your suggestions. Feel free to close this PR anytime.

PR #88, opened 2021-06-07
Prompts for `quoref`

manandey:


templates/quoref/templates.yaml
```diff
@@ -0,0 +1,70 @@
+dataset: quoref
+templates:
+  6f1d5031-1377-4b8a-9475-987b2275b8da: !Template
+    id: 6f1d5031-1377-4b8a-9475-987b2275b8da
+    jinja: 'Given the following, context:
+
+      {{context}}
+
+      ===
```
craffel, line 9: Same comment from your other PR about ===.

templates/quoref/templates.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+      Predict the start index for the answer to the following question in the above
+      context!
+
+      {{question}} |||
+
+      {{answers.answer_start}}'
```
craffel, line 44: When we tried this with T5 (predicting the index of the answer rather than the answer itself) it didn't really work (because of tokenization). I'd hesitate to have the model predict a location like this. Let me ask what people think in the slack.

craffel (changes requested):

templates/quoref/templates.yaml
manandey: Sure, I will update the prompts.

templates/quoref/templates.yaml
manandey: Sure, do let me know in case I would need to make any changes here.

templates/quoref/templates.yaml
manandey: Updated! @craffel

templates/quoref/templates.yaml
manandey: Leaving this as it is for the moment based on your confirmation yesterday.

PR #89, opened 2021-06-07
copying prompts to `winogrande` subsets

manandey:


templates/winogrande/winogrande_debiased/templates.yaml
```diff
@@ -0,0 +1,47 @@
...
+    reference: Replace with correct option
+  13b2dbe4-abf3-4b09-b7cb-459224881844: !Template
+    id: 13b2dbe4-abf3-4b09-b7cb-459224881844
+    jinja: 'The _ in the sentence below refers to {{option1}}. Answer True or False!
+
+      {{sentence}}|||
+
+      {% if answer== 1 %}True{% else%}False{% endif %}'
+    name: True or False
+    reference: True or False Type
```
craffel, line 19: Can you point me to where these came from? The reference field is for citing where the prompt originally was proposed, not for a description, so we should remove these.

craffel (changes requested):

templates/winogrande/winogrande_debiased/templates.yaml
manandey: Remove the descriptions from the reference fields.

PR #90, opened 2021-06-07
multi_x_science_sum

arunraja-hub:
Prompts for multi_x_science_sum

### DIFF
--- a/templates/multi_x_science_sum/templates.yaml
+++ b/templates/multi_x_science_sum/templates.yaml
@@ -1,0 +1,52 @@
+dataset: multi_x_science_sum
+templates:
+  2bca0197-e3d4-4870-bd95-178411e52e09: !Template
+    id: 2bca0197-e3d4-4870-bd95-178411e52e09
+    jinja: 'Use the reference abstracts to generate related work:
+
+
+      {{ref_abstract.values()}}  |||
+
+      {{related_work}}'
+    name: ref_relatedwork
+    reference: ''
+  3bd082cb-4e28-4eb7-9fa2-dd03f1f86219: !Template
+    id: 3bd082cb-4e28-4eb7-9fa2-dd03f1f86219
+    jinja: 'Given the abstract of a paper, provide some related work for readers to
+      learn further
+
+      {{abstract}} |||
+
+      {{related_work}}'
+    name: abstract_relatedwork
+    reference: ''
+  af4d550e-54b8-471e-97af-2b2c50a1382e: !Template
+    id: af4d550e-54b8-471e-97af-2b2c50a1382e
+    jinja: 'Use the related work to guess the abstract:
+
+
+      {{related_work}} |||
+
+      {{abstract}}'
+    name: relatedwork_abstract
+    reference: ''
+  b9eb67b4-415b-4a38-a90a-9dee3ae385d7: !Template
+    id: b9eb67b4-415b-4a38-a90a-9dee3ae385d7
+    jinja: 'Use the abstract and reference abstracts to generate related work:
+
+
+      {{abstract}} , {{ref_abstract.values()}}  |||
+
+      {{related_work}}'
+    name: related_work
+    reference: ''
+  f59d7e5a-5982-467f-b451-91154e311666: !Template
+    id: f59d7e5a-5982-467f-b451-91154e311666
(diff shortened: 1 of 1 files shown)


templates/multi_x_science_sum/templates.yaml
```diff
@@ -0,0 +1,52 @@
...
+    name: relatedwork_abstract
+    reference: ''
+  b9eb67b4-415b-4a38-a90a-9dee3ae385d7: !Template
+    id: b9eb67b4-415b-4a38-a90a-9dee3ae385d7
+    jinja: 'Use the abstract and reference abstracts to generate related work:
+
+
+      {{abstract}} , {{ref_abstract.values()}}  |||
+
+      {{related_work}}'
```
craffel, line 40: Does including the reference abstracts make it too easy to predict the related work? Also, does abstract + ref_abstract.values() result in a really long input sequence?

arunraja-hub: I feel that this task does not seem too easy as the reference abstracts, though they might convey the similar/same information as related work, are phrased in different words to that of related work. So the model would be challenged to understand the meaning at some level. I took a look at the prompt dataset viewer and they do not look too long for the most part

templates/multi_x_science_sum/templates.yaml
```diff
@@ -0,0 +1,52 @@
+dataset: multi_x_science_sum
+templates:
+  2bca0197-e3d4-4870-bd95-178411e52e09: !Template
+    id: 2bca0197-e3d4-4870-bd95-178411e52e09
+    jinja: 'Use the reference abstracts to generate related work:
+
+
+      {{ref_abstract.values()}}  |||
```
craffel, line 8: This is showing up like this in my prompt viewer:
```
dict_values([['This note is a sequel to our earlier paper of the same title [4]
and describes invariants of rational homology 3-spheres associated to acyclic
orthogonal local systems. Our work is in the spirit of the Axelrod–Singer papers
[1], generalizes some of their results, and furnishes a new setting for the
purely topological implications of their work.', 'Recently, Mullins calculated
the Casson-Walker invariant of the 2-fold cyclic branched cover of an oriented
link in S^3 in terms of its Jones polynomial and its signature, under the
assumption that the 2-fold branched cover is a rational homology 3-sphere. Using
elementary principles, we provide a similar calculation for the general case. In
addition, we calculate the LMO invariant of the p-fold branched cover of twisted
knots in S^3 in terms of the Kontsevich integral of the knot.'], ['@cite_16',
'@cite_26'], ['1481005306', '1641082372']])
```

I think `{{ref_abstract["abstract"]}}` probably makes more sense, WDYT?

templates/multi_x_science_sum/templates.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+      {{related_work}}'
+    name: related_work
+    reference: ''
+  f59d7e5a-5982-467f-b451-91154e311666: !Template
+    id: f59d7e5a-5982-467f-b451-91154e311666
+    jinja: 'Use reference papers to guess the abstract:
+
+
+      {{ref_abstract.values()}} |||
```
craffel, line 48: Same comment about .values()

templates/multi_x_science_sum/templates.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+      {{abstract}}'
+    name: relatedwork_abstract
+    reference: ''
+  b9eb67b4-415b-4a38-a90a-9dee3ae385d7: !Template
+    id: b9eb67b4-415b-4a38-a90a-9dee3ae385d7
+    jinja: 'Use the abstract and reference abstracts to generate related work:
+
+
+      {{abstract}} , {{ref_abstract.values()}}  |||
```
craffel, line 38: Same comment about .values()

craffel (changes requested):

templates/multi_x_science_sum/templates.yaml
arunraja-hub: Since all reference abstracts will be displayed with `dict_values([[...` I feel that the model wont develop a particular bias towards this. However, I have the made the change as suggested

craffel:
Thanks!

PR #91, opened 2021-06-07
financial_phrasebank

arunraja-hub:


templates/financial_phrasebank/sentences_allagree/templates.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+
+      {{sentence}} |||
+
+      {{["growth","neutral","decline"][label]}}'
+    name: complementary_industries
+    reference: ''
+  5fa16d31-b513-480d-bd1b-1fa8c182fb76: !Template
+    id: 5fa16d31-b513-480d-bd1b-1fa8c182fb76
+    jinja: 'Should an investor be bullish or bearish given the following detail?
```
craffel, line 26: Replace "bullish or bearish" with `{{"bullish, neutral, or bearish"}}` since we need to protect the label words, and you should include neutral as a possible label word.

craffel (changes requested):
Can you create variants of these prompts where you include the label words? I'm not sure how well the model will be able to guess the set of label words.

templates/financial_phrasebank/sentences_allagree/templates.yaml
arunraja-hub: Thank you! I have made the replacement and the output already contains "neutral"

craffel:
Merged in #118. Thanks!

PR #92, opened 2021-06-07
Circa

arunraja-hub:


templates/circa/templates.yaml
```diff
@@ -0,0 +1,55 @@
...
+      |||
+
+      {{question_X}}'
+    name: possible_qn
+    reference: ''
+  997f7f96-d420-48c1-85f7-ecade54adbd7: !Template
+    id: 997f7f96-d420-48c1-85f7-ecade54adbd7
+    jinja: "What is the most common judgement given the question-answer pair of X\
+      \ and Y in the context of {{context}} ?\n\nX: {{question_X}} \nY: {{answer_Y}}\
+      \ |||\n\n{{goldstandard1}}"
```
VictorSanh, line 34: You need to map `goldstandard1` to a string

templates/circa/templates.yaml
```diff
@@ -0,0 +1,55 @@
...
+    name: judgement
+    reference: ''
+  a15c1a30-5ef0-451f-b202-987a16752a0a: !Template
+    id: a15c1a30-5ef0-451f-b202-987a16752a0a
+    jinja: "What is the most common judgement given the question-answer pair of X\
+      \ and Y in the context of {{context}} ?\n\nX: {{question_X}} \nY: {{answer_Y}}\
+      \ |||\n\n{{   [\n  \"Yes\",\n  \"No\",\n  \"In the middle, neither yes nor no\"\
+      ,\n  \"Probably yes / sometimes yes\",\n  \"Probably no\",\n  \"Yes, subject\
+      \ to some conditions\",\n  \"Other\",\n  \"I am not sure how X will interpret\
+      \ Y\u2019s answer\"\n] [goldstandard1]}}"
```
VictorSanh, line 44: Wrap the whole template into an if condition: `{% if goldstandard1 != -1 %} template {% endif %} otherwise there will be errors introduced when the label is missing

templates/circa/templates.yaml
```diff
@@ -0,0 +1,55 @@
...
+    name: goldstandard1_judgement
+    reference: ''
+  e09ed91b-694b-454c-be26-38a291c3ff75: !Template
+    id: e09ed91b-694b-454c-be26-38a291c3ff75
+    jinja: 'Answer the following question given the {{context}} answer the following:
```
VictorSanh, line 49: - "Answer the following" is repeated.
- I am worried about this template. We are asking the model to react to a question without any grounding/information about person Y. This is basically a hallucination. We should remove this one

templates/circa/templates.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+      {{canquestion_X}}'
+    name: question_declarative
+    reference: ''
+  70b7a94a-6a39-4a81-9a6e-0709a0acdb28: !Template
+    id: 70b7a94a-6a39-4a81-9a6e-0709a0acdb28
+    jinja: "What is the most common judgement in a less strict way given the question-answer\
+      \ pair of X and Y in the context of {{context}}?\n\nX: {{question_X}} \nY: {{answer_Y}}\
+      \ |||\n\n{{   \n [\n  \"Yes\",\n  \"No\",\n  \"In the middle, neither yes nor\
+      \ no\",\n  \"Yes, subject to some conditions\",\n  \"Other\"\n]\n [goldstandard2]\
```
VictorSanh, line 18: Wrap the whole template into an if condition: `{% if goldstandard2 != -1 %} template {% endif %} otherwise there will be errors introduced when the label is missing

templates/circa/templates.yaml
```diff
@@ -0,0 +1,55 @@
...
+      \ }}"
+    name: goldstandard2_judgement
+    reference: ''
+  73466d0f-b1b1-4c61-8f03-346e121ae06c: !Template
+    id: 73466d0f-b1b1-4c61-8f03-346e121ae06c
+    jinja: 'What is a possible question to ask someone given the context of {{context}}?
```
VictorSanh, line 24: Precise that it's X asking the question, otherwise there is an uncertainty about who's asking the question for a bunch of examples

templates/circa/templates.yaml
```diff
@@ -0,0 +1,55 @@
...
+  997f7f96-d420-48c1-85f7-ecade54adbd7: !Template
+    id: 997f7f96-d420-48c1-85f7-ecade54adbd7
+    jinja: "What is the most common judgement given the question-answer pair of X\
+      \ and Y in the context of {{context}} ?\n\nX: {{question_X}} \nY: {{answer_Y}}\
+      \ |||\n\n{{goldstandard1}}"
+    name: judgement
+    reference: ''
+  a15c1a30-5ef0-451f-b202-987a16752a0a: !Template
+    id: a15c1a30-5ef0-451f-b202-987a16752a0a
+    jinja: "What is the most common judgement given the question-answer pair of X\
```
VictorSanh, line 39: When you are expecting a generation among a couple of different options, you should list and protect them (see [here](https://github.com/bigscience-workshop/promptsource/blob/main/CONTRIBUTING.md#best-practices)).

Also, it is not clear what judgement you querying the model for (i.e. judgment of how to understand Y's answer)

Same comment for template `goldstandard2_judgement`

VictorSanh (changes requested):
A few comments (please see inline)

templates/circa/templates.yaml
arunraja-hub: Sorry I dont get this. Please help to clarify
goldstandard1 is the index for judgements in the list: [
  "Yes",
  "No",
  "In the middle, neither yes nor no",
  "Probably yes / sometimes yes",
  "Probably no",
  "Yes, subject to some conditions",
  "Other",
  "I am not sure how X will interpret Y’s answer"
]

VictorSanh: Right, but here you are asking the model to generate the index rather than the string

arunraja-hub: It seems to generate the string itself such as "Yes" and "No". You can take a look at the Prompted dataset viewer.
image

arunraja-hub: sorry I got confused between goldstandard1 and this
Corrected now!

VictorSanh (approved):
lgtm!

The template `judgement` also calls `goldstandard1`, can you wrap it around an if condition (same as you did before).

I'll merge after you commit this change, don't forget to update to spreadsheet!

VictorSanh:
Thank you @arunraja-hub, merging now!

PR #93, opened 2021-06-07
Guardian authorship

arunraja-hub:


templates/guardian_authorship/cross_topic_1/templates.yaml
```diff
@@ -0,0 +1,54 @@
...
+    jinja: "Who could have authored this article based on the writing style?\n\n{{article}}\
+      \ |||\n{{\n[\n  \"catherinebennett\",\n  \"georgemonbiot\",\n  \"hugoyoung\"\
+      ,\n  \"jonathanfreedland\",\n  \"martinkettle\",\n  \"maryriddell\",\n  \"nickcohen\"\
+      ,\n  \"peterpreston\",\n  \"pollytoynbee\",\n  \"royhattersley\",\n  \"simonhoggart\"\
+      ,\n  \"willhutton\",\n  \"zoewilliams\"\n][author]\n}}"
+    name: writing_style
+    reference: ''
+  12982397-c0c3-49a9-b3ac-38735908428b: !Template
+    id: 12982397-c0c3-49a9-b3ac-38735908428b
+    jinja: "Generate an article based on the writing style of  {{\n[\n  \"catherinebennett\"\
```
VictorSanh, line 15: duplicate whitespace `style of  {{`

VictorSanh (approved):
LGTM modulo small typo!

I see a bunch of other subsets (cross_genre_*, cross_topic_*). Do you think there is value in adding them too?
I scrolled through a couple of examples and they seem to be different per subset (but you should verify this more carefully)

arunraja-hub:
Most of the examples in other subsets seem to be the same but there are a few different/extra examples in other subsets as well. In terms of format, they are all the same so the same prompts could be used for all subsets. Should I copy over the templates for other subsets as well?

VictorSanh:
> Most of the examples in other subsets seem to be the same but there are a few different/extra examples in other subsets as well. In terms of format, they are all the same so the same prompts could be used for all subsets. Should I copy over the templates for other subsets as well?

I'd recommend copy-pasting the templates over to subsets for which you are sure there are different enough. Genre and topic seem different enough but you know the dataset better than I do

VictorSanh:
@arunraja-hub  Let me know once you have copied the templates for a few different subsets, I'll merge your PR then!

PR #96, opened 2021-06-07
Banking77

arunraja-hub:


VictorSanh:
LGTM, thanks, these are quite long lists of topics!

Could you add one or two variants of the same templates with all the choices listed in the prompt (asking the model to reason over the instructions without having to learn all the labels). You can use `{% set l = [1,2,3] %}` to define a list variable

VictorSanh:
Merging now, please update the spreadsheet!

PR #98, opened 2021-06-07
Discovery

arunraja-hub:


stephenbach (changes requested):
Thank you! For the label names like `in_sum`, would you please rewrite them to be English words (no underscores)?

arunraja-hub:
Thank you, I have removed the underscores

stephenbach:
Thanks, would you also change the `[no-conn]` label to something more natural like `No connection` or `None`? In general, the prompts and outputs should be natural language that complements the text that's collect "from the wild."

stephenbach (approved):
LGTM, thanks! Please mark as completed on spreadsheet.

PR #101, opened 2021-06-07
Templates for rotten tomatoes

Urmish:
Copy pasted imdb templates for rotten tomatoes. The tempplate name still use the format imdb_, but that does not impact the template


srush:
I think it would be good to update the names.

Urmish:
Sure. Let me finish my work day, I will get to this in the evening and ping you when its done :)

Urmish:
@srush - Would it be possible to get a review for this again? Changed names, added another prompt with a bit more diversity

PR #105, opened 2021-06-08
Tests update

VictorSanh:
Small tweaks in the test suite
- update imports (no need to update PYTHONPATH)
- move show_updates to test suites folder for GH action consistency

### DIFF
--- a/.github/workflows/check_templates.yml
+++ b/.github/workflows/check_templates.yml
@@ -18,6 +18,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
+          pip install .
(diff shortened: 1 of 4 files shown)


srush:
Can you also add a check if there is no y. Forgot about that case.

VictorSanh:
> Can you also add a check if there is no y. Forgot about that case.

Yup, I fixed that case late last night in this commit (https://github.com/bigscience-workshop/promptsource/commit/9eac55422a69d6d28a547464c9883e1bbacfb37b).

srush:
Oh nice. Then I think you just need to package promptsource with a `__init__.py` ?

PR #108, opened 2021-06-08
DuoRC

gchhablani:
Adding a few templates for DuoRC.

**Notes**: 
- Some questions may not have answers, which I have handled. But the `generate_questions` and `generate_questions_by_answer` may give out similar results due to this. This needs to be kept in mind.
- There are some questions with multiple answers. I am only using the first answer to generate a prompt in that case. We can alternatively run some kind of for loop for this.

Please let me know your thoughts.
Thanks.

### DIFF
--- a/templates/duorc/ParaphraseRC/templates.yaml
+++ b/templates/duorc/ParaphraseRC/templates.yaml
@@ -3,7 +3,7 @@ subset: ParaphraseRC
 templates:
   1e6e6a07-4655-4dd8-a694-373bd3049698: !Template
     id: 1e6e6a07-4655-4dd8-a694-373bd3049698
-    jinja: "{{plot}} {{question}}
+    jinja: '{{plot}} {{question}}
 
       |||
 
@@ -17,7 +17,7 @@ templates:
 
       {% endif %}
 
-      "
+      '
     name: question_answering
     reference: Given a passage and a question, generate an answer.
   840f6409-4273-4968-9a17-a0c300e9412f: !Template
@@ -29,10 +29,32 @@ templates:
       {{ question }}'
     name: generate_question
     reference: Given the plot, generate a question.
+  9a99cb83-3b24-4516-b952-25cd70ca0f78: !Template
+    id: 9a99cb83-3b24-4516-b952-25cd70ca0f78
+    jinja: 'Is the following question answerable:
+
+      "{{question}}"
+
+      from the given plot:
+
+      "{{plot}}"
+
+      |||
+
+      {% if no_answer %}
+
+      Yes
+
+      {% else %}
+
+      No
+
+      {% endif %}'
+    name: answerable_question
+    reference: Given the plot and the question, determine if the question is answerable.
   e2d42f3f-d835-4127-a93e-2e8e257590eb: !Template
     id: e2d42f3f-d835-4127-a93e-2e8e257590eb
-    jinja:
-      "{% if no_answer %}\n Build a story around this: \"{{ question }}\".\n\
+    jinja: "{% if no_answer %}\n Build a story around this: \"{{ question }}\".\n\
       {% else %}\nBuild a story around this: \"{{ question }} {{answers[0]}}\".\n\
       {% endif %}\n|||\n{{ plot }}"
     name: build_story_around_qa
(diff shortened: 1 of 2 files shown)


templates/duorc/ParaphraseRC/templates.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+      |||
+
+      {{ question }}'
+    name: generate_question
+    reference: Given the plot, generate a question.
+  e2d42f3f-d835-4127-a93e-2e8e257590eb: !Template
+    id: e2d42f3f-d835-4127-a93e-2e8e257590eb
+    jinja: "{% if no_answer %}\n Build a story around this: \"{{ question }}\".\n\
+      {% else %}\nBuild a story around this: \"{{ question }}{{answers[0]}}\".\n{%\
```
stephenbach, line 54: Should put a space between question and answer

templates/duorc/SelfRC/templates.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+      |||
+
+      {{ question }}'
+    name: generate_question
+    reference: Given the plot, generate a question.
+  e2d42f3f-d835-4127-a93e-2e8e257590eb: !Template
+    id: e2d42f3f-d835-4127-a93e-2e8e257590eb
+    jinja: "{% if no_answer %}\n Build a story around this: \"{{ question }}\".\n\
+      {% else %}\nBuild a story around this: \"{{ question }}{{answers[0]}}\".\n{%\
```
stephenbach, line 54: Same as other subset, need a space.

templates/duorc/ParaphraseRC/templates.yaml
```diff
@@ -0,0 +1,85 @@
...
+      {% else %}
+
+      Yes
+
+      {% endif %}'
+    name: answerable_question
+    reference: Given the plot and the question, determine if the question is answerable.
+  1e6e6a07-4655-4dd8-a694-373bd3049698: !Template
+    id: 1e6e6a07-4655-4dd8-a694-373bd3049698
+    jinja: '{{plot}}{{question}}
```
stephenbach, line 25: Space between plot and question.

templates/duorc/SelfRC/templates.yaml
```diff
@@ -0,0 +1,85 @@
...
+      {% else %}
+
+      Yes
+
+      {% endif %}'
+    name: answerable_question
+    reference: Given the plot and the question, determine if the question is answerable.
+  1e6e6a07-4655-4dd8-a694-373bd3049698: !Template
+    id: 1e6e6a07-4655-4dd8-a694-373bd3049698
+    jinja: '{{plot}}{{question}}
```
stephenbach, line 25: Space between plot and question.

gchhablani:
Hi @stephenbach 

Thanks for reviewing. I'll be making the changes asap.

For the % ratio between answerable non-answerable, from the [paper](https://arxiv.org/pdf/1804.07927.pdf):

**SelfRC**:
```
We found
that in 70% of the cases the workers picked an answer directly from the document and in 30% of
the cases they synthesized the answer. 
```
**ParaphraseRC**:
```
We found that in 50% of the cases the workers selected
an answer which matched a span in the document,
whereas in 37% cases they synthesized the answer
and in 13% cases they said that question was not
answerable.
```

I think, looking at this, maybe the `answerable_questions` template can be removed. Wdyt?

stephenbach:
Ah, that's why. I think I only looked at SelfRC for this question. I would take it out for SelfRC but leave it for ParaphraseRC. 13% isn't so terrible that it might not learn anything, especially since it's a useful skill.

Also, looking at how long `plot` can be, I would make sure that it's always the last thing in the input or output, since it might get truncated.

gchhablani:
I removed it from both, and added it back to ParaphraseRC :P 
I'm not sure if the `"`s around the plot and other keys are fine. Let me know if that should be changed as well.

stephenbach:
Thanks for the updates!
- Would you rearrange the templates so that `{{plot}}` always comes at the end? Safest that way with possible truncation. For examples, there's one that is `{{plot}} {{question}}` that you could make `{{question}} {{plot}}`.
- The template named `answerable_question` has the labels backwards. It says it is answerable if `no_answer` is True.
- You could keep or remove the quotes. I err on the side of removing them since the plots usually have lots of quotations. Seems unlikely that they are meaningful delimiters.

stephenbach:
Beautiful! Thanks for all the polishing!

stephenbach (approved):

gchhablani:
Thanks for reviewing and approving  :)

PR #110, opened 2021-06-08
Codah

arnaudstiegler:
All subsets are similar so copy/pasted the template for all subsets.

Note that `{{question_propmt}}` in the template is not a typo (the typo is in the hf.datasets definition)

3 types of prompts:
- answer without being given the options
- answer with the options provided
- Find the question category (only 2 prompts on that, seems like there's quite an imbalance between classes so might not be super interesting)


srush:
Looks great!

PR #111, opened 2021-06-08
Prompts for `climate_fever`

drugilsberg:
Created five template prompt to evaluate claim/evidence pairs from the dataset.
One main note.
Looking at the schema from HuggingFace datasets (https://huggingface.co/datasets/climate_fever#data-fields), the labels should be indexed as follows:
- claim_label: a int feature, overall label assigned to claim (based on evidence majority vote). The label correspond to 0: "refutes", 1: "supports" and 2: "not enough info"
- evidence_label: a int feature, micro-verdict label. The label correspond to 0: "refutes", 1: "supports" and 2: "not enough info"
But after some inspection I realized that 0 is usually associated to `SUPPORTS` and 1 is associated to `REFUTES`.
The template I created reflects this.

### DIFF
--- a/templates/climate_fever/templates.yaml
+++ b/templates/climate_fever/templates.yaml
@@ -1,0 +1,87 @@
+dataset: climate_fever
+templates:
+  82c484bd-2ed7-4ee0-aaee-2b31ac68e751: !Template
+    id: 82c484bd-2ed7-4ee0-aaee-2b31ac68e751
+    jinja: 'Considering the following claim:
+
+      {{claim}}.
+
+      Does the following statement {{"supports"}}, {{"refutes"}}, or provide {{"not
+      enough info"}} on climate change?
+
+      {{evidences[4]["evidence"].strip(".").strip(''"'')}}.
+
+      |||
+
+      {{ ["SUPPORTS", "REFUTES", "NOT_ENOUGH_INFO"][evidences[4]["evidence_label"]]
+      }}'
+    name: fifth_evidence_claim_pair
+    reference: Relation between the claim and fifth evidence pair.
+  cb78a363-fd32-4dbd-976f-b56de644ba90: !Template
+    id: cb78a363-fd32-4dbd-976f-b56de644ba90
+    jinja: 'Considering the following claim:
+
+      {{claim}}.
+
+      Does the following statement {{"supports"}}, {{"refutes"}}, or provide {{"not
+      enough info"}} on climate change?
+
+      {{evidences[1]["evidence"].strip(".").strip(''"'')}}.
+
+      |||
+
+      {{ ["SUPPORTS", "REFUTES", "NOT_ENOUGH_INFO"][evidences[1]["evidence_label"]]
+      }}'
+    name: second_evidence_claim_pair
+    reference: Relation between the claim and second evidence pair.
+  cca7b6f5-29e3-45a4-bc8b-889f5ab2fc13: !Template
+    id: cca7b6f5-29e3-45a4-bc8b-889f5ab2fc13
+    jinja: 'Considering the following claim:
+
+      {{claim}}.
+
+      Does the following statement {{"supports"}}, {{"refutes"}}, or provide {{"not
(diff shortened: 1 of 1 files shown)


drugilsberg:
hey @VictorSanh what are the next steps, should I add some more details or are there any changes I should work on?

templates/climate_fever/templates.yaml
```diff
@@ -0,0 +1,87 @@
...
+      {{claim}}.
+
+      Does the following statement {{"supports"}}, {{"refutes"}}, or provide {{"not
+      enough info"}} on climate change?
+
+      {{evidences[4]["evidence"].strip(".").strip(''"'')}}.
+
+      |||
+
+      {{ ["SUPPORTS", "REFUTES", "NOT_ENOUGH_INFO"][evidences[4]["evidence_label"]]
```
VictorSanh, line 16: Could you please change it to ["Supports", "Refutes", "Not enough information"]?
same for the other templates

VictorSanh (changes requested):
Yes, i think you are right, can you open a PR on `datasets` to correct this? + I also see that `claim_label` can have `3` as a value (Disputed)... I don't see disputed being used in the individual evidences though.

This looks great thank you!

Could you add some more templates?
For instance, one could be:
```jinja2
Here's a statement and accompanying evidence. Does the evidence {{"supports"}}, {{"refutes"}}, or provide {{"not enough info"}} on climate change?
Statement: {{claim}}
Evidence: {{evidence[...]}}|||
your_label
```

templates/climate_fever/templates.yaml
drugilsberg: Sure will do

drugilsberg:
> > Yes, i think you are right, can you open a PR on `datasets` to correct this? + I also see that `claim_label` can have `3` as a value (Disputed)... I don't see disputed being used in the individual evidences though.
> > This looks great thank you!
> > Could you add some more templates?
> > For instance, one could be:
> > ```
> > Here's a statement and accompanying evidence. Does the evidence {{"supports"}}, {{"refutes"}}, or provide {{"not enough info"}} on climate change?
> > Statement: {{claim}}
> > Evidence: {{evidence[...]}}|||
> > your_label
> > ```
> 
> Exactly it is not used for the evidence. I will also open the PR on `datasets` and add some more templates.

I added some new templates and fixes the labeling:
- paired templates as suggested in the PR review.
- a template considering all evidence statements together.

Will now open the PR on `datasets`.

drugilsberg:
@VictorSanh As promised here the PR in `datasets`: https://github.com/huggingface/datasets/pull/2464.

templates/climate_fever/templates.yaml
```diff
@@ -1,5 +1,61 @@
...
+
+      |||
+
+      {{ ["Supports", "Refutes", "Not enough information"][evidences[0]["evidence_label"]]
+      }}'
+    name: first_evidence_and_claim_itemization
+    reference: First evidence and claim with simple itemization
+  3970f474-a9e3-4264-aefa-dd4cfadd279c: !Template
+    id: 3970f474-a9e3-4264-aefa-dd4cfadd279c
+    jinja: 'Here''s a claim and accompanying evidence statements . Do the statements
```
VictorSanh, line 22: that's the part i am confused about. Each evidence can separately refute, support, etc. the claim.
So how do you create consensus between all evidences (i.e. what is `claim_label`?)
Is it some sort of "do the evidence mostly support the claim?" or "mostly refute the claim?"

drugilsberg: The claim label per documentation should be assigned with a majority vote over the evidence statements, if I understood correctly . Not sure what is the the exact procedure used to break ties though.

drugilsberg: I decided to blindly trust the claim_label field 😅but if you think it's better to ignore this, we can simply remove this template.

VictorSanh (approved):
merging now, please update the spreadsheet!

PR #112, opened 2021-06-08
CBT

gchhablani:
This PR adds templates for CBT dataset: https://huggingface.co/datasets/cbt.

### DIFF
--- a/templates/cbt/CN/templates.yaml
+++ b/templates/cbt/CN/templates.yaml
@@ -3,10 +3,7 @@ subset: CN
 templates:
   0725fe5e-1bba-4e08-a448-9e0038164914: !Template
     id: 0725fe5e-1bba-4e08-a448-9e0038164914
-    jinja: '{{sentences | join('''')}}
-
-
-      Write the next sentence of this story.
+    jinja: 'Write the next sentence of this story: {{sentences | join('''')}}
 
       |||
 
@@ -15,29 +12,20 @@ templates:
     reference: Generate the next sentence given the story.
   2c326181-dbba-401e-accb-d84ea0162f0a: !Template
     id: 2c326181-dbba-401e-accb-d84ea0162f0a
-    jinja: "{{ sentences | join('') }}
+    jinja: "Read the passage and fill in the XXXXX:
 
-      {{question}}
-
-      Which part-of-speech fill replaces XXXXX the best?
+      {{ sentences | join('') }} {{question}}
 
       |||
 
-      Common Noun"
-    name: pos_prediction
-    reference: Determine which part of speech will best fit in the blank.
+      {{ answer }}"
+    name: answer_prediction
+    reference: Fill in the blank without options.
   b26cae56-1fbd-47a5-8c8d-d981ca098239: !Template
     id: b26cae56-1fbd-47a5-8c8d-d981ca098239
-    jinja: '{{sentences | join ('''')}} {{question}}
-
-
-      Which of the following options replaces XXXXX the best?
-
-
-      {{ options | join (", ") }}
-
-      |||
-
-      {{ answer }}'
+    jinja:
+      "Which of the following options replaces XXXXX the best?\n{{ options |\
+      \ join (\", \") }}\nin this story: \n{{sentences | join ('')}} {{question}}\n\
+      |||\n{{ answer }}"
     name: multi_choice
     reference: Given the sentences, fill the blanks using the options.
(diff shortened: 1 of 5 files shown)


templates/cbt/raw/templates.yaml
```diff
@@ -0,0 +1,24 @@
+dataset: cbt
+subset: raw
+templates:
+  2d9e9c74-550e-4838-8d1d-a804d74828f7: !Template
```
VictorSanh, line 4: Can you remove this template? it's not super interesting from the perspective of supervision (very LM-like)

templates/cbt/V/templates.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+      Verb"
+    name: pos_prediction
+    reference: Determine which part of speech will best fit in the blank.
+  b26cae56-1fbd-47a5-8c8d-d981ca098239: !Template
+    id: b26cae56-1fbd-47a5-8c8d-d981ca098239
+    jinja: '{{sentences | join ('''')}} {{question}}
+
+
+      Which of the following options replaces XXXXX the best?
```
VictorSanh, line 34: Can you put the task description first? It will help when truncating to not truncate the task description.
Will need some marginal wording reformulation

VictorSanh: same for pos_prediction

templates/cbt/CN/templates.yaml
```diff
@@ -0,0 +1,43 @@
...
+    id: 2c326181-dbba-401e-accb-d84ea0162f0a
+    jinja: "{{ sentences | join('') }}
+
+      {{question}}
+
+      Which part-of-speech fill replaces XXXXX the best?
+
+      |||
+
+      Common Noun"
```
VictorSanh, line 26: missing example-dependent label

templates/cbt/P/templates.yaml
```diff
@@ -0,0 +1,43 @@
...
+    id: 2c326181-dbba-401e-accb-d84ea0162f0a
+    jinja: '{{ sentences | join('''') }}
+
+      {{question}}
+
+      Which part-of-speech fill replaces XXXXX the best?
+
+      |||
+
+      Pronoun'
```
VictorSanh, line 26: idem

templates/cbt/V/templates.yaml
```diff
@@ -0,0 +1,43 @@
...
+    id: 2c326181-dbba-401e-accb-d84ea0162f0a
+    jinja: "{{ sentences | join('') }}
+
+      {{question}}
+
+      Which part-of-speech fill replaces XXXXX the best?
+
+      |||
+
+      Verb"
```
VictorSanh, line 26: same

templates/cbt/NE/templates.yaml
```diff
@@ -0,0 +1,43 @@
...
+    id: 2c326181-dbba-401e-accb-d84ea0162f0a
+    jinja: "{{ sentences | join('') }}
+
+      {{question}}
+
+      Which part-of-speech fill replaces XXXXX the best?
+
+      |||
+
+      Proper Noun"
```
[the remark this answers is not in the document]
gchhablani: This task I thought could be something like PoS prediction and the model should have some minimal PoS knowledge? Otherwise we can just put the answer word here without any options? Wdyt?

VictorSanh:
> Hi @VictorSanh
> Thanks for reviewing.
> I agree with you. The PoS type template is not fully dependent on the example and is almost similar to language modeling. Even if I add the answer word, it will be similar to MLM. Not sure if that is preferred?

As I was saying in the comments, we shouldn't have predictions that are not using the specific example in some way. I'd recommend removing these templates.

> Writing the story given the title is also not a very interesting task. I added a similar task in #108 also. Should I remove it there as well?

I looped with the others, there is a consensus we should include these templates (`generate story from {{title}}`) so please ignore my comments on these!

> The subsets are different only in the type of work that is masked - Named Entity, Common Noun, Pronoun, and Verb. Maybe we need to have something different for each? Wdyt?

Yup agree! modulo comment on example-independent prediction

templates/cbt/raw/templates.yaml
VictorSanh: please ignore

VictorSanh (approved):

VictorSanh:
merged! don't forget to update the spreadsheet!

PR #113, opened 2021-06-08
Prompt proto_qa dataset 

zaidalyafeai:


srush:
Thanks. This  is sms_spam though not proto_qa right?

zaidalyafeai:
> Thanks. This is sms_spam though not proto_qa right?

Oh, sorry, you are right. I was working on `proto_qa` but faced an error so I switched to `sms_spam`.

zaidalyafeai:
Here is a screenshot of the error 
image

PR #116, opened 2021-06-08
following colin's sugg. Removing reverse (cyclic) prompt augmetation.

sbmaruf:
Earlier I added generation samples by, A -> B and then B -> A.
As suggested by @craffel, this might create a memorization problem. So removing, duplicate pairs.

### DIFF
--- a/templates/paws-x/en/templates.yaml
+++ b/templates/paws-x/en/templates.yaml
@@ -1,12 +1,6 @@
 dataset: paws-x
 subset: en
 templates:
-  289448c3-63fc-4904-8387-86d232424243: !Template
-    id: 289448c3-63fc-4904-8387-86d232424243
-    jinja: "{% if label == 1 %} \nParaphrase the sentence: {{sentence2}} \n{% endif\
-      \ %}\n||| \n{{sentence1}} "
-    name: paraphrase-task-reverse
-    reference: Create a generative paraphrase task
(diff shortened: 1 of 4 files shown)


stephenbach:
Do we want to just keep this in for now? There are a couple of potential automated solutions (detecting and filtering automatically, or using one template for example). Might be easier than trying to go back and check the ones we've already done and also manually reviewing them as we go?

sbmaruf:
I am not sure what to do since I don't know how the augmentation and post-processing will work. I will rely on you for a decision. 
If you want to keep it, please let me know. I will close it.

stephenbach:
Right, sorry to not be clear, suggesting we discuss in a wider group. cc @craffel

srush:
Think we are going to keep these for now. At minimal we think we can detect them.

tianjianjiang:
My apologies for silly question, but may I ask how come this is related to augmentation?

sbmaruf:
@tianjianjiang Hi! as augmentation, I meant for supervised dataset, we will probably (not sure) create multiple samples from a single sample using diffrent temples.

tianjianjiang:
> I meant for supervised dataset, we will probably (not sure) create multiple samples from a single sample using diffrent prompt templates.
@sbmaruf Ah, I see, thank you for the clarification!

PR #117, opened 2021-06-08
Added opus_tedtalks translation templates

rbawden:


stephenbach:
Ah, sorry. We're doing English-only tasks for now. This one must have slipped past our filtering. Would you mark it as skipped because it contains non-English on the spreadsheet? Unless there's some English-only task that can be made with the data.

stephenbach:
Thank you for prompting another dataset! Sorry for the unclear setup, but would you mind opening a new PR for MWSC dataset? This way we can merge it in and leave this one open until we're ready for non-English datasets. Thanks for all your help!

rbawden:
Sorry, I didn't realise it was in the same one - I'll open a new one

PR #120, opened 2021-06-08
dataset dream

tae898:
Every sample in train / val / test splits of this dataset has exactly three choices. That's why I was able to hard code them as below.

```
choice[0]
choice[1]
choice[2]
```


tae898:
Got some comments from Sasha Rush.

srush:
These are fantastic. Thanks.

PR #121, opened 2021-06-08
Add prompts for `xsum`

JetRunner:


srush:
Can you also add reverse prompts. i.e. given the summary generate the document.

JetRunner:
@srush Yes I did that for `gigaword` (to be submitted) but I think we shouldn't do that for `xsum` since `xsum` has a very large `doc_len/summary_len` ratio.

srush:
okay, sounds good.

PR #122, opened 2021-06-08
Add a bunch of SQuAD prompts

craffel:


templates/squad/templates.yaml
```diff
@@ -0,0 +1,159 @@
...
+
+      |||
+
+
+      {{answers["text"][0]}}'
+    name: cbqa
+    reference: ''
+  46908111-143b-4ec2-9326-7959739569d8: !Template
+    id: 46908111-143b-4ec2-9326-7959739569d8
+    jinja: 'I''m creating a final exam for my reading class. Can you please come with
```
stephenbach, line 46: "come up with"

craffel: Nice catch, thanks

PR #123, opened 2021-06-08
Updating CONTRIBUTING.md with examples and other fixes

stephenbach:
This PR:
- adds a few more examples of nice templates
- clarifies that yes/no, true/false, etc. should not be escaped
- clarifies how blank outputs are handled (either side of `|||` can be whitespace, or the whole thing can be whitespace to be skipped)
- Adds a suggestion about dealing with long text in examples
- Fixes a bug in an example template where the choices were rendered as a Python list.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -180,16 +180,6 @@ Determine the relation between the following two sentences. The relations are en
 {{premise}}
 {{hypothesis}} ||| {{label}}
 ```
-* **Long text.** Some prompts will have to be truncated because of the model's
-maximum input size. You can better preserve the meaning of your template by
-putting a long field at the end. For example, consider replacing
-```jinja2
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -178,6 +180,73 @@ Determine the relation between the following two sentences. The relations are en
...
+Like above, it uses functions to present the choices in a readable way. Also, it
+uses a for loop with conditions to handle the more intricate dataset schema. 
+
+Here's one for `paws`:
+```jinja2
+{% if label == 0 or label == 1 %} 
+Sentence 1: {{sentence1}}
+Sentence 2: {{sentence2}}
+Question: Does Sentence 1 paraphrase Sentence 2? Yes or No?
+Yes
```
srush, line 237: Should this yes be in one of the labels below? Also maybe just {% else %} in this case?

PR #124, opened 2021-06-08
Finish ade-corpus-v2

tae898:


templates/ade_corpus_v2/Ade_corpus_v2_classification/templates.yaml
```diff
@@ -1,9 +1,43 @@
...
+    name: baseline-verbose
+    reference: ''
+  5398b646-fc60-4483-915c-9d1f0136412c: !Template
+    id: 5398b646-fc60-4483-915c-9d1f0136412c
+    jinja: Is "{{text}}" related to adverse drug effect (ADE)? ||| {{["Yes", "No"][label]}}
+    name: baseline-YN
+    reference: ''
+  56bd12a8-b8ee-464e-98cc-5f586ba9f74d: !Template
+    id: 56bd12a8-b8ee-464e-98cc-5f586ba9f74d
+    jinja: Is "{{text}}" related to adverse drug effect (ADE)? ||| {{["Not-Related",
```
VictorSanh, line 24: You can add `text.rstrip(".")` to your templates since I am under the impression that "text" is always ending with "."
I think that's fine if you are not adding these everywhere (we want this type of noise)

tae898: I didn't add this everywhere but I'll keep in mind for sure.

VictorSanh:
another template can be something like `Given this {{dosage}} and {{drug}}, write a prescription.||| {{text}}`

tae898:
> another template can be something like `Given this {{dosage}} and {{drug}}, write a prescription.||| {{text}}`

Thanks for the feedback. I thought about this but then what was difficult was that `text` is not necessarily a prescription. 

They are some short notes definitely written by medical staff but then it's hard to put a label on them, since the scope varies.

PR #125, opened 2021-06-08
Add prompts for `gigaword`

JetRunner:


srush:
Hey, I made this one!

srush:
So while the NLP task is summary, these aren't really documents and summaries. It is roughly the first sentence of a newspaper article and the title of the article.  Can you make the prompts say that explicitly.

JetRunner:
@srush Updated!

PR #127, opened 2021-06-09
wiki_source

arunraja-hub:


VictorSanh:
For now, our focus is on English datasets so let's skip the translation datasets as it has implications (tokenization, pre-trained model we use, continued pre-training, etc.).

I'll leave this PR open for the time being as re-evaluate when we extend the setup of Experiment D to multiple languages.

PR #128, opened 2021-06-09
wiki_bio

arunraja-hub:


srush:
Nice, this is a really good dataset.

srush:
Oh shoot, @arunraja-hub I think I merged this too quickly. Can you make the following changes: 

- Filter out "article title" and any other meta columns as one of the facts that gets printed out input_text["table"]["column_header"] everywhere.

PR #129, opened 2021-06-09
Squad v2

lintangsutawika:
Added 7 prompt templates for squad v2

### DIFF
--- a/templates/squad_v2/templates.yaml
+++ b/templates/squad_v2/templates.yaml
@@ -10,7 +10,6 @@ templates:
 
       ] %}
 
-      {% if answers.text == [] %}
 
       {{ seq | random }}
 
@@ -19,18 +18,12 @@ templates:
       Question: {{question}};
 
       Answer: |||
+
+      {% if answers.text == [] %}
 
       Answer not in context
 
       {% else %}
-
-      {{ seq | random }}
-
-      Context: {{context}};
-
-      Question: {{question}};
-
-      Answer: |||
 
       {{answers.text[0]}}
 
@@ -163,15 +156,13 @@ templates:
 
       ] %}
 
-      {% if answers.text == [] %}
-
       {{context}} {{ seq | random }}{{question}}|||
+
+      {% if answers.text == [] %}
 
       Answer not in context
 
       {% else %}
-
-      {{context}} {{ seq | random }}{{question}}|||
 
       {{answers.text[0]}}
 
--- a/templates/squad_v2/templates.yaml
+++ b/templates/squad_v2/templates.yaml
@@ -1,0 +1,118 @@
+dataset: squad_v2
+templates:
+  17b83a3f-f748-42e6-9cdf-b2951dd8299d: !Template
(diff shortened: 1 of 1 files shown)


VictorSanh (changes requested):
Thank you @lintangsutawika , these look great!
Made a few suggestions.
Could you add a few more templates with more variety in the way you query the model? we posted some recs [here](https://huggingface.slack.com/archives/C023Z96L7B4/p1623244601086900) this morning :)
For instance, this could be something like:
```jinja
My daughter is asking me something about {{title}}: {{question}}.
Here's what I found on the internet: {{context}}.
What is the answer?|||
{{answer.text[0]}}
```

lintangsutawika:
Noted, will add more prompts!

lintangsutawika:
@VictorSanh please check.

templates/squad_v2/templates.yaml
```diff
@@ -0,0 +1,224 @@
...
+
+      Context: {{context}};
+
+      Question: {{question}};
+
+      Answer: |||
+
+      {{answers.text[0]}}
+
+      {% endif %}'
```
VictorSanh, line 37: For easier read, let's factorize this: the prompt is always the same so let's move it outside of the "label" and keep the if condition `{% if answers.text == [] %}` only for the label part -> you won't have two `|||` which you almost never need

templates/squad_v2/templates.yaml
```diff
@@ -0,0 +1,224 @@
...
+
+      Answer not in context
+
+      {% else %}
+
+      {{context}} {{ seq | random }}{{question}}|||
+
+      {{answers.text[0]}}
+
+      {% endif %}'
```
VictorSanh, line 178: same comment as before. factorize the prompt outside of the if condition

templates/squad_v2/templates.yaml
```diff
@@ -0,0 +1,224 @@
...
+    name: Topic Prediction - Question and Answer Pair
+    reference: Given a Question-Answer pair, generate the topic.
+  fdcf132e-6c70-4188-999e-93601ee8e089: !Template
+    id: fdcf132e-6c70-4188-999e-93601ee8e089
+    jinja: 'What is the follow passage about?
+
+      {{context}} |||
+
+      {{title | replace("_", " ")}}'
+    name: Topic Prediction - Context
```
VictorSanh, line 223: this makes me think that we'll need to deduplicate inside a training set

VictorSanh:
> @VictorSanh please check.

thank you, small comment on readability
please ping me when you are done, I'll merge then

lintangsutawika:
Hi @VictorSanh have just factorized the templates with conditionals.

templates/squad_v2/templates.yaml
lintangsutawika: I'd be happy to help write any implementation for the duplication.

VictorSanh (approved):
thank you merging now, please update the spreadhseet!

PR #132, opened 2021-06-09
prompt nlu_evaluation_data

zaidalyafeai:


templates/nlu_evaluation_data/templates.yaml
```diff
@@ -0,0 +1,171 @@
+dataset: nlu_evaluation_data
+templates:
+  352e2261-149c-4363-a1e6-8fa4d5c93481: !Template
+    id: 352e2261-149c-4363-a1e6-8fa4d5c93481
+    jinja: "Given the following text, what is the scenario ? \n{{text}}\n|||\n{{\n\
```
VictorSanh, line 5: Use as little "linguistic terminology" as possible. You should replace the word "scenario" with something more explicit (layman's terms)! `nlu_eval_data_2` is a great example!

templates/nlu_evaluation_data/templates.yaml
```diff
@@ -0,0 +1,171 @@
...
+      ,\n  \"qa_definition\",\n  \"qa_factoid\",\n  \"qa_maths\",\n  \"qa_stock\"\
+      ,\n  \"recommendation_events\",\n  \"recommendation_locations\",\n  \"recommendation_movies\"\
+      ,\n  \"social_post\",\n  \"social_query\",\n  \"takeaway_order\",\n  \"takeaway_query\"\
+      ,\n  \"transport_query\",\n  \"transport_taxi\",\n  \"transport_ticket\",\n\
+      \  \"transport_traffic\",\n  \"weather_query\"\n][label].split('_')[0]\n}}"
+    name: nlu_eval_data_1
+    reference: ''
+  5346ace7-c124-4931-8180-9c4c6f55cec3: !Template
+    id: 5346ace7-c124-4931-8180-9c4c6f55cec3
+    jinja: "This text describes what {{\"scenario\"}} ? \n{{text}}\n|||\n{{\n[\n \
```
VictorSanh, line 30: What {{"scenario"}} does this text describe?

templates/nlu_evaluation_data/templates.yaml
```diff
@@ -0,0 +1,171 @@
...
+    reference: ''
+  60fba53b-c5b0-464f-ae75-ce1aeea72dad: !Template
+    id: 60fba53b-c5b0-464f-ae75-ce1aeea72dad
+    jinja: "If this is the {{\"scenario\"}}, what would be a query for it ? \n{{scenario}}\n\
+      |||\n{{text}}"
+    name: nlu_eval_data_8
+    reference: ''
+  83b688de-bacd-4643-87a5-9a5a64ddbc46: !Template
+    id: 83b688de-bacd-4643-87a5-9a5a64ddbc46
+    jinja: "Which {{\"scenario\"}} from the following: {{ ['music', 'news', 'alarm',\
```
VictorSanh, line 61: from the following choices ( the choices) best describe

templates/nlu_evaluation_data/templates.yaml
```diff
@@ -0,0 +1,171 @@
...
+      ,\n  \"recommendation_events\",\n  \"recommendation_locations\",\n  \"recommendation_movies\"\
+      ,\n  \"social_post\",\n  \"social_query\",\n  \"takeaway_order\",\n  \"takeaway_query\"\
+      ,\n  \"transport_query\",\n  \"transport_taxi\",\n  \"transport_ticket\",\n\
+      \  \"transport_traffic\",\n  \"weather_query\"\n][label].replace('_', '  ')\n\
+      }}"
+    name: nlu_eval_data_2
+    reference: ''
+  ce869e8f-7975-4d5c-bd91-a23624c17285: !Template
+    id: ce869e8f-7975-4d5c-bd91-a23624c17285
+    jinja: "The following piece of text, talks about what {{\"scenario\"}} ? \n{{text}}\n\
```
VictorSanh, line 149: What {{"scenario"}} does the following piece of text talk about?

VictorSanh (changes requested):
Thank you @zaidalyafeai !

Can you re-equilibrate to have more templates where you give the model the classes to choose from?

Another template I just thought about is something like:
```jinja
The virtual assistant received this query: {{text}}.
What skill should it activate among these choices: {{["alarm", ...]}}|||
{{ list[label].split("_")[0] }}
```
Could you add it? Feel free to be creative!

zaidalyafeai:
> Thank you @zaidalyafeai !
> 
> Can you re-equilibrate to have more templates where you give the model the classes to choose from?
> 
> Another template I just thought about is something like:
> 
> ```
> The virtual assistant received this query: {{text}}.
> What skill should it activate among these choices: {{["alarm", ...]}}|||
> {{ list[label].split("_")[0] }}
> ```
> 
> Could you add it? Feel free to be creative!

Thanks @VictorSanh , tried my best to respond to the review.

VictorSanh:
Thank you!

Re on the use of linguistic terms -> can we remove the word "scenario" whenever we don't have the whole list of choices (for instance `nlu_eval_data_4`)? Either we replace "scenario" with a descriptive text or we have the list of choices. My rule of thumb: if I saw this query, would I be able to give a reasonable answer without knowing about this particular dataset?

zaidalyafeai:
Thanks @VictorSanh, that makes sense. Maybe we could replace all occurrences of scenario with service ? An example might be "What service should be started when a user issues the following query ? " Not sure, though.

VictorSanh:
> Thanks @VictorSanh, that makes sense. Maybe we could replace all occurrences of scenario with service ? An example might be "What service should be started when a user issues the following query ? " Not sure, though.

Yes I am more comfortable with this!

zaidalyafeai:
@VictorSanh I rephrased all the occurrences to be more reasonable.

VictorSanh (approved):
thank you, merging now, please update the spreadsheet!

PR #133, opened 2021-06-09
lambada

arunraja-hub:


srush:
I think there is something wrong with the data. This is supposed to be a language modeling dataset on passages. I think it was processed incorrectly when added to Datasets library.

arunraja-hub:
Yes I thought so too. In https://huggingface.co/datasets/lambada, the `text` is described as:

> text: the text (concatenation of context, target sentence and target word). The word to be guessed is the last one.

But the provided dataset does not contain those 3 parts. I assumed the dataset version provided was adapted for a more general-purpose use

srush:
Interesting. Let's skip this for now then. This is one of my favorite datasets, I would like to get it right.

arunraja-hub:
Sgtm!

PR #136, opened 2021-06-09
Add templates for the WiLI-2018 dataset

abheesht17:


stephenbach:
Hi,
Thank you so much for submitting a PR! Unfortunately, this is a non-English dataset that that slipped through the cracks of our filtering. We won't be merging it in right now because we're starting with an English pre-trained model, but you could leave this open and hopefully we'll eventually grow our collection of templates.
Thanks!

abheesht17:
Aah, I see. On to another dataset, then :)

PR #138, opened 2021-06-09
Add Poem Sentiment Dataset

blackrosedragon2:
An overview of this PR:

1. Templates 1-6 are similar to the ones from IMDb's template.
2. Templates 7 and 8 constitute of variation in localization of the label
3. In template 9 I have replaced the label "no_impact" with "neutral". 

I have two questions:
1. Should we replace all "no_impact" labels with "neutral". The original example provided on the streamlit app has the former.
2. Should I add a template that includes negation. For example, after the poem the prompt will be : The sentiment the poem does not express is <any label except for the correct label>. With reference to the "Language Models are Few-Shot Learners" paper, GPT-3 does not perform well on "negation" tasks like these. Does it make sense to include it in the template here for this dataset?

### DIFF
--- a/templates/poem_sentiment/templates.yaml
+++ b/templates/poem_sentiment/templates.yaml
@@ -48,13 +48,13 @@ templates:
     reference: ''
   e53d2fe8-b83b-484e-81f9-efca32bd7012: !Template
     id: e53d2fe8-b83b-484e-81f9-efca32bd7012
-    jinja: '{{verse_text}} The poem expresses {{ ["negative", "positive","no_impact","mixed"]
+    jinja: '{{verse_text}} The poem expresses ||| {{ ["negative", "positive","no_impact","mixed"]
       [label] }} sentiment.'
     name: poem_sentiment_7
     reference: ''
   fdbebd3d-6517-4be1-8771-489e2de658ef: !Template
     id: fdbebd3d-6517-4be1-8771-489e2de658ef
-    jinja: '{{verse_text}} The poet wants the readers to feel {{ ["negative", "positive","no_impact","mixed"]
-      [label] }} sentiment.'
+    jinja: '{{verse_text}} The poet wants the readers to feel ||| {{ ["negative",
+      "positive","no_impact","mixed"] [label] }} sentiment.'
     name: poem_sentiment_8
     reference: ''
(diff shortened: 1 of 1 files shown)


templates/poem_sentiment/templates.yaml
```diff
@@ -0,0 +1,60 @@
...
+    jinja: '{{verse_text}} The poem expresses {{ ["negative", "positive","no_impact","mixed"]
+      [label] }} sentiment.'
+    name: poem_sentiment_7
+    reference: ''
+  fdbebd3d-6517-4be1-8771-489e2de658ef: !Template
+    id: fdbebd3d-6517-4be1-8771-489e2de658ef
+    jinja: '{{verse_text}} The poet wants the readers to feel {{ ["negative", "positive","no_impact","mixed"]
+      [label] }} sentiment.'
+    name: poem_sentiment_8
+    reference: ''
```
VictorSanh, line 60: missing `|||`

templates/poem_sentiment/templates.yaml
```diff
@@ -0,0 +1,60 @@
+dataset: poem_sentiment
+templates:
+  211c0765-1f51-4574-b354-040273ea7c38: !Template
+    id: 211c0765-1f51-4574-b354-040273ea7c38
+    jinja: '{{verse_text}} How does the reader feel about this poem? |||  {{ ["negative",
+      "positive","no_impact","mixed"] [label] }}'
```
VictorSanh, line 6: replace the `no_impact` by `no impact`

VictorSanh: i agree with your comment that "neutral" is a better term

VictorSanh (changes requested):
Can you include the different options for classification for half of the templates?
It makes the task kind of very difficult if you don't know what to expect, even though it's good to have this different levels of difficulty.
It would something like `Does this poem express a {{"negative"}}, {{"positive"}}, {{"neutral"}} or {{"mixed"}} sentiment? {{verse_text}}`

blackrosedragon2:
> Can you include the different options for classification for half of the templates?
> It makes the task kind of very difficult if you don't know what to expect, even though it's good to have this different levels of difficulty.
> It would something like `Does this poem express a {{"negative"}}, {{"positive"}}, {{"neutral"}} or {{"mixed"}} sentiment? {{verse_text}}`

Sure! Will do.

VictorSanh (approved):

VictorSanh:
merging now, thank you @blackrosedragon2

PR #139, opened 2021-06-09
Prompts for cc_news

elsanns:


VictorSanh (changes requested):
Thank you @elsanns!

A few comments:
- no need for extra white spaces between ||| and the generation
- Could you add a few variations of the reverse task:
```jinja2
What could be the content of a news article with the following text and summary?
Title: {{title}}
Summary: {{summary}}
|||
{{text}}
```

VictorSanh:
> I will add examples for the reverse task, I first thought it could cause the model to memorize them (pairs: standard and reverse).

I think we can greatly minimize that effect if for a given dataset and a given example in it, there is only one template applied to it. Even though for some dataset (like reading comprehension), the same x is used multiple times (multiples questions)... we should be able to flag them programmatically later by parsing the jinja

PR #141, opened 2021-06-09
Add a new task template field for fields corresponding to the original task.

srush:
image

There are a bunch of templates that are clever, but do not correspond to the original proposed task. This makes users mark them explicitly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -133,6 +133,7 @@ combined to form different (input, output) pairs i.e. different "tasks". Don't h
 introduce some diversity by prompting a given dataset into multiple tasks and provide some
 description in the "Template Reference" text box. An example is given
 in the already prompted `movie_rationales`.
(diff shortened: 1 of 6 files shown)


stephenbach (approved):
Awesome! All LGTM and runs smoothly.

A few suggestions:
- We probably want to update CONTRIBUTING.md. Should include deciding whether it's a task template as a step in the quick start guide. Should give more explicit guidance about what "task template" means so we have consistent annotations. For example, if the dataset is not associated with an explicit task, is that True or False? What if it's one of multiple possible tasks (e.g., something with coarse and fine-grained labels)? Implicit situations could also be confusing. For example [this one for AG NEWS](https://github.com/bigscience-workshop/promptsource/blob/main/templates/ag_news/templates.yaml#L5). It's an implicit situation, but I think for our purposes it's still the task for which the dataset was designed.
- Update the prompted dataset view to display the field.
- Consider updating the helicopter view to have a column showing how many are task templates or something like that, unless it's too crowded.

LMK if I can help with any of these!

srush:
Okay, think I got most of these.

stephenbach:
Looks great!

PR #142, opened 2021-06-09
Added templates for qasc dataset

rbawden:


stephenbach:
These are awesome! Thanks!

Only tiny thing I noticed is that sometimes the choices have periods on the end (and sometimes whitespace), rendering like

screenshot

I saw that you were trimming periods on other fields. Maybe can do that with choices too?

rbawden:
Good spot - I hadn't seen those. I image it could actually be something quite widespread throughout the datasets? I have added trim and trim('.') to the options.

stephenbach:
Great, thanks! And yeah, I think it could be a common thing. Just tagged you on another issue where I shared the little trick. 🙂

PR #143, opened 2021-06-09
Prompts for cc_news update

elsanns:
Removed white space around ||| and added for examples with text as a target.
I noticed that new lines are rendered differently in one file and across submissions, hope it can be handled


VictorSanh (approved):
Lgtm! Thank you @elsanns 
Merging now, please update the spreadsheet

(btw you could have continue working on the previous PR!)

PR #144, opened 2021-06-09
Prompts for art

elsanns:


stephenbach:
These are great! One neat trick I just saw @rbawden use is to trim periods off of sentences when expressing them as questions by appending `| trim('.')` to Jinja expressions. So for example, `hyp1_1` could be
```
Given that: 

{{  observation_1  }} 

And: 

{{  observation_2 }} 

Is it true that:

{{ hypothesis_1 | trim('.') }}?
||| 
{{ ["Yes", "No"][label-1] }}
```
What do you think? Could make the questions more natural?

elsanns:
Yes, I thought about it. There are other possible endings to replace with a question mark (like: !). Same for capital letters etc., sometimes the punctuation and case will be weird (btw, will it all be lowercased?). 

But the dot is the most common one, I can add that.

stephenbach:
Ok! You can compose these operators so, e.g.
```
{{ hypothesis_1 | trim('.?!') | lower }}
```
would strip `.`, `?`, and `!` off the end and then lower case everything.

elsanns:
Yes it's a good idea. There will be cases like dots in the middle and proper names but it's an improvement in most cases. Adding to my todo list (I guess for tomorrow).

stephenbach:
Thanks for the updates! It looks like this might have got a little mangled in the merging? There are some unresolved conflicts in the templates. Perhaps we can just do art in this PR and then do another PR for improving some of the other datasets?

elsanns:
Yes, I only intended to push changes to the art dataset but got the merge wrong.
I opened a new PR #167 with clean changes to the art prompts, we can just close this one.
Sorry for the confusion!

stephenbach:
Gotcha, no worries!

PR #145, opened 2021-06-09
added templates for stereoset

josrozen:
a few stereoset templates, reusing the "stereotypical" and "unrelated" annotations in a different setting


srush:
Thanks for doing these prompts. I think we are maybe going to skip this dataset. You can mark it as completed in the spreadsheet.

PR #146, opened 2021-06-09
Add MDD templates

gchhablani:
**Notes**:
- `task3_qarecs` has cases where there are 6 utterances and 4 utterances. Some of the templates will lead to empty input and output because of this. Some prompts may also be repeated across templates.
- `task4_reddit` has some inappropriate words and it also contains URLs. I'm not 100% sure if we should add this. @VictorSanh also recommended not to add this subset so I'll probably remove it. It seemed like an interesting task to me, that's why I added it here.
- Distribution of the number of utterances:
```
task1_qa
------------------------------
train
{2: 96185}
validation
{2: 9968}
test
{2: 9952}

task2_recs
------------------------------
train
{2: 1000000}
validation
{2: 10000}
test
{2: 10000}

task3_qarecs
------------------------------
train
{4: 64334, 6: 887791}
validation
{4: 433, 6: 4619}
test
{4: 445, 6: 4470}

task4_reddit
------------------------------
train
{2: 722806, 4: 160226, 6: 42266, 8: 12780, 10: 4294, 12: 1533, 14: 560, 16: 279, 18: 155, 20: 92, 22: 57, 24: 42, 26: 22, 28: 20, 30: 16, 32: 10, 34: 12, 36: 6, 38: 1, 40: 2, 42: 1, 44: 2, 46: 3, 48: 3, 50: 1, 52: 2, 56: 2, 62: 1, 66: 1, 88: 1, 92: 1, 100: 1}
validation
{2: 7576, 4: 1774, 6: 452, 8: 128, 10: 36, 12: 18, 14: 14, 16: 1, 26: 1}
test
{2: 7704, 4: 1635, 6: 448, 8: 129, 10: 55, 12: 15, 14: 7, 16: 1, 18: 2, 20: 2, 24: 1, 30: 1}
```


VictorSanh:
Thank you @gchhablani 
I am putting this PR in stand-by for the moment, waiting for recommendations from the social-impact meta WG on how to proceed with the `task4_reddit` subset and other similar potentially harmful data(sub)sets

gchhablani:
Cool. I think the other three tasks do not have such issues so maybe they can be added?

Wdyt about the issues with `task3_qarecs`?

templates/mdd/task3_qarecs/templates.yaml
```diff
@@ -0,0 +1,84 @@
...
+  92cb9273-b89f-410f-a7f2-db8c84f42862: !Template
+    id: 92cb9273-b89f-410f-a7f2-db8c84f42862
+    jinja:
+      "{% if dialogue_turns.utterance|length==6 %}{{dialogue_turns.utterance[0]}}
+
+
+      {{dialogue_turns.utterance[4]}}|||{{dialogue_turns.utterance[5]}}{% else %}|||{%
+      endif %}"
+    name: recommend_movie_4
+    reference: Given likes and preferences, recommend a movie.
```
VictorSanh, line 56: I find partial history templates kind of weird. Not sure I would be able to answer it correctly myself even if I knew a lot about films (`recommend_movie_3` and `recommend_movie_4`). I think there is a lot of interesting reasons to encourage the model to process long inputs/conditionings. I would remove these (3 and 4)

templates/mdd/task3_qarecs/templates.yaml
```diff
@@ -0,0 +1,84 @@
...
+    jinja:
+      "{% if dialogue_turns.utterance|length==6 %} {{dialogue_turns.utterance[1]}}
+
+
+      {{dialogue_turns.utterance[3]}}
+
+
+      {{dialogue_turns.utterance[4]}}|||{{dialogue_turns.utterance[5]}}{% else %}|||{%
+      endif %}"
+    name: recommend_movie_3
```
VictorSanh, line 43: I don't understand this one:
- 1,3,4,5 ? No 2?
- hard to understand if you don't describe that it is a dialogue

templates/mdd/task3_qarecs/templates.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+      Speaker 1: {{dialogue_turns.utterance[2]}}
+
+
+      Speaker 2:|||{{dialogue_turns.utterance[3]}}"
+    name: next_utterance_4
+    reference: Given the first three dialogues,  generate next utterance.
+  3e5a19e5-aa33-467a-bcd2-f84d99f32759: !Template
+    id: 3e5a19e5-aa33-467a-bcd2-f84d99f32759
+    jinja: "{{dialogue_turns.utterance[0]}}|||{{dialogue_turns.utterance[1]}}"
```
VictorSanh, line 20: Missing a task description right now
Capture d’écran 2021-06-09 à 7 40 47 PM

templates/mdd/task3_qarecs/templates.yaml
```diff
@@ -0,0 +1,84 @@
+dataset: mdd
+subset: task3_qarecs
+templates:
+  1614890b-362c-4ee8-850d-841cf511d169: !Template
+    id: 1614890b-362c-4ee8-850d-841cf511d169
+    jinja: "Speaker 1: {{dialogue_turns.utterance[0]}}
```
VictorSanh, line 6: Can you add some sentences that describe the task? something like `This is a dialogue between Speaker 1 and Speaker 2. Complete Speaker 2's intervention to answer Speaker 1's question`.

Nit: is it always alternating between speaker 1 and speaker 2?

VictorSanh (changes requested):
Thank you @gchhablani , please see my comments

VictorSanh:
> Cool. I think the other three tasks do not have such issues so maybe they can be added?

Yes

> Wdyt about the issues with `task3_qarecs`?

I think the if condition that you implemented are the right way to go. However there might be some overlap between `next_utterance_4` and `next_utterance_4_and_6` the second being more general. So I'd recommend wrapping the whole template in the if condition -> the first one would just be applied to 4 length dialogues and the second one only 6 length dialogues

templates/mdd/task3_qarecs/templates.yaml
gchhablani: The data has this pattern:
- Speaker 1 describes likes, asks for a movie.
- Speaker 2 recommends a movie.
- Speaker 1 asks a question about the movie.
- Speaker 2 answers the question.
[Optionally]
- Speaker 1 tells preferences and asks for another recommendation.
- Speaker 2 recommends another movie.

Why I thought 3 and 4 are interesting:
- In 3, the movie name and its description are provided, then a preference is provided over this recommended movie, and then a movie is recommended.
- In 4, the likes, preferences are provided, and then a movie is to be recommended.

I agree with you on the point of being able to answer it correctly. But, if a model is able to predict the next utterance using the entire history, should it also not be able to do it with these details to a certain extent?

templates/mdd/task3_qarecs/templates.yaml
gchhablani: In this, the task description is being provided in the dialogue itself - "I'm looking for a Jack Nicholon movie.", wdyt?

templates/mdd/task3_qarecs/templates.yaml
gchhablani: The only true dialogue task is `next_utterance`, others are variants which are not exactly dialogue tasks.

templates/mdd/task3_qarecs/templates.yaml
gchhablani: Yes, I'll add this description.

And yes, it is always alternating between speaker 1 and speaker 2 and both get an equal number of utterances.

gchhablani:
@VictorSanh
The `next_utterance_4_and_6` will not have `next_utterance_4` cases for dialogues with length 6 then, which are, in themselves, complete. Maybe I should add a condition in `next_utterance_4` to only generate for length 6 dialogues and generate empty input-output pairs in case of length 4?

VictorSanh:
> Basically:
> `next_utterance_4_and_6` will cover entire dialogues, while `next_utterance_4` will cover first 4 utterances for length 6 dialogues only. Wdyt?

Sounds good! let's do that

gchhablani:
Cool, I'll do that. What about other comments?

templates/mdd/task3_qarecs/templates.yaml
VictorSanh: Right, i think the interesting part of dialogue is generating the next utterance in reaction to the grounding in a coherent and logical way, looking at the examples from `recommend_movie_3` and `recommend_movie_4`, it's hard to find the consistency:

`recommend_movie_3`
Capture d’écran 2021-06-10 à 11 23 48 PM
these look like random concatenation of random entities.

`recommend_movie_4`
Capture d’écran 2021-06-10 à 11 21 16 PM
The history is destroyed and it looks very like noise (why more? why other suggestions?)

VictorSanh: you need to give more details about what these specific utterances expresse and how the next utterance is connected to these (the reasoning part). for instance this could be
`Given this celebrities name {{uterance[0]}}, and the fact that I like more {{utterance[3]}}, provide a recommendation |||`

templates/mdd/task3_qarecs/templates.yaml
VictorSanh: see comment above, if you remove the history (i.e. the logical connections between utterances, you need to reintroduce it somehow otherwise it looks like a concatenation of random entities)

templates/mdd/task3_qarecs/templates.yaml
VictorSanh: 👍gotcha

VictorSanh:
see answers above

templates/mdd/task3_qarecs/templates.yaml
gchhablani: Since the each utterance is complete in itself, can I do something like:
`recommed_movie_3`
```
I am aware that the movie {{utterance[1]}} is about {{utterance[3]}}. However, {{utterance[4]}}
|||
{{utterance[5]}}
```
and 
`recommend_movie_4`
```
{{utterance[0]}} Also, {{utterance[4]}}
|||
{{utterance[5]}}
```

gchhablani: Another way to explicitly state that a movie is to be recommended:
`
recommed_movie_3
`
```
{{ ["My friend is ", "I am", "My colleague is", "A person is "]|random }} aware that the movie {{utterance[1]}} is about {{utterance[3]}}. However, <if "I am" is not selected> they said: <endif>{{utterance[4]}}. Can you recommend a movie for <if "I am" is not selected> them <else> me <endif>?
|||
{{utterance[5]}}
```
and
`recommend_movie_4`

```
{{['','A friend said:', 'A colleague said:'| 'A person said:'] | random}}{{utterance[0]}} Also, <choose appropriate phrase here>: {{utterance[4]}}. Recommend a movie for <if "" is not selected> them <else> me <endif>.
|| 
{{utterance[5]}}
```
First one is empty because it's a first-person statement in `recommend_movie_4`.

gchhablani:
I've updated with the changes discussed. Have tried to make them slightly diverse and add more context.

templates/mdd/task3_qarecs/templates.yaml
VictorSanh: let's have both, i like them both

VictorSanh (approved):
thank you merging now!

PR #147, opened 2021-06-09
Prompts for esnli

elsanns:


templates/esnli/templates.yaml
```diff
@@ -0,0 +1,119 @@
...
+
+      |||
+
+      {{explanation_1}}'
+    name: '  explain_3'
+    reference: ''
+  3d077f02-bba0-4019-bacd-bf3e7119b1f7: !Template
+    id: 3d077f02-bba0-4019-bacd-bf3e7119b1f7
+    jinja: "{% if label == 0 %} \nWhy does the first sentence entail the second?\n\
+      {% elif label == 1 %}\nWhy are the two sentence neutral?\n{% else %}\nWhy do\
```
stephenbach, line 22: It might be more natural to say something like "Why do the two sentences neither entail nor contradict each other?" In isolation, there isn't a lot of context to infer what neutral means here.

stephenbach (commented):
These are so good! I just commented on one line where it might be better to phrase the task with more context. What do you think?

elsanns:
Yes, absolutely, I'm just using target classes everywhere, but 'neutral' is
ambiguous without a contex, I will change it

czw., 10 cze 2021, 17:22 użytkownik Stephen Bach ***@***.***>
napisał:

PR #148, opened 2021-06-09
Add app_reviews templates

gchhablani:
**Notes:**
- I am not sure about `convert_to_star_rating`. Should a model be able to generate emojis? 🤔

**Edit**
This is from `bert-base-uncased`:
```python
>>> bt('[UNK]')
{'input_ids': [101, 100, 102], 'token_type_ids': [0, 0, 0], 'attention_mask': [1, 1, 1]}
>>> bt('🤔')
{'input_ids': [101, 100, 102], 'token_type_ids': [0, 0, 0], 'attention_mask': [1, 1, 1]}
```

### DIFF
--- a/templates/app_reviews/templates.yaml
+++ b/templates/app_reviews/templates.yaml
@@ -2,7 +2,7 @@ dataset: app_reviews
 templates:
   2da8f134-58db-4f9d-b3b0-8c6b50693ab5: !Template
     id: 2da8f134-58db-4f9d-b3b0-8c6b50693ab5
-    jinja: 'Given this review: {{review}}
+    jinja: "Given this review: {{review}}
 
       Would you recommend this app to a friend?
 
@@ -28,21 +28,23 @@ templates:
 
       Not at all
 
-      {% endif %}'
+      {% endif %}"
     name: categorize_rating_using_review
     reference: Given the review, return a categorical answer.
   8086b434-a75e-45a4-87fb-4364601e2e05: !Template
     id: 8086b434-a75e-45a4-87fb-4364601e2e05
-    jinja: 'Generate a {{star}} start review about an app with package {{package_name}}.
+    jinja:
+      "Generate a {{star}} start review about an app with package {{package_name}}.
 
       |||
 
-      {{review}}'
+      {{review}}"
     name: generate_review
     reference: Generate a review from the rating.
   9746ce4b-ac58-4dfb-9783-d77c95cb62cf: !Template
     id: 9746ce4b-ac58-4dfb-9783-d77c95cb62cf
-    jinja: "What would be the star rating of this review?\n{% if star==5 %}\n\u2B50\
+    jinja:
+      "What would be the star rating of this review?\n|||{% if star==5 %}\n\u2B50\
       \u2B50\u2B50\u2B50\u2B50\n{% elif star==4%}\n\u2B50\u2B50\u2B50\u2B50\n{% elif\
       \ star==3%}\n\u2B50\u2B50\u2B50\n{% elif star==2%}\n\u2B50\u2B50\n{% else %}\n\
       \u2B50\n{% endif %}"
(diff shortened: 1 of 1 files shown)


templates/app_reviews/templates.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+      |||
+
+      {{review}}"
+    name: generate_review
+    reference: Generate a review from the rating.
+  9746ce4b-ac58-4dfb-9783-d77c95cb62cf: !Template
+    id: 9746ce4b-ac58-4dfb-9783-d77c95cb62cf
+    jinja:
+      "What would be the star rating of this review?\n|||{% if star==5 %}\n\u2B50\
```
craffel, line 47: This prompt does not have the review in it.

templates/app_reviews/templates.yaml
```diff
@@ -0,0 +1,91 @@
...
+      \u2B50\u2B50\u2B50\u2B50\n{% elif star==4%}\n\u2B50\u2B50\u2B50\u2B50\n{% elif\
+      \ star==3%}\n\u2B50\u2B50\u2B50\n{% elif star==2%}\n\u2B50\u2B50\n{% else %}\n\
+      \u2B50\n{% endif %}"
+    name: convert_to_start_rating
+    reference: Given the review, generate a star rating.
+  d34e1413-2699-4701-baa2-05d931d012ba: !Template
+    id: d34e1413-2699-4701-baa2-05d931d012ba
+    jinja: "What would be the star rating of this review: {{review}}?
```
craffel, line 55: This prompt and the one above are ambiguous - the model won't know whether you mean for it to generate the number of stars, a certain number of emojis, etc. I think you need to give the model more instructions as to which you want.

craffel (changes requested):

templates/app_reviews/templates.yaml
gchhablani: I've updated the rating information for both cases. I've added emojis for the star-case, hope that is fine?

templates/app_reviews/templates.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+      |||
+
+      {{star}}"
+    name: convert_to_rating
+    reference: Convert review to rating
+  f923aeeb-1150-486d-99cd-c4ab93ed5130: !Template
+    id: f923aeeb-1150-486d-99cd-c4ab93ed5130
+    jinja:
+      "Would you recommend an app to a friend which has a {{star}}-star rating on a scale of 1 to 5 (1 being lowest and 5 being highest)?
```
craffel, line 66: This prompt will only result in 5 unique examples - I'm not sure that's enough, I think the model would just memorize the mapping right away.

templates/app_reviews/templates.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+      |||
+
+      {{review}}"
+    name: generate_review
+    reference: Generate a review from the rating.
+  9746ce4b-ac58-4dfb-9783-d77c95cb62cf: !Template
+    id: 9746ce4b-ac58-4dfb-9783-d77c95cb62cf
+    jinja:
+      "What would be the \u2B50-rating of this review (\u2B50 being lowest and \u2B50\u2B50\u2B50\u2B50\u2B50 being highest) : {{review}}?\n|||{% if star==5 %}\n\u2B50\
```
craffel, line 47: Looking in to it, T5's vocab can't tokenize the star emoji:
https://colab.research.google.com/drive/1uXD0f2QdRdk9hMTIW9wF2S_R5tvDperz?usp=sharing
Why not just use * instead?

craffel (changes requested):

templates/app_reviews/templates.yaml
gchhablani: Makes sense. I will remove this template then.

craffel:
Thanks!

PR #149, opened 2021-06-10
Add choice filter

craffel:


craffel:
Had to update test to use env from the templates submodule itself, since otherwise the custom filters were not getting added.

PR #152, opened 2021-06-10
Prompt for Atomic

nihalnayak:
This dataset has conditional questions and answers list. Referred to the original paper for ideas.

### DIFF
--- a/templates/atomic/templates.yaml
+++ b/templates/atomic/templates.yaml
@@ -1,0 +1,483 @@
+dataset: atomic
+templates:
+  23bd81a2-b1f0-42cd-ac51-78624fae94bc: !Template
+    id: 23bd81a2-b1f0-42cd-ac51-78624fae94bc
+    jinja: 'Answer the following questions when {{event}}:
+
+      {% if xIntent|length>0 %}
+
+      What was PersonX''s intention?
+
+      {% endif %}
+
+      {% if xNeed|length>0 %}
+
+      What did PersonX need?
+
+      {% endif %}
+
+      {% if xAttr|length>0 %}
+
+      How is PersonX seen as?
+
+      {% endif %}
+
+      {% if xReact|length>0 %}
+
+      How does PersonX feel?
+
+      {% endif %}
+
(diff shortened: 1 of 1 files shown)


stephenbach:
I'm not sure that I understand these prompts. These seems like plausible things that could be true, not things that can be concluded. Perhaps we can adjust the templates to make that more explicit?

I also don't understand what the blanks are. Like what kind of reasoning is being used to generate a response like this?
```
Prompt + X

What did PersonX need if PersonX calls ___ to the importance?

What does PersonX want if PersonX calls ___ to the importance?

What do others want if PersonX calls ___ to the importance?

Y

PersonX needed to be on stage, to be speaking to audience

PersonX wants to give speech, to sign autographs

Others want none
```

I think it's fine to ask multiple questions in a prompt, but I would maybe skip the ones where the answer is none?

nihalnayak:
Back to this PR. You're right in saying that these events seem plausible. I think the dataset is designed to be plausible rather than definitive. How do you think we should make the templates more explicit? 

The blank is a placeholder for possible words or phrases that could be used with the event. They construct the base events with the placeholders as follows 
```
Following Rashkin et al. (2018), we define events as verb phrases 
with a verb predicate and its arguments (“drinks dark roast in the 
morning”). If a verb and its arguments do not co-occur frequently 
enough,2 we replace the arguments with a blank placeholder 
(“drinks in the morning”). In order to learn more general 
representations of events, we replace tokens referring to people 
with a Person variable (e.g. “PersonX buys PersonY coffee”). In
future work, other types of variables could be added for other 
entity references (e.g. “PersonX moves to CityX”).
```
See `Compiling Base Events` section in https://homes.cs.washington.edu/~msap/atomic/data/sap2019atomic.pdf for more details. 

`none` means the worker explicitly responded with the empty response, whereas [] means the worker did not annotate this dimension (https://huggingface.co/datasets/atomic). Since there is a difference between the two outcomes, do you think we should keep the `none` response?

nihalnayak:
Closing this PR.

PR #153, opened 2021-06-10
health fact template

debajyotidatta:


stephenbach (changes requested):
Thanks! This is an interesting data set. One bug and one question:
- Sometimes the label can be 2, which causes an exception because the label name lists only have 2 elements.
- In the template `main_explanation` the `main` text isn't actually included. Is it meant to be?

PR #156, opened 2021-06-10
templates for google_wellformed_query

josrozen:
a few simple templates asking about the wellformedness of queries


stephenbach:
These are nice, thanks! Please update the spreadsheet to mark it as completed.

PR #157, opened 2021-06-10
Added pubmed_qa prompts

jason-fries:
Adds 6 templates for various question answering and text classification prompt formulations for PubMed abstracts.


stephenbach:
These are lovely! Merging them in now. Do you think that the other subsets could also templates? Or are they the same abstracts? For example, the unlabeled subset still has mesh terms. If you think there's more to do, would you please mark just this subset as done on the spreadsheet? Otherwise can mark it all done. Thanks!

PR #158, opened 2021-06-10
Added templates for selqa dataset

rbawden:


rbawden:
I'm looking into this, but there are no problems with the template showing up in the interface - is the same version of Jinja used here? I'm not quite sure what is going wrong yet

stephenbach:
Awesome! The only thing is that we want to avoid using the random filter so that we can control the way prompts are generated more carefully, for things like retraining, adjusting text size for the model, etc. So we have a PR #149 in progress. When that's merged in, can we switch these over to use that instead? Thanks!

rbawden:
Ok, yes of course! I'll look into using that instead

stephenbach:
Awesome, thank you!

PR #159, opened 2021-06-10
Add neural_code_search prompts

thomasw21:
Small question: Do we want the model to output markdown-like answers?


stephenbach:
Nice, thank you! Please mark this as completed on the spreadsheet.

PR #161, opened 2021-06-10
Add social_i_qa prompts

thomasw21:


stephenbach:
There are great! One thing is that we want to avoid using the random filter so that we can control the way prompts are generated more carefully, for things like retraining, adjusting text size for the model, etc. So we have a PR #149 in progress that will add a similar filter called "choice" that we have more control over. When that's merged in, can we switch these over to use that instead? Thanks!

The other thing is that I'm not so sure about `social_i_qa3`. Would it ever be useful for a model to output answers like this? Perhaps we could replace it with one that gives more of a prompt, like "Help me answer this question about social interactions." or something?

thomasw21:
1) Okay will rebase when I get the chance

2) I'm not really sure what do you mean. Usually instead of regenerating the answer, we could have a much shorter answer like "First/Second/Third answer". 

Moreover the idea is that the model would be "forced" to look at the possible answers for example? My intuition is that if our model is strong enough in the generative sense and solves `social_i_qa2`, it should very easily solve `social_i_qa1` (it just needs to not look at the possible answers). Whereas `social_i_qa3` can only be solved if you look at the possible answers, because they only make sense when you know the possible answers.

stephenbach:
Ah yes, I completely see your point. I was coming at it from the angle that there's nothing to indicate that the speaker wants an ordinal value, but we certainly have multiple choice questions where the answer is specified by index in other datasets. Let's leave it. Thanks!

thomasw21:
I've changed the prompt slightly to indicate that the model needs to general ordinal values.

stephenbach:
Thank you! Sorry, one more small thing: I notice that `social_i_qa2` and `social_i_qa4` don't have whether they're a task template selected. And perhaps some (or all?) of the ones marked false should be true. I think anything that's answering the question in the example would be a task template. Would you mind updating that? Thanks!

stephenbach:
Great, thank you!

PR #162, opened 2021-06-10
Templates for spider

NohTow:
Templates for the spider dataset.
One major issue was the variation in the writing style of question (some are not really "questions"), so I tried to make it sounds like exercise statements, hope it is good enough.

### DIFF
--- a/templates/spider/templates.yaml
+++ b/templates/spider/templates.yaml
@@ -1,5 +1,30 @@
 dataset: spider
 templates:
+  058ae5d8-cd6b-44d6-ab36-bbe462b5fa1a: !Template
+    id: 058ae5d8-cd6b-44d6-ab36-bbe462b5fa1a
+    jinja: '"{{query}}" allows me to lookup ||| {{question}}'
+    name: SQL2text_2
+    reference: ''
+    task_template: true
+  0f475e13-201f-47e6-824d-728c9296e9c5: !Template
+    id: 0f475e13-201f-47e6-824d-728c9296e9c5
+    jinja: How would you question someone to get the information returned by the SQL
+      query  "{{query}}" ? ||| {{question}}
+    name: SQL2text_4
+    reference: ''
+    task_template: true
+  1c7edb93-dd92-4e20-a584-e26f0c1976ab: !Template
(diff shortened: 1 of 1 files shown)


templates/spider/templates.yaml
```diff
@@ -0,0 +1,28 @@
+dataset: spider
+templates:
+  4f63f872-06e2-4503-a440-c1908d5cb642: !Template
+    id: 4f63f872-06e2-4503-a440-c1908d5cb642
+    jinja: Can you give me the SQL request which respond to "{{question}}" ? ||| {{query}}
```
srush, line 5: Respond should be "corresponds"

templates/spider/templates.yaml
```diff
@@ -0,0 +1,28 @@
...
+    name: text2SQL_1
+    reference: ''
+  b4168ce2-4854-4c09-a5aa-08cbcb073ab0: !Template
+    id: b4168ce2-4854-4c09-a5aa-08cbcb073ab0
+    jinja: How would you query a SQL database to answer "{{question}}" ? ||| {{query}}
+    name: text2SQL_4
+    reference: ''
+  ba6415e9-9467-4519-8789-57d65524464e: !Template
+    id: ba6415e9-9467-4519-8789-57d65524464e
+    jinja: The SQL query associated to the question "{{question}}" is ||| {{query}}
```
srush, line 15: "associated with"

templates/spider/templates.yaml
```diff
@@ -0,0 +1,28 @@
...
+    name: text2SQL_3
+    reference: ''
+  e843ba1e-09af-48ba-92ca-1b6032404985: !Template
+    id: e843ba1e-09af-48ba-92ca-1b6032404985
+    jinja: 'To respond to "{{question}}", I can use the SQL query: ||| {{query}}'
```
srush, line 20: Instead of "respond to" , I would say "lookup "

srush (changes requested):
These are great. But can you also include the reverse entries. "I would describe the SQL statement ... to a new learner as ||| ..."

NohTow:
I added your corrections/recommendations and checked the main task box.
I also added the reverse entries, feel free to tell me if something is odd.

PR #163, opened 2021-06-10
Add prompts of cosmos_qa dataset

HanNight:


craffel:
These look good, can you make one additional prompt that is just context, question, choices, and answer with no prompting? Also can you create a variant that does not include the choices - just free-form answer?

HanNight:
Thanks for your suggestion, I have added the corresponding prompts.

templates/cosmos_qa/templates.yaml
```diff
@@ -0,0 +1,352 @@
...
+
+      ===
+
+      The best answer is
+
+      |||
+
+      {% if label == 0 %}
+
+      {{ answer0 }}
```
craffel, line 27: For prompts where you label the answers (A. B. etc.) can you always have the model predict the label (A) instead of the answer? If we sometimes have it predict the label and sometimes the answer, it will get ambiguous training.

craffel (changes requested):

templates/cosmos_qa/templates.yaml
HanNight: Apologies for my carelessness. It has been fixed.

craffel:
Thanks!

PR #165, opened 2021-06-10
Common gen

nihalnayak:
Contains template for common_gen


stephenbach:
Would you please make this PR just for common gen?

nihalnayak:
@stephenbach made the requested changes.

stephenbach:
Nice! I wonder if we can increase the diversity a bit by replacing some of the similar ones with a template or two that goes the other way: generate concepts from the sentence? Could also use the new choice filter to ask for a single concept.

nihalnayak:
Replaced two previous templates with (1) sentence to concept list (2) choice prompts.

stephenbach:
Awesome, thanks! Can you just mark them all as task template or not? I think the ones where you generate the sentence given all the concepts would be the task templates?

nihalnayak:
Marked all as task templates or not.

stephenbach:
Great, thank you! Please mark the dataset as completed on the spreadsheet.

PR #166, opened 2021-06-10
Templates for quarel

NohTow:
I wrote prompts for two distincts tasks :
    
- Multiple Choice Answering that output (A) or (B)
- Machine translation that write the logical form describing the question

### DIFF
--- a/templates/quarel/templates.yaml
+++ b/templates/quarel/templates.yaml
@@ -1,0 +1,68 @@
+dataset: quarel
+templates:
+  1068da45-7f91-42df-b44f-55ddeccb5b97: !Template
+    id: 1068da45-7f91-42df-b44f-55ddeccb5b97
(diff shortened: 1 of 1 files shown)


srush:
Hmm I had trouble understanding these. What does "notations" mean in this context? I feel like there needs to be a bit more explanation to imagine a system could get these right.

NohTow:
> Hmm I had trouble understanding these. What does "notations" mean in this context? I feel like there needs to be a bit more explanation to imagine a system could get these right.

I meant "variable names" but was unsure about the terminology of this field, thus called it "notations" but it is indeed a bit weird.
Do you have any suggestions on a better way to give this information to the model ?

I can also only keep templates related to multiple choice answering, but I think it would be nice to have both tasks.

srush:
We are just going to leave this one out. Too weird.

PR #168, opened 2021-06-10
Add ambig_qa prompt for light subset

thomasw21:


thomasw21:
Turns out assignement in if clause breaks test (despite working perfectly fine). See https://github.com/pallets/jinja/issues/1314

thomasw21:
@srush Pinging you in case you might have missed the review notification on this PR.

srush:
These prompts are much more complex than almost any other dataset. Can you explain why you need comments, variables, and randomness? We should try to avoid these all if possible.

thomasw21:
- Comments are linked to an issue about the way we setup some tests concerning template validity. Issue was tracked here: https://github.com/bigscience-workshop/promptsource/issues/216 . I don't mind removing them, if you feel they are unecessary.
 - Variables are used because behaviours are often conditional on the type of  annotations we have `singleQA` vs `multipleQAs`
 - Randomness is used, since current framework doesn't allow to map 1-to-many templates AFAIK, so instead of taking only the first element of a fixed indexed element, we use the newly introduced `choice` operator https://github.com/bigscience-workshop/promptsource/pull/149 .

PR #170, opened 2021-06-10
[WIP] CoNLL-2003 dataset -Ner, Chunk and POS tagger

sbmaruf:
As the dataset is smaller, I only add two prompts for each of the tasks.

Two major issue to follow,

[ISSUE 1]:
After writing the template, on the right side, there is a sample shown in the app.
I see `<p>` tags inside the samples (from the `inspect` of the browser). I am not sure if this is because of HTML or not.

[ISSUE 2]:
After writing the template I don't see any error generated. But when I go the `Prompted dataset viewer` I get the following error, 

```
TemplateSyntaxError: unexpected '|'
Traceback:
File "/usr/local/lib/python3.7/site-packages/streamlit/script_runner.py", line 338, in _run_script
    exec(code, module.__dict__)
File "/Users/sbmaruf/Documents/big-science/promptsource/promptsource/promptsource.py", line 294, in <module>
    prompt = template.apply(example, highlight_variables=True)
File "promptsource/templates.py", line 302, in apply
    rtemplate = env.from_string(jinja)
File "/usr/local/lib/python3.7/site-packages/jinja2/environment.py", line 941, in from_string
    return cls.from_code(self, self.compile(source), globals, None)
File "/usr/local/lib/python3.7/site-packages/jinja2/environment.py", line 638, in compile
    self.handle_exception(source=source_hint)
File "/usr/local/lib/python3.7/site-packages/jinja2/environment.py", line 832, in handle_exception
    reraise(*rewrite_traceback_stack(source=source))
File "/usr/local/lib/python3.7/site-packages/jinja2/_compat.py", line 28, in reraise
    raise value.with_traceback(tb)
File "<unknown>", line 15, in template
```

### DIFF
--- a/templates/conll2003/templates.yaml
+++ b/templates/conll2003/templates.yaml
@@ -1,0 +1,38 @@
+dataset: conll2003
+templates:
+  3b777e30-bdd1-4c4a-b5d9-3372c7a756f4: !Template
+    id: 3b777e30-bdd1-4c4a-b5d9-3372c7a756f4
+    jinja: "Generate named entities from the following sentence.\n{% set flag = 0\
+      \ %}\n{% for i in tokens -%}\n    {% if flag != 0 -%}\n        &nbsp;\n    {%\
+      \ endif -%}\n    {% set flag = 1 -%}\n    {{ i }}\n{% endfor %} \n|||\n{% set\
+      \ flag = 0 %}\n{% set _ner_lable_dict = ({\n0:\"O\",\n1:\"B-PER\",\n2:\"I-PER\"\
+      ,\n3:\"B-ORG\",\n4:\"I-ORG\",\n5:\"B-LOC\",\n6:\"I-LOC\",\n7:\"B-MISC\",\n8:\"\
+      I-MISC\"\n}) \n%}\n{% set outer_cnt = namespace(value=0) -%}\n{% for tok in\
+      \ tokens -%}\n    {% set inner_cnt = namespace(value=0) -%}\n    {% for lable\
+      \ in ner_tags -%}\n         {% if outer_cnt.value == inner_cnt.value -%}\n \
+      \           {% if flag != 0 -%}\n                &nbsp;\n            {% endif\
+      \ -%}\n            {% set flag = 1 -%}\n            {{tok}}:{{ _ner_lable_dict[lable]\
+      \ }}\n        {% endif -%}\n        {% set inner_cnt.value = inner_cnt.value\
+      \ + 1 -%}\n    {% endfor -%} \n    {% set outer_cnt.value = outer_cnt.value\
+      \ + 1 -%}\n{% endfor -%} \n\n"
+    name: flat_question_without_lable
+    reference: Natural question
+  8613a9bb-94c7-4b91-998c-3c298087d719: !Template
+    id: 8613a9bb-94c7-4b91-998c-3c298087d719
+    jinja: "{% set _ner_lable_dict = ({\n0:\"O\",\n1:\"B-PER\",\n2:\"I-PER\",\n3:\"\
+      B-ORG\",\n4:\"I-ORG\",\n5:\"B-LOC\",\n6:\"I-LOC\",\n7:\"B-MISC\",\n8:\"I-MISC\"\
+      \n}) %}\nGenerate named entities from the following sentence. The named entities\
+      \ are {% set flag = 0 %}\n{% for i in ner_tags -%}\n    {% if flag != 0 -%}\n\
+      \        ,&nbsp;\n    {% endif -%}\n    {% set flag = 1 -%}\n    {{ _ner_lable_dict[i]\
+      \ }}\n{% endfor %} \n{% set flag = 0 %}\n{% for i in tokens -%}\n    {% if flag\
+      \ != 0 -%}\n        &nbsp;\n    {% endif -%}\n    {% set flag = 1 -%}\n    {{\
+      \ i }}\n{% endfor %} \n|||\n{% set flag = 0 %}\n{% set outer_cnt = namespace(value=0)\
+      \ -%}\n{% for tok in tokens -%}\n    {% set inner_cnt = namespace(value=0) -%}\n\
+      \    {% for lable in ner_tags -%}\n         {% if outer_cnt.value == inner_cnt.value\
+      \ -%}\n            {% if flag != 0 -%}\n                &nbsp;\n           \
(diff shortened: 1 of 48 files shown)


templates/conll2003/templates.yaml
```diff
@@ -0,0 +1,124 @@
+dataset: conll2003
+templates:
+  3b777e30-bdd1-4c4a-b5d9-3372c7a756f4: !Template
+    id: 3b777e30-bdd1-4c4a-b5d9-3372c7a756f4
+    jinja: "{% set _ner_lable_dict = ({\n0:\"O\",\n1:\"B-PER\",\n2:\"I-PER\",\n3:\"\
+      B-ORG\",\n4:\"I-ORG\",\n5:\"B-LOC\",\n6:\"I-LOC\",\n7:\"B-MISC\",\n8:\"I-MISC\"\
+      \n}) %}\nGenerate named entities from the following sentence. \n{{\"\"}}\n{%\
+      \ for i in tokens -%}\n    {{- \" \" if not loop.last else \"\" -}}\n    {{\
+      \ i }}\n{% endfor %} \n|||\n{% set flag = 0 %}\n{% set outer_cnt = namespace(value=0)\
+      \ -%}\n{% for tok in tokens -%}\n    {% set inner_cnt = namespace(value=0) -%}\n\
```
srush, line 10: This looks very complex. If there a reason you can't use loop index? 

https://stackoverflow.com/questions/1567291/get-loop-index-of-outer-loop

srush (changes requested):

templates/conll2003/templates.yaml
```diff
@@ -0,0 +1,124 @@
...
+      \ -%}\n            {% if flag != 0 -%}\n                &nbsp;\n           \
+      \ {% endif -%}\n            {% set flag = 1 -%}\n            {{tok}}:{{ _ner_lable_dict[lable]\
+      \ }}\n        {% endif -%}\n        {% set inner_cnt.value = inner_cnt.value\
+      \ + 1 -%}\n    {% endfor -%} \n    {% set outer_cnt.value = outer_cnt.value\
+      \ + 1 -%}\n{% endfor -%} \n\n"
+    name: ner_flat_question_with_lable
+    reference: Natural question
+  87bb05ff-f6bf-4c0d-bbcc-56f75095f4a1: !Template
+    id: 87bb05ff-f6bf-4c0d-bbcc-56f75095f4a1
+    jinja: "{% set _chunk_lable_dict = ({\n0:\"O\",\n1:\"B-ADJP\",\n2:\"I-ADJP\",\n\
```
srush, line 83: lable - > label

sbmaruf:
@srush

As suggested, changes are in this commit, https://github.com/bigscience-workshop/promptsource/pull/170/commits/6539b8310db0a4d37cf4a1e769e993e053838c32

**New Prompt**: 3 additional prompt added. The prompts intends to solve the following issue,

-  It is not expected that user will always provide labels strictly from the dataset.
- When user provided a subsample of labels of ner, pos or chunk tags, the model should understand that the given lables should be used for classification instead of original labels in the dataset. In this way, inference will be more natural.


**TODO**: 
- Randomly shuffle the tags. Now the labels are always in the same order. We should shuffle the lables. But I cannot find `shuffle` filter in the app.

**Issue**:
- Test failing because of unused variable. I cannot update a dictionary in jinja without `set` method. There is a workaround using `do` extension. But I cannot find fo extension in the app.

For convenience, A random template looks like below. I think the test is not passing because of `_dummy` variable.
```
{% set _ner_label_dict = ({
0:"O",
1:"B-PER",
2:"I-PER",
3:"B-ORG",
4:"I-ORG",
5:"B-LOC",
6:"I-LOC",
7:"B-MISC",
8:"I-MISC"
}) %}
{% set _random_ner_label_dict = ({}) %}
{% for k,v in _ner_label_dict.items() -%}
    {% set _rand_num=range(0, 100) | random %}
    {% if _rand_num > 50 %}
        {% set _dummy = _random_ner_label_dict.update({k:v}) %}
    {% endif %}
{% endfor %}
Generate named entities from the following sentence. The named entities are
{% for k,v in _random_ner_label_dict.items() -%}
    {{ v }}
    {{- ", " if not loop.last else "" -}}
{% endfor %}
{{""}}
{% for i in tokens -%}
    {{- " " if not loop.last else "" -}}
    {{ i }}
{% endfor %}
|||
{%- for tok in tokens -%}
    {%- set outer_loop = loop -%}
    {%- for label in ner_tags -%}
         {%- if outer_loop.index == loop.index -%}
            {{- "" if outer_loop.first else " " -}}
            {%- if label in _random_ner_label_dict -%}
                {{tok}}:{{ _random_ner_label_dict[label] }}
            {%- else -%}
                {{tok}}:O
            {% endif %}
        {%- endif -%}
    {%- endfor -%}
{%- endfor -%}
```

Please also look into this issue before merge, https://github.com/bigscience-workshop/promptsource/issues/174

sbmaruf:
****More details on random label selection by prompt ** 
We always include "O" tag for ner, pos and chunk labels. "O" lable indicated it is "other" (uncategorized) label.

In case of parts-of-speech tags, there are few labels that are weird, 

```
Generate parts of speech from the following sentence. The parts of speech tags are ", '', #, $, (, ), ,, ., :, ``, CC, CD, DT, EX, FW, IN, JJ, JJR, JJS, LS, MD, NN, NNP, NNPS, NNS, NN|SYM, PDT, POS, PRP, PRP$, RB, RBR, RBS, RP, SYM, TO, UH, VB, VBD, VBG, VBN, VBP, VBZ, WDT, WP, WP$, WRB
```
In my opinion, here first 9 labels should be normalized. However in the regular prompt I didn't change them. But in the random label selection prompt named as `pos_flat_question_with_label` I did the following normalization, 

```
{% if k>=0 and k<= 9 %}
    {% set _dummy = _random_pos_label_dict.update({0:"O"}) %}
{% else %}
    {% set _dummy = _random_pos_label_dict.update({k:v}) %}
{% endif %}
```
This will convert the first 9 weird label as "O" tag. Note that, there is not "O" in dataset for pos labels.

templates/conll2003/templates.yaml
sbmaruf: Done.

templates/conll2003/templates.yaml
sbmaruf: Done

sbmaruf:
Deleting pull request. Creating a new one with updated `zip` function.

PR #176, opened 2021-06-10
Templates for `ncbi_disease`

drugilsberg:
This one compared to the others has been a real pain, but I think the templates are quite interesting.
Not sure the Jinja style is top notch, but it seems functional on the example sets I checked.

### DIFF
--- a/.github/workflows/check_templates.yml
+++ b/.github/workflows/check_templates.yml
@@ -18,8 +18,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
+          pip install .
           pip install -r requirements.txt
       - name: Check templates
         run: |
-          export PYTHONPATH=promptsource:$PYTHONPATH
           pytest test/test_templates.py
--- a/.github/workflows/show_new_templates.yml
+++ b/.github/workflows/show_new_templates.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
     branches: [ main ]
   workflow_dispatch:
-    
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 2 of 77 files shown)


drugilsberg:
Not sure what the check is detecting here: https://github.com/bigscience-workshop/promptsource/runs/2795972372?check_suite_focus=true.
Anyone who can help?

drugilsberg:
@stephenbach sorry to tag you, but I noticed you are one of the gurus on slack ;)

drugilsberg:
> @stephenbach sorry to tag you, but I noticed you are one of the gurus on slack ;)

Not sure what is going on, I followed the usual procedure (did the same for the other two datasets I worked on) :(

stephenbach:
No worries. 😃 The error message is
```
ValueError: Template for dataset ncbi_disease/None with uuid 04458e59-37f1-48dc-bb20-823e836a8c44 has unrecognized variable disease_token.
```
But not sure why it's unrecognized since you're setting it. Does it render ok in the app?

drugilsberg:
See here for example, the variable is set and used, and rendered properly. Same applies when I inspect the examples outside of the `Sourcing` view. Could it be that test is overlooking the `set` directive?

image

stephenbach:
Hmm, not sure. We use Jinja itself to decide which variables are unset. Will have to look more into this and get back to you.

stephenbach:
Per slack discussion, seems the issue is that the variable `disease_token` is only set inside an if statement. If you also set a default value for the if statement, it should pass the test. A little hacky, but easier than messing around with Jinja internals.

drugilsberg:
@stephenbach ready for review, I also made sure that the task template flag is present in all the templates.

stephenbach:
Great, thank you!

PR #178, opened 2021-06-10
commonsense_qa

arunraja-hub:


arunraja-hub:
Thank you. I have added more prompts. One of them is a variant where all the false answers have to be chosen. I feel that having yes/no will make all the answers as yes which could lead to severe imbalance and the options are also not polar opposites or binary in nature like 'cup' and 'bowl' which might not make sense for the model

templates/commonsense_qa/templates.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+      {{question}}
+
+
+      Options:
+
+
+      {% for i in range(choices[''text'']|length) %}
+
+      <b>{{choices[''label''][i]}}</b>: {{choices[''text''][i]}}
```
craffel, line 41: Can you remove this HTML markup? The model will only be able to process plain text.

craffel (changes requested):

craffel:
Thanks!

PR #179, opened 2021-06-10
Add scientific_papers templates

gchhablani:
**Notes**:
- I have only tested for `pubmed`. For `arxiv`, based on the dataset card, I believe it will be similar.
- The contexts might be too long for this to work nicely.


craffel:
I think the abstract->section names task is somewhat ok... you also could do "Predict the first one hundred characters from the abstract" kind of task, or "predict the first sentence"?

gchhablani:
I went for paragraphs and 100 words. The introduction is an expansion of the first few lines of the abstract, right? Is that doable?

gchhablani:
@craffel
Can you please explain the test output?

craffel:
Thanks!

gchhablani:
@craffel
Thanks for reviewing and merging :)

PR #180, opened 2021-06-10
Templates for hotpot_qa fullwiki

yongzx:


templates/hotpot_qa/fullwiki/templates.yaml
```diff
@@ -0,0 +1,66 @@
...
+      information that explain the answer '
+  e1aadf60-85b4-4c1b-9803-c5231e71e74d: !Template
+    id: e1aadf60-85b4-4c1b-9803-c5231e71e74d
+    jinja: "Information:\n{% for sents in context.sentences %}\n  - {{sents | join(\"\
+      \")}}\n{% endfor %}\nQuestion: {{question}}\n||| \nAnswer: {{answer}}"
+    name: Generate Answer
+    reference: Given information and question, generate answer.
+  e20171e3-8965-4878-9014-0b72c84e9fec: !Template
+    id: e20171e3-8965-4878-9014-0b72c84e9fec
+    jinja: "Generate titles (separated by semi-colons) for each of the paragraph below:\n\
```
stephenbach, line 61: For each of the "paragraphs" below

templates/hotpot_qa/fullwiki/templates.yaml
```diff
@@ -0,0 +1,66 @@
...
+      \ context.title%}\n{% if title==paragraph %}\n- {{ context.sentences[loop.index0][supporting_facts.sent_id[outer_loop.index0]]\
+      \ }}\n{% endif %}\n{% endfor %}\n{% endfor %}"
+    name: Generate Answer and Explanations
+    reference: Given information and question, answer it and list the sentences from
+      information that explain the answer.
+    task_template: true
+  3c9260ef-ca10-40fc-b775-39bae9d28ae5: !Template
+    id: 3c9260ef-ca10-40fc-b775-39bae9d28ae5
+    jinja: "Generate titles (in the format of \"paragraph : title\") for each of the\
+      \ paragraph below:\n{% for sents in context.sentences %}\n  - {{sents | join(\"\
```
stephenbach, line 19: For each of the "paragraphs" below

stephenbach (commented):
These are great! Should we do the distractor subset too? Even if it's just one or two, are there other forms of tasks that subset lets us do?

templates/hotpot_qa/fullwiki/templates.yaml
yongzx: Resolved!

templates/hotpot_qa/fullwiki/templates.yaml
yongzx: Resolved!

yongzx:
> Should we do the distractor subset too? Even if it's just one or two, are there other forms of tasks that subset lets us do?
I have duplicated the templates for the distractor subset, if that's fine. My initial plan was to make sure I got the templates right for the fullwiki subset first before I do it for everything.

stephenbach:
Ok, great! Let's take `question_type` out of distractors since there's no difference between the subsets for that one?

yongzx:
Resolved!

stephenbach (approved):
LGTM, thanks!

PR #181, opened 2021-06-10
Templates for wiki_hop original

yongzx:


stephenbach:
These are great! Couple of questions:
- For the relation names, does it make sense to replace the underscores with spaces so that they are more like natural language?
- What do you think about creating some for the masked subset, at least just one for the intended task? The idea of masking is to force the model to read the passage to get the right answer (as opposed to relying on background knowledge or how often the same text occurs in the passage).

yongzx:
> For the relation names, does it make sense to replace the underscores with spaces so that they are more like natural language?

You are absolutely right. I thought about that before this commit, but I reasoned that we need to follow the Wikidata format. Now that I have come across [Resonator](https://reasonator.toolforge.org/), I think I can make such an adjustment.

yongzx:
> What do you think about creating some for the masked subset, at least just one for the intended task? The idea of masking is to force the model to read the passage to get the right answer (as opposed to relying on background knowledge or how often the same text occurs in the passage).

I think the templates for the original subject can be directly applicable to the masked subset. I have reflected the changes in my code.

stephenbach:
Awesome, thanks!

PR #184, opened 2021-06-10
Add go emotions templates

gchhablani:
**Notes**:
- The `raw` subset is relatively complex and filled with lots of `neutral` data (based on a glance at the examples). The `simplified` version is relatively better but has fewer examples.
- The dataset is collected from Reddit, but since it is sentiment classification it could be allowed? From the paper:```We identify harmful comments using pre-defined
lists containing offensive/adult, vulgar (mildly offensive profanity), identity, and religion terms (included as supplementary material). These are used
for data filtering and masking, as described below.
Lists were internally compiled and we believe they
are comprehensive and widely useful for dataset
curation, however, they may not be complete```

Need help with fixing the failing test. `namescope` is not working on the app.

### DIFF
--- a/templates/go_emotions/raw/templates.yaml
+++ b/templates/go_emotions/raw/templates.yaml
@@ -26,6 +26,7 @@ templates:
       {% endif %}"
     name: get_all_sentiments_from_text
     reference: Given the text, provide all the sentiments.
+    task_template: false
   c0a0af62-9282-4bb4-9aed-883d2a0e9a8d: !Template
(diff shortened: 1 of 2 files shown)


gchhablani:
A template that I was thinking of 
```
Choose the sentiments from the following: [...] which best describe this text: {{text}}
|||
...
```
I'm not sure it will be too many labels? Because it will have 28 different sentiments.

VictorSanh:
Built from Reddit, let's skip it for now
putting pr in standby

PR #185, opened 2021-06-10
Prompts for quartz

elsanns:
There are two examples in the test set with repeated blanks, other than that there are always 2 options and one blank (or none: blank at the end) - seems ok.

`dataset['test'].select((261, 704))['question']`
`['Keely had just run a mile and was now running faster to _____ _____.',
 'Keely had just run a mile and was now running slower to _____ _____.']`

### DIFF
--- a/templates/quartz/templates.yaml
+++ b/templates/quartz/templates.yaml
@@ -11,6 +11,7 @@ templates:
       text\"][i]}}\n\n{%- endfor %}"
     name: question-para-2
     reference: ''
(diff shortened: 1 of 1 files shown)


stephenbach:
These are great! Just wondering, should any of them (maybe all of them?) be marked as task templates?

elsanns:
That's right, they do correspond to the original task, marked them all as task templates

stephenbach:
Perfect, thank you!

PR #186, opened 2021-06-10
newspop template

debajyotidatta:


debajyotidatta:
@stephenbach: This PR is just for newspop.

stephenbach (changes requested):
Thank you!

Just a couple of suggestions:
- For evaluation purposes and consistency with other datasets, can you take out the prefixes in the output like "Topic:". You could put them in the input, particularly in cases like `template_6` where there's nothing in the input telling the model what to do.
- Can you capitalize proper nouns like Microsoft, Obama, and Palestine?

stephenbach (approved):
Looks good, thanks!

PR #189, opened 2021-06-11
health fact template

debajyotidatta:
@stephenbach : Addressed the labels issue from https://github.com/bigscience-workshop/promptsource/pull/153

Addressed the new task template checkbox

Also named the template better to avoid confusion

### DIFF
--- a/templates/health_fact/templates.yaml
+++ b/templates/health_fact/templates.yaml
@@ -11,9 +11,9 @@ templates:
     task_template: false
   23a2ef7f-2032-4771-bf39-94b840aee763: !Template
     id: 23a2ef7f-2032-4771-bf39-94b840aee763
-    jinja: "After reading:\n\n {{main_text }}\n\nI believe :\n\n{{claim}}\n\n|||\n\
-      \nis {{[\"definitely false\", \"likely false\", \"definitely true\", \"not proven\"\
-      ][label]}} ."
+    jinja: "After reading:\n\n {{main_text }}\n\nI believe :\n\n{{claim}}\n\nis \n\
+      |||\n\n{{[\"definitely false\", \"likely false\", \"definitely true\", \"not\
+      \ proven\"][label]}} ."
     name: main_explanation_claim5
     reference: ''
     task_template: true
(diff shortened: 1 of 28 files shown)


templates/health_fact/templates.yaml
```diff
@@ -0,0 +1,67 @@
...
+    id: 6ca1299e-216d-4111-bfb8-a4de2c2bee36
+    jinja: "Given the:\n\n{{main_text}}\n\nand the explanation: \n\n{{explanation}}\n\
+      \n|||\n\nyou  {{[\"cannot\", \"cannot\", \"can\",  \"cannot\"][label]}}  claim:\
+      \ \n\n{{claim}}"
+    name: main_explanation_claim4
+    reference: it is only true if the entire claim is true
+    task_template: true
+  a7a041a1-07cf-4787-b19a-ae04e270dd25: !Template
+    id: a7a041a1-07cf-4787-b19a-ae04e270dd25
+    jinja: "After reading:\n\n {{main_text }}\n\n|||\n\nI believe :\n\n{{claim}}\n\
```
craffel, line 62: I don't think the model would be able to extract the claim correctly from that big chunk of text - I would stick to keeping the main text and claim before the |||.

templates/health_fact/templates.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+
+      '
+    name: main_explanation_claim3
+    reference: ''
+    task_template: false
+  6ca1299e-216d-4111-bfb8-a4de2c2bee36: !Template
+    id: 6ca1299e-216d-4111-bfb8-a4de2c2bee36
+    jinja: "Given the:\n\n{{main_text}}\n\nand the explanation: \n\n{{explanation}}\n\
+      \n|||\n\nyou  {{[\"cannot\", \"cannot\", \"can\",  \"cannot\"][label]}}  claim:\
```
craffel, line 55: Same comment about claim being before |||. Also best to avoid putting prompt words ("you") before the |||.

templates/health_fact/templates.yaml
```diff
@@ -0,0 +1,67 @@
...
+      \ think it is completely true that\", \"believe\", \"do not think it has been\
+      \ proven yet that\"][label]}}  :\n\n{{claim}}\n\nis true because of: |||\n\n\
+      {{explanation}}"
+    name: main_explanation_claim2
+    reference: ''
+    task_template: false
+  23a2ef7f-2032-4771-bf39-94b840aee763: !Template
+    id: 23a2ef7f-2032-4771-bf39-94b840aee763
+    jinja: "After reading:\n\n {{main_text }}\n\nI believe :\n\n{{claim}}\n\n|||\n\
+      \nis {{[\"definitely false\", \"likely false\", \"definitely true\", \"not proven\"\
```
craffel, line 15: Please put "is" before |||.

craffel (changes requested):

templates/health_fact/templates.yaml
debajyotidatta: Thanks @craffel : Since the main text and explanation are long and the claim was generally a single succinct sentence, I was under the impression that this would work. From here, https://arxiv.org/abs/2102.07350, I thought just the task specification needs to be as clear as possible. SInce the claims generally follow a somewhat consistent pattern I thought it was okay. 

Thanks for the feedback. I will change them. I am fairly new to this prompt engineering thing.

debajyotidatta:
@craffel : I made the changes you requested.

craffel:
Thanks!

PR #190, opened 2021-06-11
Prompt bing_coronavirus_query_set

zaidalyafeai:


templates/bing_coronavirus_query_set/templates.yaml
```diff
@@ -0,0 +1,53 @@
+dataset: bing_coronavirus_query_set
+templates:
+  27e325b3-68d2-4cb4-9418-a683a8870239: !Template
+    id: 27e325b3-68d2-4cb4-9418-a683a8870239
+    jinja: "This query does not talk about the coronavirus explicitly ? \n{{Query}}\n\
+      |||\n{% if IsImplicitIntent %}\nYes\n{% else %}\nNo\n{% endif %}"
```
craffel, line 6: `{% if IsImplicitIntent %}` will always evaluate to True because it's actually a string! You need to do `{% if IsImplicitIntent == "True" %}` I believe.

templates/bing_coronavirus_query_set/templates.yaml
```diff
@@ -0,0 +1,53 @@
+dataset: bing_coronavirus_query_set
+templates:
+  27e325b3-68d2-4cb4-9418-a683a8870239: !Template
+    id: 27e325b3-68d2-4cb4-9418-a683a8870239
+    jinja: "This query does not talk about the coronavirus explicitly ? \n{{Query}}\n\
```
craffel, line 5: Does this query talk about the coronavirus explicitly?

templates/bing_coronavirus_query_set/templates.yaml
```diff
@@ -0,0 +1,53 @@
...
+    id: 992d541f-9e0c-466d-b4c4-92e9e236f863
+    jinja: "This query about coronavirus was issued in {{Country}} on {{Date}}. Is\
+      \ the intent implicit or explicit ? \n{{Query}}\n|||\n{% if IsImplicitIntent\
+      \ %}\nimplicit\n{% else %}\nexplicit \n{% endif %}"
+    name: 'is_explicit_country_date '
+    reference: ''
+    task_template: true
+  d4a251d7-0e23-4feb-8bf2-18e32c553199: !Template
+    id: d4a251d7-0e23-4feb-8bf2-18e32c553199
+    jinja: "In what date was this query issued, during the Covid-19 pandemic ? \n\
```
craffel, line 35: On what date

templates/bing_coronavirus_query_set/templates.yaml
```diff
@@ -0,0 +1,53 @@
...
+      {{Query}}\n|||\n{{Date}}"
+    name: 'what_date '
+    reference: ''
+    task_template: false
+  df53652c-36dc-45fe-a015-d0781e32cd33: !Template
+    id: df53652c-36dc-45fe-a015-d0781e32cd33
+    jinja: "This query has an indirect relation to Covid-19 ? \n{{Query}}\n|||\n{%\
```
craffel, line 42: Does this query have an indirect...

templates/bing_coronavirus_query_set/templates.yaml
```diff
@@ -0,0 +1,53 @@
...
+      \ if IsImplicitIntent %}\nYes\n{% else %}\nNo\n{% endif %}"
+    name: is_implicit_query
+    reference: ''
+    task_template: true
+  df7bc2ee-686c-4826-ad84-3a056a2da4d4: !Template
+    id: df7bc2ee-686c-4826-ad84-3a056a2da4d4
+    jinja: "This query talks about the  coronavirus explicitly ? \n{{Query}}\n|||\n\
```
craffel, line 49: Does this query talk about...

craffel (changes requested):

craffel:
It's probably worth mentioning in some of the prompts that these are search engine queries (and refer to them as something other than "query" in some cases).

templates/bing_coronavirus_query_set/templates.yaml
zaidalyafeai: > It's probably worth mentioning in some of the prompts that these are search engine queries (and refer to them as something other than "query" in some cases).

Hey @craffel, thanks for the review. I don't quiet understand what you mean in this comment. Do I need to change the term query ?

zaidalyafeai:
Hey @craffel, thanks for the review. I don't quiet understand what you mean in this comment. Do I need to change the term query ?

craffel:
> Hey @craffel, thanks for the review. I don't quiet understand what you mean in this comment. Do I need to change the term query ?

Yes, I just mean instead of saying just "query" in the prompt, you should use stuff like "search on bing.com", "search engine query" etc.

craffel:
Thanks!

PR #191, opened 2021-06-11
Prompts for `gutenberg_time`

drugilsberg:
The dataset is super interesting. I initially tried to use the `is_ambiguous` flag to create more interesting prompts where the there was uncertainty on the time of the day (e.g., 5:00 vs. 17:00), but unfortunately after looking at the first 100 examples I realized that it is not reliable. I decided to use the time as is reported, because it seemed more accurate.

### DIFF
--- a/templates/gutenberg_time/templates.yaml
+++ b/templates/gutenberg_time/templates.yaml
@@ -1,68 +1,68 @@
-dataset: gutenberg_time
-templates:
-  06dce7dd-ae32-4acb-a1c8-6a01303b577b: !Template
-    id: 06dce7dd-ae32-4acb-a1c8-6a01303b577b
-    jinja: "Given the following text. What time reference is reported in the text?\n\
-      \n{{tok_context}}\n\nThe time reference reported is \n|||\n{{time_phrase}}"
-    name: asking_the_time_reference_before
-    reference: Asking the time reference before the text
-    task_template: false
-  1e880fc4-6df7-4cab-8658-82cae44135cf: !Template
-    id: 1e880fc4-6df7-4cab-8658-82cae44135cf
-    jinja: "{{tok_context}}\n\nGiven the previous text. What time reference is reported\
-      \ in the text?\n\nThe time reference reported is \n|||\n{{time_phrase}}"
-    name: asking_the_time_refence_after
-    reference: ''
-    task_template: false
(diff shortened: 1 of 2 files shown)


templates/gutenberg_time/templates.yaml
```diff
@@ -0,0 +1,101 @@
...
+      indicate?
+
+
+      {{tok_context}}
+
+
+      |||
+
+
+      It indicates to {{hour_reference}}:00.'
```
VictorSanh, line 63: no need for "to" i believe

VictorSanh (changes requested):
- remove `:00` as it is incorrect quite frequently
Capture d’écran 2021-06-11 à 3 38 27 PM

- when the question is simple (what time), no need to write "It indicates .." or "The time reference reported is" and go directly to the answer as the preceding text doesn't provide much signal

VictorSanh:
> * when the question is simple (what time), no need to write "It indicates .." or "The time reference reported is" and go directly to the answer as the preceding text doesn't provide much signal

Actually a better strategy would be to move the `|||` and include the "in indicates.." in the prompt and leave only the prediction in the label part. This paradigm greatly simplifies evaluation in the general case. 
So for instance, this would look like `It indicates to ||| {{hour_reference}}:00.'` instead of `||| It indicates to {{hour_reference}}:00.'`

drugilsberg:
Thanks for the review @VictorSanh, I implemented the suggested changes. What should I do with the templates: `asking_the_time_reference_and_actual_time_before`, `asking_the_time_reference_and_actual_time_after`?
There it's not clear how to only leave the phrase and the reference after `|||`. Are they ok as is? Should I change them?

VictorSanh (approved):
looks good to me, merging now, thank you @drugilsberg , please update the spreadsheet!

Nothing much to do for `asking_the_time_reference_and_actual_time_before` and `asking_the_time_reference_and_actual_time_after`, let leave it like this

I replaced "time" by "hour (between 0 and 23)" when necessary

drugilsberg:
> looks good to me, merging now, thank you @drugilsberg , please update the spreadsheet!
> 
> Nothing much to do for `asking_the_time_reference_and_actual_time_before` and `asking_the_time_reference_and_actual_time_after`, let leave it like this
> 
> I replaced "time" by "hour (between 0 and 23)" when necessary

Thanks to you will update now.

PR #192, opened 2021-06-11
Prompts for `math_dataset/{algebra__linear_{1,2}d,algebra__linear_{1,2}d_composed`}

drugilsberg:
Started with the templates on `algebra__linear_1d`. I worked only on this subset, because I was unsure about the templates I was generating. In case they are ok, I can work on additiional subsets after merging the PR, or even in this very same PR. I would like to have some feedback from maintainers on this.

### DIFF
--- a/templates/climate_fever/templates.yaml
+++ b/templates/climate_fever/templates.yaml
@@ -1,0 +1,87 @@
+dataset: climate_fever
+templates:
+  82c484bd-2ed7-4ee0-aaee-2b31ac68e751: !Template
+    id: 82c484bd-2ed7-4ee0-aaee-2b31ac68e751
+    jinja: 'Considering the following claim:
+
+      {{claim}}.
+
+      Does the following statement {{"supports"}}, {{"refutes"}}, or provide {{"not
+      enough info"}} on climate change?
+
+      {{evidences[4]["evidence"].strip(".").strip(''"'')}}.
+
+      |||
+
+      {{ ["SUPPORTS", "REFUTES", "NOT_ENOUGH_INFO"][evidences[4]["evidence_label"]]
+      }}'
+    name: fifth_evidence_claim_pair
+    reference: Relation between the claim and fifth evidence pair.
+  cb78a363-fd32-4dbd-976f-b56de644ba90: !Template
+    id: cb78a363-fd32-4dbd-976f-b56de644ba90
+    jinja: 'Considering the following claim:
+
+      {{claim}}.
+
+      Does the following statement {{"supports"}}, {{"refutes"}}, or provide {{"not
+      enough info"}} on climate change?
+
+      {{evidences[1]["evidence"].strip(".").strip(''"'')}}.
+
+      |||
+
+      {{ ["SUPPORTS", "REFUTES", "NOT_ENOUGH_INFO"][evidences[1]["evidence_label"]]
+      }}'
+    name: second_evidence_claim_pair
+    reference: Relation between the claim and second evidence pair.
+  cca7b6f5-29e3-45a4-bc8b-889f5ab2fc13: !Template
+    id: cca7b6f5-29e3-45a4-bc8b-889f5ab2fc13
+    jinja: 'Considering the following claim:
+
+      {{claim}}.
+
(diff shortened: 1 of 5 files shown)


VictorSanh (changes requested):
Thank you! looks great!

A few comments:
- please move the `|||` so that you isolate the label from the prompt. Also no need for `.` at the end
- As far as I understand, all of them are mapping one to one with the original task, so you should always have the flag `task_template` set to `True` which is the case for none of your templates right now

Feel free to continue in the same PR once you have made the changes

drugilsberg:
> Thank you! looks great!
> 
> A few comments:
> 
>     * please move the `|||` so that you isolate the label from the prompt. Also no need for `.` at the end
> 
>     * As far as I understand, all of them are mapping one to one with the original task, so you should always have the flag `task_template` set to `True` which is the case for none of your templates right now
> 
> 
> Feel free to continue in the same PR once you have made the changes

Thanks will address the comments and continue

drugilsberg:
Addressed comments on the `algebra__linear_1d` subset, moving now to another one

drugilsberg:
Not sure why the check is failing, it complains that:  `ValueError: Template for dataset math_dataset/algebra__linear_1d with uuid 378fed28-b09c-4549-bc04-5bad70ce3cbe has duplicate definition`.
But I don't find any duplication. Am I missing something?
I even checked for a conflicting identifier in other templates but could not find any.
```console
(promptsource)  $ pwd                                              
/Users/tte/Code/open_source/drugilsberg/promptsource/templates
(promptsource)  $ grep -iR 378fed28-b09c-4549-bc04-5bad70ce3cbe .
./math_dataset/algebra__linear_1d/templates.yaml:  378fed28-b09c-4549-bc04-5bad70ce3cbe: !Template
./math_dataset/algebra__linear_1d/templates.yaml:    id: 378fed28-b09c-4549-bc04-5bad70ce3cbe
```

drugilsberg:
> Not sure why the check is failing, it complains that: `ValueError: Template for dataset math_dataset/algebra__linear_1d with uuid 378fed28-b09c-4549-bc04-5bad70ce3cbe has duplicate definition`.
> But I don't find any duplication. Am I missing something?
> I even checked for a conflicting identifier in other templates but could not find any.
> 
> ```
> (promptsource)  $ pwd                                              
> /Users/tte/Code/open_source/drugilsberg/promptsource/templates
> (promptsource)  $ grep -iR 378fed28-b09c-4549-bc04-5bad70ce3cbe .
> ./math_dataset/algebra__linear_1d/templates.yaml:  378fed28-b09c-4549-bc04-5bad70ce3cbe: !Template
> ./math_dataset/algebra__linear_1d/templates.yaml:    id: 378fed28-b09c-4549-bc04-5bad70ce3cbe
> ```

Issue solved, it was a redefintion of the `jinja` field, probably caused by an erroneous copy/paste :)
Going on with the other subsets now...

drugilsberg:
image
while inspecting some templates I'm creating for more complex subsets I realized that `*` and `-` symbols are causing some rendering problems (italic and itemization), is this problem? I don't think it should matter since the raw text used fro prompting will be exactly the same, but I just wanted to bring this to maintainers' attention.

drugilsberg:
Templates for (6 templates each):
- `algebra__linear_1d`
- `algebra__linear_1d_composed`
- `algebra__linear_2d`
- `algebra__linear_2d_composed`

VictorSanh:
> image
> 
> while inspecting some templates I'm creating for more complex subsets I realized that `*` and `-` symbols are causing some rendering problems (italic and itemization), is this problem? I don't think it should matter since the raw text used fro prompting will be exactly the same, but I just wanted to bring this to maintainers' attention.

Yes don;t worry too much about rendering: it uses `streamlit.write` which can be confused about whether you are using markdown notation...

VictorSanh (approved):
Awesome! thank you @drugilsberg 
This dataset is 🔥
As you said, I'll merge it for now, we can come back to it later for the other subsets (the other are incredibly interesting too!). (You can use the same branch for the other subsets)

PR #193, opened 2021-06-11
has_part

nihalnayak:
template for has_part

### DIFF
--- a/templates/has_part/templates.yaml
+++ b/templates/has_part/templates.yaml
@@ -1,0 +1,47 @@
+dataset: has_part
+templates:
+  03f9060e-df5f-41ed-8cd4-4de7be1538f5: !Template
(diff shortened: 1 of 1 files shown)


nihalnayak:
This is dataset is not a task on its own. I'm assuming the task is to predict the confidence score of the two tuples. I've indicated accordingly in the task template. Let me know if I should remove it.

VictorSanh:
This dataset is very puzzling to me.
I don't think we should include it cause it's very difficult to prompt it.
It's a database so it's not targeting a task in particular.
From what I understand, the score serves as a filter for high precision pairs: the minimum value on all the datasets is `0.9985001087188721`, so we won't be able to learn anything relevant (rounding always lead to 1).
I initially thoughtt of predicting what is included in what but that seems very noisy...
So closing this PR unless you have a strong conviction on it!

PR #194, opened 2021-06-11
Allow the python function zip() to be used.

tae898:
Some people, including me, thought that the python function `zip()` is useful to be used in the templates.


stephenbach:
LGTM, thanks!

PR #195, opened 2021-06-11
Added templates for web_questions

rbawden:


templates/web_questions/templates.yaml
```diff
@@ -0,0 +1,45 @@
...
+      answer: "{{answers[0]}}"{% endif %}. Can you suggest a question? ||| {{ question|capitalize
+      }}'
+    name: suggest-question
+    reference: ''
+    task_template: false
+  d5523864-cc61-4d1b-ac67-80605dc59640: !Template
+    id: d5523864-cc61-4d1b-ac67-80605dc59640
+    jinja: 'The question is "{{ question }}" and the possible answers are: {% if answers|length
+      > 1 %}"{{ answers[0:-1]|join(''", "'') }}" and "{{ answers[-1] }}"{% else %}"{{answers[0]}}"{%
+      endif %}. How many answers are there? ||| {{ answers|length }}
```
craffel, line 41: I like that you're coming up with new tasks, but I think this one might be too easy.

craffel (changes requested):
Can you add one more template which is just {{question}} ||| {{answer}}?

templates/web_questions/templates.yaml
rbawden: Shall I remove it or try to make it more complex somehow (for example reformulating as "I think it might be" instead of "the possible answers are")?

craffel: I sort of feel like you are either listing all the answers, or you aren't. If you list all the answers, it's just counting, and if you don't list all the answers, it's kind of a total guess (since there are a crazy number of ways to state someone's name, for example).

rbawden: I might have been unclear, but I wasn't suggesting not listing all the answers, which of course would just be impossible. I was suggesting making the answer less straightforward by not having the repetition of "answers" in "the possible answers" and "how many answers are there". I agree counting should be easy (at least for small numbers) but then it could be interesting to see if the model can do something we expect it to find easy to do

craffel: Sorry, wasn't saying you were suggesting to not include the list of answers - was just saying that wasn't an option. The worry about too-easy tasks is that fundamentally we only have a limited amount of training budget available, so it might be better spent training on task with more learning signal.

rbawden: No worries :-) Has anyone else include prompts like this? If not, it could be interesting to have a couple of prompts that test things like this (things that we think should be easy, but could benefit from being demonstrated), given the large number of prompts overall. I've reformulated it to make it slightly trickier - but can just remove it otherwise.

craffel:
Looks good, thanks!

PR #196, opened 2021-06-11
added scan dataset

teelinsan:
Hi all, added prompts for the [SCAN](https://github.com/brendenlake/SCAN) dataset.

### DIFF
--- a/templates/scan/addprim_jump/templates.yaml
+++ b/templates/scan/addprim_jump/templates.yaml
@@ -1,0 +1,162 @@
+dataset: scan
+subset: addprim_jump
+templates:
+  10f81c2c-5907-4065-ba83-a65812a8678b: !Template
+    id: 10f81c2c-5907-4065-ba83-a65812a8678b
+    jinja: '{{ actions }}
+
+
+      Given the SCAN actions above, produce the corresponding correct sequence of
+      commands in natural language.
+
+      |||
+
+      {{ commands }}'
+    name: reversed_scan_affirmative_bottom
+    reference: This is a version of SCAN with the task reversed (predict the commands
+      in natural language given the SCAN actions)
+    task_template: false
+  23eeb3e4-7ae6-48b4-9d04-e712613860a6: !Template
+    id: 23eeb3e4-7ae6-48b4-9d04-e712613860a6
+    jinja: '{{ actions }}
+
+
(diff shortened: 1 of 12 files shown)


VictorSanh (changes requested):
Thank you @teelinsan , SCAN is a fun one!

A few comments:
- Can you indicate the `task_template` flag in the yaml file? I believe it should be all the templates that are not "reversed" (I.e. taking the natural language and predict the instructions)
- It's not realistic to expect the model to know things like "SCAN" or the commands. For all templates, please include a description like `the possible actions are: I_TURN_RIGHT: turn right, I_WALK: walk straight, etc.`, and copy-paste it for all templates (not necessarily at the same position depending on the template) + protect the commands: `{{"I_TURN_RIGHT"}}`

teelinsan:
Hi @VictorSanh 

- Thanks, I've accidentally missed the task_template flag (now fixed). I also decided to remove the "reversed scan" templates since I realized that the reversed task wasn't well defined (e.g., I_TURN_RIGHT I_TURN_RIGHT could have been both "turn opposite right" and "turn right twice")
- If the model is trained on the training dataset of SCAN, I think it can be realistic to ask the SCAN actions without giving the explicit mapping to the model. At the end we are asking the model to translate the commands in natural language to the specific SCAN actions and, like in Machine Translation, we don't give a word-to-word mapping. However I think that an explicit mapping can be useful and might be interesting to do comparison experiments, so I added other prompt templates that include this explicit mapping and kept also the previous templates without mapping.

VictorSanh (approved):
thank you, merging now, please update the spreadsheet!

VictorSanh:
> * If the model is trained on the training dataset of SCAN, I think it can be realistic to ask the SCAN actions without giving the explicit mapping to the model. At the end we are asking the model to translate the commands in natural language to the specific SCAN actions and, like in Machine Translation, we don't give a word-to-word mapping. However I think that an explicit mapping can be useful and might be interesting to do comparison experiments, so I added other prompt templates that include this explicit mapping and kept also the previous templates without mapping.

Thank you!
Yes I agree. My rationale is that you want the model to reason over the inputs and not necessarily learn the specifics of the dataset (i.e. the mapping between individual command and what they mean). Ultimlately, these commands could be something like `command 3`, `command 5`, `command 78`, and it wouldn't change the task that much (it turns out that the command codes are in semi-formal language).

PR #197, opened 2021-06-11
Templates for Limit dataset

arnaudstiegler:
Task template: find all entities
For this task, should return a specific text output if there are no entities

Side tasks:
- is there any moving entity?
- how many moving entities?
- Is there more than X moving entities?

Sometimes, created both question and instruction format.

### DIFF
--- a/templates/limit/templates.yaml
+++ b/templates/limit/templates.yaml
@@ -82,7 +82,7 @@ templates:
     id: 957deab1-7570-4cbf-a31f-55bfad5212a7
     jinja: "Name the motion entities in the following sentence. Respond {{\"'No entity'\"\
       }} if you can't find any. \n\n{{sentence}}\n\n|||\n\n{% if (motion_entities\
-      \ | length) == 0 %}\n{{ \"'No entity'\"}}\n{% else %}\n{{motion_entities | map(attribute=\"\
+      \ | length) == 0 %}\n{{\"No entity\"}}\n{% else %}\n{{motion_entities | map(attribute=\"\
       entity\") | join(\", \")}}\n{% endif %}"
     name: find_entities_affirm
     reference: ''
--- a/templates/limit/templates.yaml
+++ b/templates/limit/templates.yaml
@@ -1,0 +1,148 @@
+dataset: limit
+templates:
+  0898caf1-f3e4-493f-a838-84a678176c14: !Template
+    id: 0898caf1-f3e4-493f-a838-84a678176c14
+    jinja: '{{sentence}}
+
+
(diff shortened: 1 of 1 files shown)


templates/limit/templates.yaml
```diff
@@ -0,0 +1,147 @@
...
+      {% endif %}'
+    name: last_moving_entity
+    reference: ''
+    task_template: false
+  3b88c578-db77-4fd0-ad50-c78a39197ce5: !Template
+    id: 3b88c578-db77-4fd0-ad50-c78a39197ce5
+    jinja: '{{sentence}}
+
+
+      Is there any motion entities in the sentence?
```
VictorSanh, line 25: are there

templates/limit/templates.yaml
```diff
@@ -0,0 +1,147 @@
...
+
+
+      |||
+
+
+      {% if (motion_entities  | length) > 0 %}
+
+      {{ (motion_entities | sort(attribute="start_index") | first)["entity"] }}
+
+      {% endif %}'
```
VictorSanh, line 105: add else condition to say "no entities in motion" if none

templates/limit/templates.yaml
```diff
@@ -0,0 +1,147 @@
...
+
+      What is the last moving entity mentioned in the sentence if any?
+
+      |||
+
+      {% if (motion_entities  | length) > 0 %}
+
+      {{ (motion_entities | sort(attribute="start_index") | last)["entity"] }}
+
+      {% endif %}'
```
VictorSanh, line 16: add else condition to say "no entities in motion" if none

VictorSanh (changes requested):
Thank you @arnaudstiegler!

A few minor changes in comments.
"motion entities" sounds weird. They seem to use "entities in motion" much more frequently in the paper. please change it

templates/limit/templates.yaml
arnaudstiegler: Added. For other prompts with a similar possible answer, I literally told what to output in case there's no entity (i.e "Return no entity if there's none" in the prompt). See `find_entities_question` below.
Do you think it's desirable or not?

templates/limit/templates.yaml
arnaudstiegler: Done

arnaudstiegler:
Thanks for the review! Made the requested changes.
I removed all mentions of  "motion entities", and there's now a mix of `entity in motion` and `moving entity` to have some variability.

VictorSanh (approved):
thanks, merging now, please update the spreadsheet!

PR #199, opened 2021-06-11
Add template for emo dataset

trishalaneeraj:


templates/emo/templates.yaml
```diff
@@ -0,0 +1,60 @@
...
+      This was a dialogue exchange between 2 people who took turns to talk. Given
+      the context, which of the following options best describes the emotion that
+      last speaker is feeling? {{"happy"}}, {{"sad"}}. {{"angry"}}, or {{"something
+      else"}}?
+
+
+      |||
+
+
+      {{ ["others", "happy", "sad", "angry"] [label] }}'
```
VictorSanh, line 57: for all templates, please change to "something else" as it is the choice described in the prompt

templates/emo/templates.yaml
```diff
@@ -0,0 +1,60 @@
...
+    task_template: true
+  f686cef0-6174-466d-b87d-672aaf5f9caa: !Template
+    id: f686cef0-6174-466d-b87d-672aaf5f9caa
+    jinja: '"{{text}}"
+
+
+      This was a dialogue exchange between 2 people who took turns to talk. Given
+      the context, which of the following options best describes the emotion that
+      last speaker is feeling? {{"happy"}}, {{"sad"}}. {{"angry"}}, or {{"something
+      else"}}?
```
VictorSanh, line 51: That doesn't correctly explain the task: you don't see it's a dialogue, there is no context/dialogue history.
Better: `This is an extract of a dialogue between two people. Which of the following options best describes the emotion the speaker is feeling?`

VictorSanh (changes requested):
thank you @trishalaneeraj, this one is weird because the dialogue is concatenaneted (and not separated in any way...).
One small comment.
please ping me when it is addressed and I'll merge!

trishalaneeraj:
@VictorSanh thanks! I've made the change.

VictorSanh (approved):
thank you!
merging now, please update the spreadsheet!

PR #201, opened 2021-06-11
Add acronym_identification templates

gchhablani:
**Notes for the map-based templates**:
- Most examples have the form `Expansion (Abbreviation)` and I am assuming that the first letter of the first word and the first letter of the last word in the expansion are present in the abbreviation - Please check the notable examples below for this reasoning.
- Since there is no direct mapping provided, the map-based templates are slightly tricky, but are almost perfect:
   - I tested a lot of examples (at least a hundred) for this template. Most work fine. Please report any errors.
   - The notable examples in the `train` set, with which the template works: 1739, 12853, 13676, 12078.
   - Some examples which may be erroneous in the dataset itself: 4136, 423, 11749, 9940.
   - The only error I found: 9919. Since I'm checking for the first and last word's first character, another acronym with the same characters may also be considered if an expansion of the same isn't found. Here is the output:
```
Given the following tokens, find the abbreviations and their expansions, if present:

Notice that if , then the WCSP is nothing but the famous Constraint Satisfaction Problem or CSP ( not the Max - CSP ) .

Y

WCSP : Constraint Satisfaction Problem

CSP :
```
What happens is - I get `WCSP` first, I find expansions and I get `Constraint Satisfaction Problem`. I see `C` and `P` are in `WCSP`, so it should match. Note that I cannot check the first character of the abbreviation, because then one of the "notable examples" will fail. Also cannot match the first and last character of the acronym because, then, another one of them will fail. Please let me know if this needs to be changed.
  
I find this task to be very interesting from the model's perspective - finding the mapping, finding the abbreviation, and finding the expansion.

### DIFF
--- a/templates/acronym_identification/templates.yaml
+++ b/templates/acronym_identification/templates.yaml
@@ -1,9 +1,47 @@
 dataset: acronym_identification
 templates:
+  64f438f2-9968-459f-82d2-24bad632b358: !Template
+    id: 64f438f2-9968-459f-82d2-24bad632b358
+    jinja: "{% set abbr_exp_dict = namespace(value = {}) %}\n{% set abbr_string=namespace(value='')\
+      \ %}\n{% set exp_string=namespace(value='')%}\n \n{% for label_idx in range(labels|length)\
+      \ %}\n  {% if labels[label_idx] == 0 %}{# Long Beginning #}\n    {% set exp_string.value\
+      \ = tokens[label_idx] %}{# Create new long string #}\n  {% elif labels[label_idx]\
+      \ == 1 %}{# Short Beginning #}\n    {% if abbr_string.value!='' and abbr_string.value\
+      \ not in abbr_exp_dict.value.keys()%}{# Some string already present #}\n   \
+      \   {% set _dummy = abbr_exp_dict.value.update({abbr_string.value:''}) %}{#\
+      \ Discard this string as a new short string is coming #}\n    {% endif %}\n\
+      \    {% set abbr_string.value = tokens[label_idx] %}{# Create new short string\
+      \ #}\n  {% elif labels[label_idx] == 2 %}{# Long Intermediate #}\n    {% set\
+      \ exp_string.value = exp_string.value+' '+tokens[label_idx] %}{# Update existing\
+      \ string #}\n  {% elif labels[label_idx] == 3 %}{# Short Intermediate #}\n \
+      \   {% set abbr_string.value = abbr_string.value+tokens[label_idx] %}{# Update\
+      \ existing string #}\n  {% else %}{# Other #}\n    {# Both non-empty, and first\
+      \ characters match #}\n    {% if abbr_string.value!='' and exp_string.value!=''\
+      \ and exp_string.value.split()[0][0]|lower in abbr_string.value|lower and exp_string.value.split()[-1][0]|lower\
+      \ in abbr_string.value|lower%}\n      {# Update both the dictionaries #}\n \
+      \     {% set _dummy = abbr_exp_dict.value.update({abbr_string.value:exp_string.value})\
+      \ %}\n      {# Empty both the strings #}\n      {% set abbr_string.value= ''\
+      \ %}\n      {% set exp_string.value= '' %}\n    {% endif %}\n  {% endif %}\n\
+      {% endfor %}\n{# Both non-empty, and first characters match #}\n{% if abbr_string.value!=''\
+      \ and exp_string.value!='' %}\n  {% if exp_string.value.split()[0][0]|lower\
+      \ in abbr_string.value|lower and exp_string.value.split()[-1][0]|lower in abbr_string.value|lower\
+      \ %}\n    {# Update both the dictionaries #}\n    {% set _dummy = abbr_exp_dict.value.update({abbr_string.value:exp_string.value})\
+      \ %}\n  {% elif abbr_exp_dict.value.items()|length==0 and exp_abbr_dict.value.items()|length==0\
+      \ %}\n    {% set _dummy = abbr_exp_dict.value.update({abbr_string.value:exp_string.value})\
+      \ %}\n  {% endif %}\n{% else %}\n  {% if abbr_string.value!=''%}\n    {% if\
+      \ abbr_string.value not in abbr_exp_dict.value.keys() %}\n      {% set _dummy\
+      \ = abbr_exp_dict.value.update({abbr_string.value:''}) %}\n    {% endif %}\n\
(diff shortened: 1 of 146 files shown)


templates/acronym_identification/templates.yaml
```diff
@@ -0,0 +1,161 @@
...
+      \ %}\n  {% elif exp_abbr_dict.value.items()|length==0 %}\n    {% set _dummy\
+      \ = exp_abbr_dict.value.update({exp_string.value:abbr_string.value}) %}\n  {%\
+      \ endif %}\n{% else %}\n  {% if exp_string.value!='' %}\n    {% if exp_string.value\
+      \ not in exp_abbr_dict.value.keys() %}\n      {% set _dummy = exp_abbr_dict.value.update({exp_string.value:''})\
+      \ %}\n    {% endif %}\n  {% endif %}\n{% endif %}\n{% if exp_abbr_dict.value.items()|length!=0\
+      \ %}\n{% set random_exp = exp_abbr_dict.value.keys()|list|choice %}\nGiven the\
+      \ following tokens, find the abbreviation for {{random_exp}}:\n \n{{tokens|join('\
+      \ ')}}\n|||\n{% if random_exp in exp_abbr_dict.value.keys() and exp_abbr_dict.value[random_exp]!=''\
+      \ %}\n{{exp_abbr_dict.value[random_exp]}}\n{% else %}\nUnclear\n{% endif %}\n\
+      {% else %}\n|||\n{% endif %}"
```
VictorSanh, line 158: bring the `|||` out of the your if condition. if the template is blank, it should return an empty string `` and then through the `.split("|||")` it should be `[""]`

templates/acronym_identification/templates.yaml
```diff
@@ -0,0 +1,161 @@
...
+      {% endfor %}\n{# Both non-empty, and first characters match #}\n{% if abbr_string.value!=''\
+      \ and exp_string.value!='' %}\n  {% if exp_string.value.split()[0][0]|lower\
+      \ in abbr_string.value|lower and exp_string.value.split()[-1][0]|lower in abbr_string.value|lower\
+      \ %}\n    {# Update both the dictionaries #}\n    {% set _dummy = abbr_exp_dict.value.update({abbr_string.value:exp_string.value})\
+      \ %}\n  {% elif abbr_exp_dict.value.items()|length==0 %}\n    {% set _dummy\
+      \ = abbr_exp_dict.value.update({abbr_string.value:exp_string.value}) %}\n  {%\
+      \ endif %}\n{% else %}\n  {% if abbr_string.value!=''%}\n    {% if abbr_string.value\
+      \ not in abbr_exp_dict.value.keys() %}\n      {% set _dummy = abbr_exp_dict.value.update({abbr_string.value:''})\
+      \ %}\n    {% endif %}\n  {% endif %}\n{% endif %}\n{% set random_abbr = abbr_exp_dict.value.keys()|list|choice\
+      \ %}\nGiven the following tokens, find the expansion of {{random_abbr}}:\n \n\
```
VictorSanh, line 35:
```suggestion
      \ %}\nGiven the following tokens, find the expansion of {{random_abbr}}. Return "Unclear" if the expansion can't be found.\n \n\
```

VictorSanh: please do similar modification for `find_abbreviation`

templates/acronym_identification/templates.yaml
```diff
@@ -0,0 +1,161 @@
...
+      \ in abbr_string.value|lower and exp_string.value.split()[-1][0]|lower in abbr_string.value|lower\
+      \ %}\n    {# Update both the dictionaries #}\n    {% set _dummy = abbr_exp_dict.value.update({abbr_string.value:exp_string.value})\
+      \ %}\n  {% elif abbr_exp_dict.value.items()|length==0 %}\n    {% set _dummy\
+      \ = abbr_exp_dict.value.update({abbr_string.value:exp_string.value}) %}\n  {%\
+      \ endif %}\n{% else %}\n  {% if abbr_string.value!=''%}\n    {% if abbr_string.value\
+      \ not in abbr_exp_dict.value.keys() %}\n      {% set _dummy = abbr_exp_dict.value.update({abbr_string.value:''})\
+      \ %}\n    {% endif %}\n  {% endif %}\n{% endif %}\n \nGiven the following tokens,\
+      \ find the abbreviations and their expansions, if present:\n \n{{tokens|join('\
+      \ ')}}\n|||\n{% for item, value in abbr_exp_dict.value.items() %}\n{{item}}\
+      \ : {{value}}\n{%endfor%}"
```
VictorSanh, line 119: maybe an "unclear" if can't find the abbeviation?
Capture d’écran 2021-06-16 à 3 16 24 PM

templates/acronym_identification/templates.yaml
VictorSanh: and for `list_expansions` ("No expansions found")

VictorSanh (changes requested):
waoo this one is very hard!
thanks for taking it @gchhablani !

Made a few suggestions/changes

templates/acronym_identification/templates.yaml
```diff
@@ -0,0 +1,161 @@
...
+    reference: Given the tokens, find the expansion of an abbreviation in the tokens.
+    task_template: false
+  81babc83-18cd-4eed-a343-8ede56b21df5: !Template
+    id: 81babc83-18cd-4eed-a343-8ede56b21df5
+    jinja: "Given the BIO encoding as follows:  \"{{\"B-short\"}}\" and \"{{\"I-short\"\
+      }}\" represent the beginning and intermediate tokens for abbreviations.\"{{\"\
+      B-long\"}}\" and \"{{\"I-long\"}}\" represent the beginning and intermediate\
+      \ tokens for expansions of the abbreviations. All other tokens are represented\
+      \ by \"{{\"O\"}}\". \nGenerate comma-separated BIO encoding for the following\
+      \ comma-separated tokens: \n\n{{tokens|join(', ')}}.\n|||\n{% for label in labels\
```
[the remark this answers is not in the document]
gchhablani: All the tokens I have seen usually have their own ".". I think this could be redundant?

templates/acronym_identification/templates.yaml
VictorSanh: seems like it's persisting:
Capture d’écran 2021-06-17 à 2 27 37 PM

It should literally be displaying `∅∅∅ *Blank result*` (see https://github.com/bigscience-workshop/promptsource/blob/main/promptsource/promptsource.py#L303)
and I think right now there are a lot of "\n" in what's being returned...

VictorSanh:
note: weird formatting of jinja. merging for now, will fix later

PR #202, opened 2021-06-11
Adding ASNQ

patrick-s-h-lewis:
ASNQ: https://huggingface.co/datasets/asnq

This dataset contains some:

1) Full Sentence Question answering 
2) answer-unconditional Question generation from sentences

NOTE: this dataset is very very imbalanced. (The original task is a bit different, has a huge number of datapoints - millions - and has 99% negatives. Here, I have only written templates for positives. The vast majority of the examples in the viewer will look thus empty, but there are useful examples to be generated here) . 

See the discussion on slack: https://huggingface.slack.com/archives/C023Z96L7B4/p1623427960130100


stephenbach:
LGTM! Thanks!

PR #205, opened 2021-06-11
bug-fixes, task template checkbox updated

sbmaruf:
Earlier there was a bug in the template.
The templates were written targeting specific labels. For some cases, if the condition fails, it stills augments text on the target (label).
Now, if the condition fails no text will be generated both for input and label.


stephenbach:
Haha, I almost bugged you about some of these, but then I figured we could just check whether either input or output is blank. These are nicer. Thanks for cleaning them up and adding the task template annotations!

sbmaruf:
Hi! @stephenbach, bug me anytime for anything. Would love to go back to change or re-organize if needed.

PR #208, opened 2021-06-12
add: templates for multi_nli_mismatch

nihalnayak:
PR contains the template for MNLI mismatched. Copied the templates from MNLI.

### DIFF
--- a/templates/multi_nli_mismatch/templates.yaml
+++ b/templates/multi_nli_mismatch/templates.yaml
@@ -6,7 +6,7 @@ templates:
       \ == \"entailment\" %} \nYes\n{% elif label == \"neutral\" %}\nMaybe\n{% else\
       \ %}\nNo\n{% endif %}"
     name: GPT3-style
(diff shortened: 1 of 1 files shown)


stephenbach:
Templates look good, but what's the difference between this dataset and MNLI? I can't seem to find that documented anywhere.

nihalnayak:
In this subset of the MNLI dataset, they include examples from genres that are not included in the training set. Here's a quote from their dataset paper (https://cims.nyu.edu/~sbowman/multinli/paper.pdf).
```
All of the genres appear in the test and development sets, but only five
are included in the training set. Models thus can
be evaluated on both the matched test examples,
which are derived from the same sources as those
in the training set, and on the mismatched examples, which do not closely resemble any of those
seen at training time.
```
This subset was the test set in the RepEval 2017 shared task (https://repeval2017.github.io/shared/). We could include this as a test set or evaluation set.

stephenbach:
Oh great, if it's different data then copying the templates sounds good.

PR #210, opened 2021-06-12
Evidence infer treatment

debajyotidatta:
This dataset is very interesting because the prompts can be varied based on the intervention, comparator, outcome, and the final label of the treatment. Since the number of annotations can vary and the number of sub annotations can also vary, a random subset is selected as the prompt for each example.

### DIFF
--- a/.github/workflows/show_new_templates.yml
+++ b/.github/workflows/show_new_templates.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python 3.8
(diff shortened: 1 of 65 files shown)


stephenbach (changes requested):
Thank you! These are interesting!

Would you please change these to use the `choice` filter, instead of `random`?

Also, these produce really long prompts that are unlikely to fit in the input to the model. (2048 byte pairs is a reasonable estimate.) We'll try to heuristically trim them, but if there's something smarter that you can do (such as cutting the sections that are least likely to be useful) that would be more likely to produce prompts that preserve your meaning.

debajyotidatta:
@stephenbach: After going through multiple examples, I think the heuristic of using the first 1500 characters of the text and the last 300 characters should provide all the information for making inferences about the data. These sections generally contain the abstract, introduction, key findings, and conclusion.

stephenbach:
Wonderful thanks! Would you also just replace the `|random` filters with `|choice`? Then we should be good to go.

debajyotidatta:
@stephenbach : Replaced random with choice. Sorry about that.

stephenbach:
Great, thank you!

PR #211, opened 2021-06-12
covid_qa_castorini

arunraja-hub:


stephenbach:
Thanks! I'm not sure all of these templates make sense to me. For example, in all_answers and all_papers, a lot more context seems to be needed to answer questions and find titles. There's no article provided. For classify_question, the categories seem very narrow and specialized. Perhaps choices need to be provided? This dataset only has 26 training examples, so I wouldn't over optimize it (t's ok if we don't get 5 templates).

arunraja-hub:
Thank you. I have removed those templates. Regarding classify_question, there is no list of possible categories provided so I have removed it as well.

stephenbach:
Ok, sounds good. Thanks!

PR #212, opened 2021-06-12
Add Quora Duplicate Question Classification Dataset

blackrosedragon2:
I am a little unsure on if the `quora_localization_1` is correct or not. Do let me know if there are any other things I need to fix!

### DIFF
--- a/.github/workflows/show_new_templates.yml
+++ b/.github/workflows/show_new_templates.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python 3.8
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -270,6 +270,13 @@ do_something
 do_something_else
 {% endif %}
 ```
+- Using `zip()` to zip multiple lists
+```jinja
+{% for a, b in zip(list_A, list_B) %}
+do_something_with_a_and_b
+{% endfor %} 
+```
+
 
 Jinja includes lots of complex features but for most instances you likely only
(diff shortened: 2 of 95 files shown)


stephenbach (changes requested):
Thanks! These are nice! I would just take out quora_basic_1 and quora_localization_1. quora_basic_1 conflicts with quora_basic_2 because they have the same input but different outputs (and I think yes/no is a more natural answer).

And I would take out quora_localization_1 because I don't think there's enough context for the model to understand what it is you want it to do. Perhaps you could change it to a paraphrasing task for the ones that are duplicates? (Just output nothing for examples that are not duplicates.)

Also any template that does duplicate detection should be marked as a task template.

Thank you!

blackrosedragon2:
> Thanks! These are nice! I would just take out quora_basic_1 and quora_localization_1. quora_basic_1 conflicts with quora_basic_2 because they have the same input but different outputs (and I think yes/no is a more natural answer).
> 
> And I would take out quora_localization_1 because I don't think there's enough context for the model to understand what it is you want it to do. Perhaps you could change it to a paraphrasing task for the ones that are duplicates? (Just output nothing for examples that are not duplicates.)
> 
> Also any template that does duplicate detection should be marked as a task template.
> 
> Thank you!

On it !

stephenbach:
Looks great! Would you just check the task template box for the other templates that do duplicate detection as well? Thank you!

stephenbach:
Sorry to drag this out, I think all the duplicate detection ones should be marked true? Only the paraphrasing one seems different from the original task?

blackrosedragon2:
> Sorry to drag this out, I think all the duplicate detection ones should be marked true? Only the paraphrasing one seems different from the original task?

I'm sorry, I think I misunderstood the checkbox button! I will make the fix now.

stephenbach:
Awesome, thank you!

PR #213, opened 2021-06-12
Prompts for aqua_rat

shanyas10:
Prompts for aqua_rat


stephenbach:
Thanks, these are great! I'm just wondering if we can improve the rendering of the options? Right now it dumps it out as a Python list, plus there's no space between a) and the answer because that's how it's stored in the dataset. Maybe we could write a little more Jinja to render that nicely with spaces in the appropriate places? I think that will increase the chance it gets tokenized in a sensible way. Thanks!

shanyas10:
Hi @stephen, my bad! I somehow missed that. Updated it now and also updated the task_templates value.

stephenbach:
Awesome, thank you! And thanks for catching the task template. Missed that one too.

PR #217, opened 2021-06-12
templates for xquad

shanyas10:


stephenbach:
Thanks, these are great! Would you just check the "task template" boxes for the ones that are answering the question?

stephenbach:
Great, thank you!

PR #218, opened 2021-06-12
prompts for `liar`

manandey:
Fresh PR for #87. Updated the code taking into account all the suggestions @craffel


craffel:
Thanks!

PR #221, opened 2021-06-12
ai2_arc

arunraja-hub:


stephenbach (changes requested):
Thanks! It seems that the task template annotations are missing. Would you please add them?

Also, maybe we should cut the template without choices? Looks like a lot of questions are phrased as "Which of the following..." etc. What do you think?

stephenbach:
Thank you! Just one other question: why is the template named "test" not a task template?

stephenbach (approved):
Good now, thanks!

PR #222, opened 2021-06-12
qa_srl

arunraja-hub:


templates/qa_srl/templates.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+
+      {{sentence}} |||
+
+      {{question|join(" ")|replace("_","")}} '
+    name: context_qn
+    reference: ''
+  bec2ce78-fb31-4529-8b13-240fa6c8bc88: !Template
+    id: bec2ce78-fb31-4529-8b13-240fa6c8bc88
+    jinja: 'Identify the predicate in this sentence:
```
craffel, line 41: This template results in duplicate examples since some examples in the dataset differ only in the question. I'm not sure if we have a way to deal with this, I will ask around.

craffel: Checked and we are going to do post-prompting deduping, so this looks fine.

craffel:
Thanks1

PR #223, opened 2021-06-12
Add template for Mocha dataset

yongzx:
Notes:
1. The dataset has a "minimal_pairs" split that has two candidate answers instead of one (as in the "train", "validation", "test" split). I have include if-else blocks to handle them. 
2. For the `Pick Correct Answer` template, to prevent the reference answer from always showing up as the first answer, I use the following hack: use the length of the passage to determine if we should right shift the list of answers (that includes reference and candidates). If so, perform the right shift before listing the answers out.

### DIFF
--- a/templates/mocha/templates.yaml
+++ b/templates/mocha/templates.yaml
@@ -11,6 +11,7 @@ templates:
     name: (Regression) Scoring Candidate against Reference
     reference: Similarity measure between candidate and reference answers (in a regression
       manner)
+    task_template: true
   2816084e-0193-4284-9a4f-9de4ae03e9d6: !Template
     id: 2816084e-0193-4284-9a4f-9de4ae03e9d6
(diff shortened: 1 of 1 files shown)


craffel:
It seems for the "scoring candidate" variants, the model could judge equivalency of answers without access to the context and question. If that's the case, do you want to include variants of the task where the model is just rating/classifying similarity of the answers? You could also do it with only the question and answer, and not the context sentence. WDYT?

yongzx:
> If that's the case, do you want to include variants of the task where the model is just rating/classifying similarity of the answers? You could also do it with only the question and answer, and not the context sentence.

I could definitely include these two variants!

yongzx:
@craffel Added the variants for both regression and classification tasks.

craffel:
Thanks!

PR #224, opened 2021-06-12
qa_zre

arunraja-hub:


templates/qa_zre/templates.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+      {{question}} |||
+
+      {{relation}}'
+    name: relation
+    reference: https://www.aclweb.org/anthology/K17-1034.pdf
+  5a970b88-53a0-4148-b45e-7ac410df263f: !Template
+    id: 5a970b88-53a0-4148-b45e-7ac410df263f
+    jinja: "What is a possible question that can be generated from the following context\
+      \ and answer(s)?\n\n{{context}} \n\n{{answers|join(\", \")}} |||\n{{question|replace(\"\
```
craffel, line 16: It's a little confusing to say "and answer(s)" for the examples where there are no answers. One option would be to only say "and answer(s)" when answers is non-empty, but that would create duplicates with the possible_qn template. I think it's probably best to modify the "possible_qn_with_answer" template so that it only includes "and answer(s)" when answers is non-empty and remove the "possible_qn" template.

templates/qa_zre/templates.yaml
```diff
@@ -0,0 +1,50 @@
...
+    name: subject
+    reference: ''
+  b2195890-a3c5-4e33-be4a-5e53af75e6dd: !Template
+    id: b2195890-a3c5-4e33-be4a-5e53af75e6dd
+    jinja: '{{context}}
+
+      {{question.replace("XXX",subject)}} |||
+
+      {{answers|join(", ")}}'
+    name: qa
```
craffel, line 39: You should put an if statement on this to only include examples where answers is non-empty.

templates/qa_zre/templates.yaml
```diff
@@ -0,0 +1,50 @@
+dataset: qa_zre
+templates:
+  2d6b6ec6-4cba-4a07-a0d1-f6b7cb103281: !Template
+    id: 2d6b6ec6-4cba-4a07-a0d1-f6b7cb103281
+    jinja: 'Extract the appropriate relation from the following question
```
craffel, line 5: Maybe include a variant of this task where the XXX is filled in with the subject? It's a valid task either way.

craffel (changes requested):
Mostly comments pertaining to the fact that answers can be an empty list.

craffel:
Thanks!

PR #227, opened 2021-06-13
KELM KB Triple to Text Generation Task

jason-fries:


craffel:
Hey Jason, these look good but we are targeting at least 5 prompts per dataset. Feel free to just paraphrase the existing prompts, since it's not obvious what other tasks you could create out of this dataset. Also note that it looks like this dataset often has multiple triples per example, so it's not quite right to say "**a** relational triple" e.g. 
```
{
  "sentence": "The series featured appearances from famous pioneering space scientists and explorers, and was narrated by Samuel West in the original 1999 edition, and Mark Halliley in the 2004 remastered edition.",
  "triple": "The Planets ( 1999 TV series ) publication date 01 January 1999, narrator Samuel West"
}
```

jason-fries:
Thanks @craffel! I found the triple input format to be a bit confusing, even after referring back to the [paper](https://arxiv.org/abs/2010.12688). It seems like some triples have sub-properties and those might also be encoded in the string. There is some general confusion around triple formatting in the dataset (see https://github.com/google-research-datasets/KELM-corpus/issues/1). 

I'll come up with some better language and resubmit.

craffel:
Thanks!

PR #228, opened 2021-06-13
squad_adversarial-AddSent

arunraja-hub:
I  have only added prompts for the AddSent subset because the other subset, AddOneSent, seems to be of the same format and contain the same examples as AddSent

### DIFF
--- a/templates/squad_adversarial/AddSent/templates.yaml
+++ b/templates/squad_adversarial/AddSent/templates.yaml
@@ -3,12 +3,10 @@ subset: AddSent
 templates:
   0b69eb58-0ad7-45b4-baf2-f3abac775fe2: !Template
     id: 0b69eb58-0ad7-45b4-baf2-f3abac775fe2
-    jinja: 'Generate a title for the following context-question pair:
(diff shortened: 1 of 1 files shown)


templates/squad_adversarial/AddSent/templates.yaml
```diff
@@ -0,0 +1,74 @@
+dataset: squad_adversarial
+subset: AddSent
+templates:
+  0b69eb58-0ad7-45b4-baf2-f3abac775fe2: !Template
+    id: 0b69eb58-0ad7-45b4-baf2-f3abac775fe2
+    jinja: 'Generate a title for the following context-question pair:
```
craffel, line 6: I think the title only really applies to the context - so you could either remove the question from this prompt, or create a separate prompt where the question doesn't appear.

craffel (changes requested):
Can you make it so that one or more prompts do not use the word "context"? That's not a typical word that people use outside of the context of reading comprehension systems.

craffel:
Thanks!

PR #247, opened 2021-06-14
docred

tae898:


templates/docred/templates.yaml
```diff
@@ -0,0 +1,51 @@
...
+      \ == 'LOC' -%}\n{{ner['name']}}{{\", \"}}\n{%- endfor %}\n{%- endfor %}\n\n\
+      Organizations:\n{% for ners in vertexSet -%}\n{% for ner in ners if ner['type']\
+      \ == 'ORG' -%}\n{{ner['name']}}{{\", \"}}\n{%- endfor %}\n{%- endfor %}\n\n\
+      Time:\n{% for ners in vertexSet -%}\n{% for ner in ners if ner['type'] == 'TIME'\
+      \ -%}\n{{ner['name']}}{{\", \"}}\n{%- endfor %}\n{%- endfor %}\n\nNumbers:\n\
+      {% for ners in vertexSet -%}\n{% for ner in ners if ner['type'] == 'NUM' -%}\n\
+      {{ner['name']}}{{\", \"}}\n{%- endfor %}\n{%- endfor %}\n\nMiscellaneous:\n\
+      {% for ners in vertexSet -%}\n{% for ner in ners if ner['type'] == 'MISC' -%}\n\
+      {{ner['name']}}{{\", \"}}\n{%- endfor %}\n{%- endfor %}"
+    name: type-to-entity
```
craffel, line 20: This template results in commas at the end of each type, e.g. `Locations: New Britain, New Ireland, Australia, Rabaul, Rabaul, Kavieng, Kavieng, Japan, Japanese, `. I think it's more natural to not have a comma at the end. You should be able to use the `join` jinja filter for this. Do you mind making this change?

templates/docred/templates.yaml
```diff
@@ -0,0 +1,51 @@
...
+    reference: ''
+    task_template: false
+  3ab9cfc9-3ba3-41dd-959c-60182def11af: !Template
+    id: 3ab9cfc9-3ba3-41dd-959c-60182def11af
+    jinja: "Read the following text and answer the questions.\n\nText:\n{% for sent\
+      \ in sents -%}\n{{ sent | join(\" \") }}\n{% endfor %} \n\nQuestion:\n\nAssign\
+      \ an entity type to the following entities. The choices are PER (Person), LOC\
+      \ (Location), ORG (Organization), TIME (Time), NUM (Number), and MISC (Miscellaneous).\n\
+      \nExample: \n{% for ners in vertexSet[:1] %}\n{% for ner in ners %}\n{{ner['name']}}:\
+      \ {{ner['type']}}\n{% endfor %}\n{% endfor %}\n\nNow do the same with below:\n\
```
craffel, line 30: This ends up being a bit weird because entities are duplicated in the list. Can you deduplicate?

craffel (commented):
Thanks for this. We are shooting for at least 5 prompts per dataset. Can you add some more? One option would be to choose a random single entity and ask its type, or choose a random pair and ask the relation, etc.

tae898:
> Thanks for this. We are shooting for at least 5 prompts per dataset. Can you add some more? One option would be to choose a random single entity and ask its type, or choose a random pair and ask the relation, etc.

alright sure!

templates/docred/templates.yaml
tae898: The duplication happens because multiple of the same entities can be detected throughout the text, and that's what the labels of the dataset give us. I don't think it's trivial to deduplicate them but if you have a better idea, please let me know.

craffel: I think you should be able to use this filter: https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.unique

tae898:
The code might look messy because often times I had to loop over a list of lists. Of course this could have been dealt with easily in python using list comprehensions but this is not supported in jinja.

tae898:
damn I messed this up with some wrong merge / revert. I'll make a new PR.

templates/docred/templates.yaml
```diff
@@ -0,0 +1,91 @@
...
+templates:
+  02af700c-e9e9-4a84-b75c-5fb29a5b7993: !Template
+    id: 02af700c-e9e9-4a84-b75c-5fb29a5b7993
+    jinja: "Read the following text and answer the questions.\n\nText:\n{% for sent\
+      \ in sents -%}\n{{ sent | join(\" \") }}\n{%- endfor %} \n\nQuestion:\nFrom\
+      \ the above text, find people, locations, organizations, time, numbers, and\
+      \ miscellaneous.\n|||\n{{\"People: \"}}\n{%- set people = [] -%} \n{%- for ners\
+      \ in vertexSet -%}\n{%- for ner in ners if ner['type'] == 'PER' -%}\n{{people.append(ner['name'])\
+      \ | default(\"\", True)}}\n{%- endfor -%}\n{%- endfor -%}\n{{ people | unique\
+      \ | join(\",\")}}{{\".\"}}\n\n{{\"Locations: \"}}\n{%- set locations = [] -%}\
```
craffel, line 11: I think we want join(", ") not join(",") for tokenization reasons.

craffel:
> damn I messed this up with some wrong merge / revert. I'll make a new PR.

NP, please assign me to the new PR and note the join suggestion above.

templates/docred/templates.yaml
tae898: hey sorry I made a new PR:

https://github.com/bigscience-workshop/promptsource/pull/268

PR #248, opened 2021-06-14
prompts for the conv_ai_2 datasets

josrozen:
5 prompts possibly matching the profile and utterances of the human in a conversation.

### DIFF
--- a/templates/conv_ai_2/templates.yaml
+++ b/templates/conv_ai_2/templates.yaml
@@ -2,67 +2,166 @@ dataset: conv_ai_2
 templates:
   04f7a3d8-c40f-45d1-b9ae-5bd23ff11628: !Template
     id: 04f7a3d8-c40f-45d1-b9ae-5bd23ff11628
-    jinja: "{% if (0 < (dialog | selectattr(\"sender_class\", \"equalto\", \"Bot\"\
-      ) | list | length))\n  and (0 < (dialog | selectattr(\"sender_class\", \"equalto\"\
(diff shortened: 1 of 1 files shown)


craffel:
Thanks for this. It looks like most of these templates can result in an empty string, e.g.
```
I wonder if somebody describing herself, or himself, using these statements:

    "i like watching war documentaries.",

    "i volunteer at a soup kitchen.",

    "cheeseburgers are my favorite food.",

    "i was poor growing up.",

    "i'm a retired gym teacher.",

could utter things like:

in a conversation... What's your guess: yes or no?
```
Can you debug? Take a look at more examples from the dataset or the "prompted dataset viewer" to see what I mean.

josrzn:
Indeed, @craffel, there are dialogs with no utterance from a participant. I added a surrounding test to exclude those, in all prompts. thanks.

Now I'm facing another issue as check_templates is failing. I thought it was a yaml issue coming from my setup but this is not it. I'm looking into that error...

josrzn:
weird, I had to move ```{% set class = ["Human", "Bot"] | random %}``` outside of the surrounding ```if``` for the tests to stop complaining about the variable ```class``` being "unrecognized". I hope this still skips incomplete examples with missing utterances.

josrozen:
fyi, from a similar issue I had on conv_ai, and according to @VictorSanh's answer https://github.com/bigscience-workshop/promptsource/pull/252#discussion_r652226296, moving the setting outside of the surrounding ```if``` is 1/ indeed mandatory and 2/works fine.

craffel:
Thanks!

PR #249, opened 2021-06-14
Prompt meta_woz 

zaidalyafeai:


templates/meta_woz/dialogues/templates.yaml
```diff
@@ -0,0 +1,111 @@
+dataset: meta_woz
+subset: dialogues
+templates:
+  4b64c6e9-0aa0-431f-85b5-8367daa4773a: !Template
+    id: 4b64c6e9-0aa0-431f-85b5-8367daa4773a
+    jinja: "What does this conversation between a chatbot and a client talk about\
+      \ ? \n{{turns | join(',  ')}}\n|||\n{{domain.replace('_', ' ') | lower }}"
```
craffel, line 7: The "domains" listed in this dataset are superintuitive to me. Can you provide the list of possible domains to the model in the prompt?

templates/meta_woz/dialogues/templates.yaml
```diff
@@ -0,0 +1,111 @@
...
+  98b0697f-3144-40b6-b6ae-5d57c3f08db3: !Template
+    id: 98b0697f-3144-40b6-b6ae-5d57c3f08db3
+    jinja: "Chatbot: {{turns[0]}}\nHuman: {{turns[1]}}\nChatbot: \n|||\n{{turns[2]}}"
+    name: predict_next_stmt_1
+    reference: ''
+    task_template: true
+  a02a980d-630f-4845-8b4d-b0eb072110e9: !Template
+    id: a02a980d-630f-4845-8b4d-b0eb072110e9
+    jinja: "{% if turns|length % 2 == 1 %}\n{% for i in range((turns|length) // 2)\
+      \ %}\nBot: {{turns[2*i]}}\nUser: {{turns[2*i+1]}}\n{% endfor %}\nBot: \n|||\n\
```
craffel, line 69: It looks like the lines are turning out like
```
Bot: Hello how may I help you? User: I need some event help
Bot: Certainly, help with an event. How can I assist you? User: Can you please get me some tickets
```
Can you debug so they appear on new lines?

templates/meta_woz/dialogues/templates.yaml
```diff
@@ -0,0 +1,111 @@
...
+    reference: ''
+    task_template: true
+  5e90f705-9d63-4917-acc9-3baabc6ee5e9: !Template
+    id: 5e90f705-9d63-4917-acc9-3baabc6ee5e9
+    jinja: "{% set rand_index= namespace(value=range(turns|length)|random()) %}\n\
+      {% for utterance in turns %}\n{% if loop.index0 == rand_index.value %}\n{{[\"\
+      Chatbot\", \"Human\"][loop.index0 % 2]}}: \n{% else %}\n{{[\"Chatbot\", \"Human\"\
+      ][loop.index0 % 2]}}: {{utterance}}\n{% endif %}\n{% endfor %}\n|||\n{{turns[rand_index.value]\
+      \ }}"
+    name: 'predict_missing '
```
craffel, line 41: It seems a little hard for the model to know you want it to fill in this blank. Can you put something like `(blank)` in place of the missing utterance, and then at the end of the prompt ask the model to fill in the blank?

templates/meta_woz/dialogues/templates.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+      {{["AI Assistant", "Client"][ count.value % 2]}}:
+
+      |||
+
+      {{turns[count.value] }}
+
+
+      '
+    name: predict_random
```
craffel, line 31: It seems like this can result in predicting a single utterance, e.g. the prompt is just "AI Assistant:". This seems too hard. Can you make sure there is at least one turn of dialog before the model has to predict anything?

craffel (changes requested):

templates/meta_woz/dialogues/templates.yaml
craffel: Sorry, "super counterintuitive"

templates/meta_woz/dialogues/templates.yaml
zaidalyafeai: My intuition is that the model can learn to predict some statements for initializing a dialog like for example : How may I help you ? , what do you want to ask for ? , etc.

templates/meta_woz/dialogues/templates.yaml
zaidalyafeai: I thought the same, however the list of domains is big, it seems 51 https://huggingface.co/datasets/meta_woz#data-splits

templates/meta_woz/dialogues/templates.yaml
craffel: I guess I'm just not sure this is a useful task - if, on the one hand, there are only a finite/small number of feasible generations, then you're basically just training the model to memorize the small set of things a helperbot would start a conversation with. On the other hand, if there are lots of possible "first utterances", then you're approaching unconditional generation, and we will have already trained the model a lot on unconditional generation.

templates/meta_woz/dialogues/templates.yaml
craffel: I see, yes, that seems like it is probably too big.

templates/meta_woz/dialogues/templates.yaml
zaidalyafeai: Ok, that makes sense!

zaidalyafeai:
@craffel thanks for the review, made a commit with the changes.

promptsource/promptsource.py
```diff
@@ -143,22 +143,13 @@ def show_text(t, width=WIDTH):
...
     #
     # Select a dataset - starts with ag_news
     #
     dataset_key = st.sidebar.selectbox(
-        "Dataset",
-        dataset_list,
-        key="dataset_select",
-        index=12,  # AG_NEWS
-        help="Select the dataset to work on.",
+        "Dataset", dataset_list, key="dataset_select", index=12, help="Select the dataset to work on.",  # AG_NEWS
```
craffel, line 152: Did these changes come from an automatic syntax formatter? It may be best to revert them if they don't match the style used for the codebase. It also looks like the #AG_NEWS comment got deassociated from its corresponding argument.

templates/meta_woz/dialogues/templates.yaml
craffel: I think it's probably better to join the turns with a newline - it's possible that ", " could appear in a turn, making it ambiguous.

craffel (changes requested):

promptsource/promptsource.py
zaidalyafeai: Sorry, it was by mistake, I reverted the changes.

templates/meta_woz/dialogues/templates.yaml
```diff
@@ -4,7 +4,8 @@ templates:
   4b64c6e9-0aa0-431f-85b5-8367daa4773a: !Template
     id: 4b64c6e9-0aa0-431f-85b5-8367daa4773a
     jinja: "What does this conversation between a chatbot and a client talk about\
-      \ ? \n{{turns | join(',  ')}}\n|||\n{{domain.replace('_', ' ') | lower }}"
+      \ ? \n{{turns |  join(\"\\n- \")}}\n|||\n{{domain.replace('_', ' ') | lower\
```
craffel, line 7: When you use join like this, the first line will not get a - before it, but all the other ones will. Can you just do the human/bot labels like you did in the other tasks?

craffel (changes requested):

craffel:
Thanks!

zaidalyafeai:
Thank @craffel and sorry that was messy.

PR #250, opened 2021-06-14
Templates for WIQA

arnaudstiegler:
This dataset (or at least the template tasks) is difficult to correctly phrase. It might make sense to discard the prompts related to the template tasks and only keep the side tasks (but is it worth keeping the dataset at all in this case?)

### DIFF
--- a/templates/wiqa/templates.yaml
+++ b/templates/wiqa/templates.yaml
@@ -1,8 +1,158 @@
 dataset: wiqa
 templates:
+  1bc8d95b-0a50-49f4-a46b-bd752929926d: !Template
+    id: 1bc8d95b-0a50-49f4-a46b-bd752929926d
+    jinja: '{% for k in range(1, question_para_step | length) %}
+
+      - {{question_para_step[k]}}
+
+      {% endfor %}
+
+
+      Find the first step of the process
+
+
+      |||
+
+
+      {{ question_para_step | first }}
+
+      '
+    name: remove_first_step
+    reference: ''
+    task_template: false
+  360cd99a-2f83-469a-a505-d80808159dd2: !Template
+    id: 360cd99a-2f83-469a-a505-d80808159dd2
+    jinja: '{% if question_para_step[-1] == "" %}
+
+      {% set process_list = question_para_step[:-2] %}
+
+      {% else %}
+
+      {% set process_list = question_para_step %}
+
+      {% endif %}
+
+
+      {% for k in range(process_list | length) %}
+
+      - {{question_para_step[k]}}
+
+      {% endfor %}
+
+
+      Find the last step of the process
+
+
+      |||
+
+
+      {{ process_list | last }}
+
+      '
+    name: remove_last_step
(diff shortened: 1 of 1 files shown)


arnaudstiegler:
Working on fixing the failing test (due to some local variable not correctly set)

templates/wiqa/templates.yaml
```diff
@@ -0,0 +1,193 @@
...
+      - {{ question_para_step | join("\n- ")}}
+
+
+      Question:
+
+      {{question_stem}}
+
+
+      How is the supposed perturbation influence the second effect mentioned. Answer
+      by {{"more, less or no effect"}}
```
VictorSanh, line 105:
```suggestion
      How does the supposed perturbation influence the second effect mentioned? Answer
      by {{"more, less or no effect"}}.
```

templates/wiqa/templates.yaml
```diff
@@ -0,0 +1,193 @@
...
+
+
+      Is the supposed perturbation:
+
+      - {{"directly impacting a step of the process"}}
+
+      - {{"indirectly impacting a step of the process"}}
+
+      - {{"not impacting any step of the process"}}
+
```
VictorSanh, line 133:
```suggestion
      - {{"not impacting any step of the process"}}
      
      ?
```

templates/wiqa/templates.yaml
```diff
@@ -0,0 +1,193 @@
+dataset: wiqa
+templates:
+  1bc8d95b-0a50-49f4-a46b-bd752929926d: !Template
+    id: 1bc8d95b-0a50-49f4-a46b-bd752929926d
+    jinja: '{% for k in range(1, question_para_step | length) %}
+
+      - {{question_para_step[k]}}
+
+      {% endfor %}
```
VictorSanh, line 9:
```suggestion
    jinja: '- \n- {{ question_para_step | join("\n- ") }}
```

VictorSanh: same for `remove_first_step_bis`

VictorSanh: and `remove_last_step` and `remove_last_step_bis`

VictorSanh (changes requested):
looks good to me!
few minor suggestions. please ping me once corrected, will merge then

templates/wiqa/templates.yaml
arnaudstiegler: Those were actually on purpose:
- for `remove_first_step` and `remove_first_step_bis`, the range starts at 1 to remove the first step. Not sure whether there's an easy way to using slicing with jinja
- for `remove_last_step` and `remove_last_step_bis`, the range stops at length-1 to remove the last step. Again not sure whether we can do the slicing directly with jinja

arnaudstiegler: Actually, my bad, slicing is working. Will update

VictorSanh:
thanks! merging now, please update the spreadsheet

PR #252, opened 2021-06-14
prompts for the conv_ai dataset

josrozen:
6 prompts around engagement in conversations given a context

### DIFF
--- a/templates/conv_ai/templates.yaml
+++ b/templates/conv_ai/templates.yaml
@@ -2,7 +2,8 @@ dataset: conv_ai
 templates:
   1664cdd9-54e8-4679-821b-8013e9df197e: !Template
     id: 1664cdd9-54e8-4679-821b-8013e9df197e
-    jinja: '{{context}}
+    jinja: '{% if 0 < (thread | selectattr("userId", "equalto", "Alice") | list | length) %}
+      {{context}}
 
 
       {% for utterance in thread %}
@@ -25,13 +26,16 @@ templates:
 
       {% endif %}
 
-      {% endfor %}'
+      {% endfor %}
+      
+      {% endif %}'
     name: engagement_alice_1
     reference: ''
     task_template: false
   2d3ca9f5-60f0-459d-932f-ab30e1e79b22: !Template
     id: 2d3ca9f5-60f0-459d-932f-ab30e1e79b22
-    jinja: '"{{context}}""
+    jinja: '{% if 0 < (thread | selectattr("userId", "equalto", "Bob") | list | length) %}
+      "{{context}}"
 
 
       Would you say Bob engagement is real, given the previous context, in this conversation:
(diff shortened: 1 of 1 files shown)


templates/conv_ai/templates.yaml
```diff
@@ -0,0 +1,152 @@
...
+      |||
+
+      {% set alice = (evaluation|selectattr("userId", "equalto", "Alice")|first)["engagement"]
+      %}
+
+      {% set bob = (evaluation|selectattr("userId", "equalto", "Bob")|first)["engagement"]
+      %}
+
+
+      {% if alice == bob %}both{% elif alice < bob %}Bob{% else %}Alice{% endif %}'
```
VictorSanh, line 82: this template should be conditional: it gives a very weird prompted example where the thread has only 1 or 2 utterances (or even 0)...
Capture d’écran 2021-06-15 à 5 12 25 PM
```suggestion
    jinja: '{% if thread | length >= 4%}
    {{context}}


      Given the previous context, who do you think is more engaged in this conversation
      (Alice, Bob, or both):

      {% for utterance in thread %}

      - {{ utterance["userId"] }}: {{ utterance["text"] }}

      {% endfor %}

      |||

      {% set alice = (evaluation|selectattr("userId", "equalto", "Alice")|first)["engagement"]
      %}

      {% set bob = (evaluation|selectattr("userId", "equalto", "Bob")|first)["engagement"]
      %}


      {% if alice == bob %}both{% elif alice < bob %}Bob{% else %}Alice{% endif %}
      {% endif %}'
```

VictorSanh (approved):
this is such a strange dataset
please verify my suggestion, otherwise lgtm!

templates/conv_ai/templates.yaml
josrozen: it does look weird. I should update all templates though. I'd suggest using something like ```{% if (0 < (thread | selectattr("userId", "equalto", "Bob") | list | length)) and (0 < (thread | selectattr("userId", "equalto", "Alice") | list | length)) %}```. Would that be ok?

VictorSanh: yeah your proposition looks good and should cover much of the cases I was worried about!
i think that would make sense to update all templates cause not sure whether it is useful who's the most engaged when someone is talking alone....

josrozen: there's still a (minor?) issue if I do this: the tests don't like the setting of alice and bob variables (```{% set alice = (evaluation|selectattr("userId", "equalto", "Alice")|first)["engagement"]```) inside of the all surrounding ```if``` and complain about "unrecognized" variables. This means I have to move the 2 ```set```at the top of the prompt, outside the ```if```. It works but will it correctly skip the incomplete examples we're trying to filter out?

josrozen:
thanks @VictorSanh, I've added the surrounding tests to exclude edge examples lacking some utterances from a participant.

templates/conv_ai/templates.yaml
VictorSanh: yes you'll have to do `set alice` and `set bob` outside of the if condition indeed!

The example will be filtered out as long as the template doesn't return anything (which is the case when you wrap the template inside an if condition independently of what variables you define outside the if condition)

VictorSanh (approved):
lgtm, thank you @josrozen !
please update the spreadsheet, merging now

PR #254, opened 2021-06-15
Craigslist bargains prompts

relaxedplan:
Added prompts covering the following tasks related to craigslist_bargains:
 - generating the next line of negotiation from a specified intention (main task)
 - determining if the buyer or the seller got the best deal
 - determining the gap between the buyer's target and the seller's target

### DIFF
--- a/promptsource/templates.py
+++ b/promptsource/templates.py
@@ -1,7 +1,7 @@
 import os
 import random
 import uuid
-from collections import defaultdict
+from collections import Counter, defaultdict
 from shutil import rmtree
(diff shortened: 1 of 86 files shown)


srush (changes requested):
It seems like the for loop here is failing on some examples in the test.

relaxedplan:
> It seems like the for loop here is failing on some examples in the test.

You're right, it was failing when the only items in "Price" were -1. Should be fixed now :)

srush:
Looks like it is still failing :)

relaxedplan:
seems like it passes now :)

PR #256, opened 2021-06-15
Add prompts for cos_e dataset

HanNight:
Add prompts for cos_e dataset

### DIFF
--- a/.github/workflows/check_templates.yml
+++ b/.github/workflows/check_templates.yml
@@ -18,8 +18,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
+          pip install .
           pip install -r requirements.txt
       - name: Check templates
         run: |
-          export PYTHONPATH=promptsource:$PYTHONPATH
           pytest test/test_templates.py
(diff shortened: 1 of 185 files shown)


stephenbach (changes requested):
Thank you! These are nice! Should any of them be marked as task templates?

Also, what do you think about the value of answering the question after providing the explanation. Looking through the examples, it seems like the answer can be found just by token matching after the explanation is given. Maybe not very worthwhile?

HanNight:
I also thought of this question when I was writing the prompts. I study the method of the paper of this dataset. The authors first generate explanations based on the questions and options, and then predict the answers based on the original questions, options, and the generated explanations. So based on this, I wrote some prompts with the explanation.  

In addition, the explanation of some examples includes all the options, and some examples only include the correct options.  

I do not have any good solutions for this now. Do you have any suggestions? Thanks!

stephenbach:
Oh interesting. I guess that's a different situation because they have to generate the explanations whereas, as currently written, we sometimes condition on the explanation. I would just remove the templates that condition on the explanation. Seems too easy.

Also, would you mark any of the ones that require answering the question as a task template? Let's take question answering to be the true task since it's their ultimate goal. Thanks!

HanNight:
Thanks for your suggestion. I removed the templates that condition on the explanation and marked the templates that require answering the question as task template.

stephenbach:
Great, thank you!

PR #258, opened 2021-06-15
Add E2E-NLG Dataset

blackrosedragon2:
I could not find a way to use Regex with ordinary jinja, so I have manually created loops to find elements inside brackets for some of the prompts. If there is a better way of doing this, do let me know !


srush:
Yeah, I think this is a bit complex. You shouldn't need to do any sort of parsing like this in jinja. can you explain why you need regex's?

This dataset is a text generation dataset. It is mainly designed to require the system to take in the table and generate the output. That should be the main focus of the prompts I believe.

blackrosedragon2:
> Yeah, I think this is a bit complex. You shouldn't need to do any sort of parsing like this in jinja. can you explain why you need regex's?
> 
> This dataset is a text generation dataset. It is mainly designed to require the system to take in the table and generate the output. That should be the main focus of the prompts I believe.

I wanted to add some prompts that took the human reference and ask some questions from it (like a q-a task). To do that I had to parse the "meaning_representation" string to extract the right key-value pairs. 
For example:
 `{
  "human_reference":"The Eagle is a low rated coffee shop near Burger King and the riverside that is family friendly and is less than £20 for Japanese food."
  "meaning_representation":"name[The Eagle], eatType[coffee shop], food[Japanese], priceRange[less than £20], customer rating[low], area[riverside], familyFriendly[yes], near[Burger King]"
}` If I wanted to ask, given the human reference what was the name of the dinery. To do this would require me to use regex or parse the string.

I've added the generation tasks to the latest commit.

PR #260, opened 2021-06-15
tmu-gfm-dataset

tae898:


srush:
Are the scales from 0 to 4 or 1 to 4 (seems to be inconsistent). 

Also can you confirm that the choice part works? I think I would rather you assign the choice to a variable that have two if statements.

srush (changes requested):
Changes

tae898:
Hi thanks for the feedback.


Are the scales from 0 to 4 or 1 to 4 (seems to be inconsistent).
> There are three metrics used in this dataset. They are `grammar`, `fluency`, and `meaning`. I checked the original dataset and all of the samples and I am very certain that `grammar` and `meaning` are on a scale from 0 to 4, whereas `fluency` is from 1 to 4. I know this is weird but it is what it is. So what I did is that when the prompt asks all of the three metrics, then it simply asks to use the scale from 0 to 4, but when I created prompts that specifically ask for one metric, then I specify if it's from 0 to 4 or 1 to 4.
 
Also can you confirm that the choice part works? I think I would rather you assign the choice to a variable that have two if statements.
> Yeah `{% if range(0,2) | choice %}` works good. It really gives you 50% chance. I think your suggestion is something like this?

```
{% set foo = range(0,2) | choice %}
{% if foo == 1 %}
{{"foo is True"}}
{% else %}
{{"foo is False"}}
{% endif %}
```

I'm pretty certain above snippet and what I did result in the same thing. But if you think this is better, I can change it.

Thanks.

srush:
Great thanks!

PR #261, opened 2021-06-15
Squad adversarial prompts

relaxedplan:
Added prompts for squad_adversarial. Prompts are based on SQUAD prompts. The key difference is that to account for human answers which aren't always the same, we select the most common human answer, or select randomly in case of a tie.


srush (changes requested):
* Do you think it is worth adding a function to the python code to run this loop? It seems like it is used in a ton of filters and is hard to read. 
* We aren't using random, but a new filter we added called `choice`

relaxedplan:
> * Do you think it is worth adding a function to the python code to run this loop? It seems like it is used in a ton of filters and is hard to read.

Couldn't agree more - I added `most_frequent` as a helper function

> * We aren't using random, but a new filter we added called `choice`

Sure, I've replaced `random` with `choice`

Arun Raja wrote some prompts for this dataset which were approved this morning, I've consolidated our prompts.

Finally, if the new function is fine, I'll edit the [squadshifts PR](https://github.com/bigscience-workshop/promptsource/pull/262) to use it.

srush (approved):

srush:
Oh this is soooo much better.

PR #262, opened 2021-06-15
Squadshifts prompts

relaxedplan:
Added prompts for squadshifts. Prompts are based on SQUAD prompts. The key difference is that to account for human answers which aren't always the same, we select the most common human answer, or select randomly in case of a tie.

I excluded the "reddit" subset of the dataset.


srush (changes requested):
Marking `request changes` to remind myself this is being updated.

relaxedplan:
I've tidied templates up using the most_frequent helper function, Should be good to merge.

PR #265, opened 2021-06-15
Added templates for fever subsets 1.0 and 2.0

TevenLeScao:
This PR adds templates for the fever dataset, treating it like closed-book QA (instead of the original task, fact-verification with a source)


stephenbach:
Thanks, these are great! Do you think there's a benefit to including the ones with "I don't know" answers? It seems like whether the answer is "I don't know" depends on the unseen document, as opposed to being a closed-book test of knowledge.

TevenLeScao:
Actually I was wondering about that too - I liked the idea of including parts where the model should say it doesn't know, but after looking, these are quite inconsistent (hard to separate from either of the other 2 classes without the attached context). Now, I'd be in favour of dropping them.

stephenbach:
Yup, I agree. So you could wrap the whole template in an if statement that checks that the answer isn't "I don't know."

TevenLeScao:
Done!

stephenbach:
Perfect, thanks!

PR #268, opened 2021-06-15
Fix wrong puncutations. Remove duplicates. Add two more templates.

tae898:
I made this PR again since I messed up the last one.

### DIFF
--- a/templates/docred/templates.yaml
+++ b/templates/docred/templates.yaml
@@ -71,8 +71,8 @@ templates:
       \ = [] -%}\n{% for ners in vertexSet %}\n{% for ner in ners if ner['name'] not\
       \ in names %}\n{{ names.append(ner['name']) | default(\"\", True) }} \n{{ types.append(ner['type'])\
       \ | default(\"\", True) }} \n{% endfor %}\n{% endfor %}\n{% for name, type in\
-      \ zip(names, types) %}\n{{name}}{{\": \"}}{{type}}\n{% endfor %}\n|||\nText:\n\
-      {% for sent in sents -%}\n{{ sent | join(\" \") }}\n{%- endfor -%} "
+      \ zip(names, types) %}\n{{name}}{{\": \"}}{{type}}\n{% endfor %}\n|||\n{% for\
+      \ sent in sents -%}\n{{ sent | join(\" \") }}\n{%- endfor -%} "
     name: ner-to-text
     reference: ''
     task_template: false
--- a/templates/docred/templates.yaml
+++ b/templates/docred/templates.yaml
@@ -1,0 +1,91 @@
+dataset: docred
+templates:
+  02af700c-e9e9-4a84-b75c-5fb29a5b7993: !Template
+    id: 02af700c-e9e9-4a84-b75c-5fb29a5b7993
+    jinja: "Read the following text and answer the questions.\n\nText:\n{% for sent\
+      \ in sents -%}\n{{ sent | join(\" \") }}\n{%- endfor %} \n\nQuestion:\nFrom\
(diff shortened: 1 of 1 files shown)


tae898:
@craffel

Here are the points that we've been discussing.

1. Thanks for this. We are shooting for at least 5 prompts per dataset. Can you add some more? One option would be to choose a random single entity and ask its type, or choose a random pair and ask the relation, etc.
    > yeah I added two more templates where the output is generated from ners and relations, respectively.
2. This ends up being a bit weird because entities are duplicated in the list. Can you deduplicate?
    > yeah I fixed them. I couldn't use a simple `unique` filter since I was dealing with a list of lists of dictionaries. The code might look messy but this was the best I could do. Of course this could have been dealt with easily in python using list comprehensions but this is not supported in jinja.
3. I think we want join(", ") not join(",") for tokenization reasons.
    > You have very keen eyes! Actually this is the only case where I used `","` instead of `", "`. The reason for this is that this text contains things such as ` - `, which is rendered as bullet points and make the whole thing messy. I thought that not having a white space after the commas is better than having those bullet points. If you think otherwise, I can change it.

Thanks.

templates/docred/templates.yaml
```diff
@@ -0,0 +1,91 @@
...
+templates:
+  02af700c-e9e9-4a84-b75c-5fb29a5b7993: !Template
+    id: 02af700c-e9e9-4a84-b75c-5fb29a5b7993
+    jinja: "Read the following text and answer the questions.\n\nText:\n{% for sent\
+      \ in sents -%}\n{{ sent | join(\" \") }}\n{%- endfor %} \n\nQuestion:\nFrom\
+      \ the above text, find people, locations, organizations, time, numbers, and\
+      \ miscellaneous.\n|||\n{{\"People: \"}}\n{%- set people = [] -%} \n{%- for ners\
+      \ in vertexSet -%}\n{%- for ner in ners if ner['type'] == 'PER' -%}\n{{people.append(ner['name'])\
+      \ | default(\"\", True)}}\n{%- endfor -%}\n{%- endfor -%}\n{{ people | unique\
+      \ | join(\",\")}}{{\".\"}}\n\n{{\"Locations: \"}}\n{%- set locations = [] -%}\
```
craffel, line 11: Please remove these periods. It can result in stuff like
"People: ."
when the list is empty.

craffel (changes requested):

templates/docred/templates.yaml
```diff
@@ -0,0 +1,91 @@
...
+    reference: ''
+    task_template: true
+  6efa4d1a-3368-4b12-9e30-588b53801077: !Template
+    id: 6efa4d1a-3368-4b12-9e30-588b53801077
+    jinja: "{% if labels['relation_text'] %}\nGiven the following entities and relations,\
+      \ make a creative text. The types are PER (Person), LOC (Location), ORG (Organization),\
+      \ TIME (Time), NUM (Number), and MISC (Miscellaneous).\n\n{% for  head, tail,\
+      \ relation in zip(labels['head'], labels['tail'], labels['relation_text']) %}\n\
+      head: {{vertexSet[head][0]['name']}}, tail: {{vertexSet[tail][0]['name']}},\
+      \ relation: {{relation}}\n{% endfor %}\n|||\nText:\n{% for sent in sents -%}\n\
```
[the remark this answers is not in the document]
tae898: alright I removed it.

templates/docred/templates.yaml
tae898: I added a `if` statement so that the recognized entities are only displayed when the list is not empty. The period `.` is still there.

PR #272, opened 2021-06-16
Corrections for the stsb_multi_mt_en templates

NohTow:
Following discussions on slack, I corrected templates where I translated task to a binary classification so they are not "dataset task" anymore.
Also changed the name of those for more clarity.


stephenbach:
Great, thanks!

PR #274, opened 2021-06-16
adding prompts for freebase_qa

rteehas:
Added prompt templates

### DIFF
--- a/templates/freebase_qa/templates.yaml
+++ b/templates/freebase_qa/templates.yaml
@@ -1,0 +1,49 @@
+dataset: freebase_qa
+templates:
+  02b12d5c-a481-494d-84ea-a210eefa66d2: !Template
+    id: 02b12d5c-a481-494d-84ea-a210eefa66d2
+    jinja: "{% set question_context = Parses.TopicEntityName | choice %}\n{% set inference_context\
+      \ = Parses.InferentialChain | choice %}\nThe topic of this question is: {{question_context.split(\"\
+      \ \") | map(\"capitalize\") | join(\" \")}}.\n\nTo solve this question, use\
+      \ the following chain of relations between the topic and answer: {{inference_context}}\n\
+      \nUsing this, answer the following question:\n\n{{RawQuestion}}\n||| \n{% set\
+      \ answer = Parses.Answers | choice %}\n{{answer.AnswersName[0][0].split(\" \"\
+      ) | map(\"capitalize\") | join(\" \") }}"
+    name: qa_context_2
+    reference: qa prompt with topic and inference chain provided
+    task_template: false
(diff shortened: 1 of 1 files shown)


stephenbach:
Thanks!

I'm wondering about these inference chain prompts. The output might be easier for the model to generate and less sensitive to vocab if we made the output more natural.
```
Prompt + X

What chain of relations between topic entity and answer could be used to solve the following question:

The final of which sport at the 1900 Olympics in Paris attracted 6,000 spectators, the largest crowd at any event at that Olympics?

Y

olympics, olympic games, sports
```

rteehas:
> Thanks!
> 
> I'm wondering about these inference chain prompts. The output might be easier for the model to generate and less sensitive to vocab if we made the output more natural.
> 
> ```
> Prompt + X
> 
> What chain of relations between topic entity and answer could be used to solve the following question:
> 
> The final of which sport at the 1900 Olympics in Paris attracted 6,000 spectators, the largest crowd at any event at that Olympics?
> 
> Y
> 
> olympics, olympic games, sports
> ```

That makes sense to me. Originally, I kept the "." notation because I thought a more natural output might lose the structural/compositional aspect of the inference chain, but in hindsight that seems like a less relevant concern. Do you think   something like `olympics, olympic games, sports` is preferable to `olympics to olympic games to sports`?

stephenbach:
Great! Nope, no preference between the two. Up to you!

rteehas:
I ended up sticking with the ", " format. I thought "to" would imply causality or direction when it's not strictly true that that's the case

srush:
These chain of inference prompts seem basically impossible to know without the dataset. I think we should just remove them?

rteehas:
> These chain of inference prompts seem basically impossible to know without the dataset. I think we should just remove them?

I mainly included them based on a comment in the Freebase paper about training to search through the database. I thought that that could be a proxy for learning a set of inference steps. One thing I could do that might make more sense, if you think that's less useful, is exclude everything but the first and last link in the chain and rephrase the prompt to ask what kinds of objects the question is connecting (i.e. what kind of object is the question about and what kind of object do you expect the answer to be).

stephenbach:
Yeah, agree with @srush here. Was on the fence, but not clear that there’s a generalizable skill here. Fine to just cut them and have the QA templates.

rteehas:
Ok, in that case I'll add some more QA style templates to make up for the removed inference chain ones

rteehas:
Do you think it wouldn't be useful to have a task where the goal is to identify the category the answer is in?

stephenbach:
Yup, I think that makes more sense.

stephenbach:
Thank you!

PR #278, opened 2021-06-18
Nq-Open Correction

sbmaruf:
Adapted to some correction you suggested in "trivia-qa".

I think for open-domain QA, we can at least mention the KB in a question like below, 

1. Can you tell me the answer of the following question `from wikipedia`?
2. According to `wikipedia`, {{question}}

@VictorSanh


VictorSanh:
It says the are conflicts on `templates/nq_open/templates.yaml`, can you solve the conflict so that I can merge?

> I think for open-domain QA, we can at least mention the KB in a question like below,
> 
> 1. Can you tell me the answer of the following question `from wikipedia`?
> 2. According to `wikipedia`, {{question}}

After much thought, I think the task description should be self-contained. Adding references to things that are not in the prompt is more like noise that I am not sure we need at this step

sbmaruf:
@VictorSanh Removed those prompts. Now it is saying `This branch has no conflicts with the base branch`. Please let me know if I need to do anything more.

PR #279, opened 2021-06-18
Updating task template checkbox for art

elsanns:
I only marked templates that provide the model with both choices. Left unmarked those with one choice and a yes/no option. Right or too strict?


srush:
That's perfect. Thanks.

PR #281, opened 2021-06-18
Make templates package data

craffel:
This moves the templates folder into the source tree and installs it along with the package. That way, you can access the  templates without being in the base directory of this git repository. This is necessary for flexibly running code that makes use of the templates from locations other than the repo base directory. If you are in the base directory and you run streamlit (as people are supposed to do), it still edits the template in the repository (not the installed one in whatever Python package directory it was installed in).

Please test this out before merging.


VictorSanh (approved):
looks good to me!

stephenbach (approved):
Works beautifully!

I complicated things though by merging in a couple of template PRs before I noticed this, so just a reminder to move those two datasets as well.

craffel:
Moved the new templates. Anyone know what's up with the failing check? https://github.com/bigscience-workshop/promptsource/runs/2862977471

stephenbach:
Ah, it's because the change in path to templates breaks the test's parsing of paths. It seems I can't make the changes directly because it's not a modified file in this PR, but the changes that are needed are:
```
diff --git a/test/show_templates.py b/test/show_templates.py
index 0c44055..bd22350 100644
--- a/test/show_templates.py
+++ b/test/show_templates.py
@@ -14,8 +14,8 @@ if "templates.yaml" not in args.dataset_path:
 
 path = args.dataset_path.split("/")
 
-dataset_name = path[1]
-subset_name = path[2] if len(path) == 4 else ""
+dataset_name = path[2]
+subset_name = path[3] if len(path) == 5 else ""
 template_collection = TemplateCollection()
 
 dataset, _ = get_dataset(dataset_name, subset_name)
```

stephenbach:
I should add, this PR will still probably fail after this fix because it's going to try to download ~150 datasets, but it will at least work for future changes.

craffel:
Thanks. I'm going to assume that the remaining breaking test is due to the big download and will squash and merge.

PR #282, opened 2021-06-19
True task flags for S/M/X/ANLI, SciTail, HANS

awebson:
Checked the `task template` checkboxes for the NLI prompts I wrote prior to the introduction of this flag. 

As noted in #59, contrapositive prompts like "does the premise contradict the hypothesis" while naively negating the annotation label sometimes lead to weird pragmatic effects, especially when the text itself contains negation words. So to err on the cautious side I only marked the premise-entails-hypothesis prompts with the true task flag.


stephenbach:
Thanks Albert! We just merged in #281, which moves where the templates directory lives. Would you mind pulling main and updating the template files with the task template flags in their new locations? Thank you!

awebson:
Done!

stephenbach:
Thank you!

PR #286, opened 2021-06-21
template for emotion dataset

sIncerass:


promptsource/templates/emotion/templates.yaml
```diff
@@ -0,0 +1,37 @@
+dataset: emotion
+templates:
+  2da087fe-8cca-4f92-b19c-babccb26a510: !Template
+    id: 2da087fe-8cca-4f92-b19c-babccb26a510
+    jinja: '{{text}} The emotion expressed for the message is ||| {{ ["sadness", "joy",
```
srush, line 5: Can you add one where it gives the 6 options in the input? That seems like the real Task Template.

srush (changes requested):

promptsource/templates/emotion/templates.yaml
sIncerass: Sure, just added.

srush (approved):

PR #289, opened 2021-06-22
templates for sem_eval_2010_task_8

sIncerass:


sIncerass:
It seems to be a connection issue, it passed locally on my machine.

PR #291, opened 2021-06-22
Updating task template checkbox for cc_news

elsanns:


srush:
I am not sure I understand how these can all be task_templates if they ask for different outputs?

elsanns:
It's not all that clear, thanks. 
I decided to mark those asking for a shorter form of the "text" field, be it description or title, and left out templates having "text" as a target.
I am not sure whether the primary purpose of this dataset is generating a description or a title (kind of: summarization vs topic modelling?).

srush:
Ah good point. Yeah I think this is just a text collection, not really a task.

PR #299, opened 2021-06-23
Skip empty answers in private test sets

awebson:
Per [discussion](https://huggingface.slack.com/archives/C020GCD4YTV/p1624470853227200?thread_ts=1624382646.187700&cid=C020GCD4YTV) in Slack, this adds an if condition that skips the private test set of SuperGLUE ReCoRD, which has empty `answers` that cause seqio_cache_tasks to fail.


craffel:
Thanks!

PR #301, opened 2021-06-23
Avoid empty choice for evidence infer

craffel:


srush:
Looks like the test is failing for a real reason. "  ValueError: Template for dataset evidence_infer_treatment/1.1 with uuid c999469a-20e0-4c10-a707-3c057d5c0245 has unrecognized variable sub_sub_annotation."

craffel:
> Looks like the test is failing for a real reason. " ValueError: Template for dataset evidence_infer_treatment/1.1 with uuid c999469a-20e0-4c10-a707-3c057d5c0245 has unrecognized variable sub_sub_annotation."

It's a real reason but I don't think it's true - `sub_sub_annotation` is defined in template `c999469a-20e0-4c10-a707-3c057d5c0245`: `{% set sub_sub_annotation = range(0, sub_annotation_length) | choice %}`. I looked at this for a minute earlier and couldn't figure out why it's saying it's unrecognized. It's only defined if a conditional is True but the same is true for the other templates.

PR #305, opened 2021-06-24
Handling empty subannotations with choice filter

elsanns:
This PR excludes samples where an error would be thrown because of applying the choice filter to an empty sub-annotation list.

- If an empty sub annotation gets selected, the X+Prompt becomes empty (and sometimes Y) -> the sample will be omitted
- it will randomly produce empty or valid (x, y) pairs for samples with empty and non-empty sub annotations (based on choice filter). So the same sample may or may not be excluded depending on 'choice'. I can't see why it could be a  problem but alternatively, we could exclude samples with at least one empty sub annotation.


elsanns:
Seems to be related to variable visibility, I will prepare another commit

stephenbach:
Great! I don't see an easy way to filter the annotations before applying choice, so I think it's ok as is. Thanks!

PR #308, opened 2021-06-24
Debug after jz

lintangsutawika:
Test sets have empty lists that caused the error.


VictorSanh:
thanks! merging now, will report back on slack

PR #310, opened 2021-06-24
fixed STSB type errors

awebson:
Jinja seems to require parentheses around arithmetic expressions and also no whitespace around operators.


stephenbach:
Thanks!

PR #320, opened 2021-07-15
Annotated clean prompts

awebson:
I will add documentations about each column of the annotated prompt CSV later…


awebson:
Hmm it failed check_code_quality even after I have run `black`. Any ideas?

PR #321, opened 2021-07-16
new-line removed from prompt

sbmaruf:
Earlier before submitting the prompt, I tested `question_answer` prompt with open-ai `da-vinci` api and it worked perfectly with a "\n"

However, according to the recent suggestion of @awebson  I removed the `\n` from the prompts. Hope it helps.


sbmaruf:
Please note that t5 like tokenizer ignores the "\n". An example, 

`tok.decode(tok.encode("Answer the following question. \nwhere did they film hot tub time machine"))`

Output

`Answer the following question. where did they film hot tub time machine</s>`

However @awebson insisted removing newlines or adding space between them.

I am still not full convinced on the requirement of this. But just added this pull in case it helps.

PR #322, opened 2021-07-23
Load proper metrics for do_eval tasks

awebson:
@craffel This is a first stab at loading the appropriate metric for each dataset_subset_template.

Caveats:
1. `t5.evaluation.metrics` seems to do something special with SuperGLUE WSC. So I haven't handled that yet.
2. No seqio pre/postprocessing is handled, which is most relevant for…
3. The `rank_classification` metric, which I assume implements the trick of of plugging in the answers and then compare LM probabilities as class predictions.
4. For the caching error of leading space in ground truth targets, stripping of targets is not yet handled by metrics. We talked about using functools.partial but that can only fix an argument to a constant, not applying strip() to an argument. We can write a lambda wrapping every metric, but maybe postprocessing is still the way to go.


craffel:
The postprocess_fn does get applied to the target: https://github.com/google/seqio/blob/main/seqio/evaluation.py#L209 so, for all tasks, I would suggest adding the equivalent of strip() as a postprocessor.
https://github.com/google/seqio#postprocessor

promptsource/seqio_tasks/tasks.py
```diff
@@ -37,6 +37,18 @@
...
+
+def strip_whitespace(output_or_target, example=None, is_target=False):
+    """
+    Temporary hack. Cached tasks from promptsource all have a leading space
+    on the ground-truth targets.
+    """
+    if is_target:
+        return output_or_target.strip()
+    else:
+        return output_or_target
```
craffel, line 49: I think it's probably fair/correct to strip() both inputs and targets.. WDYT?

promptsource/seqio_tasks/load_annotated_prompts.py
```diff
@@ -0,0 +1,109 @@
...
+            # Required for all prompts with non_trivial_choices_hidden,
+            # but could be used for other tasks as well where answer choices are given.
+            if "metrics" not in task:
+                task["metrics"] = [rank_classification]
+            elif rank_classification not in task["metrics"]:
+                task["metrics"].append(rank_classification)
+
+        # should be already handled by NON_GLUE_METRICS
+        # if task['generative_true_task'] or task['generative_non_true_task']:
+        #     task['metrics'] = rogue
```
craffel, line 90: I know this is a comment but it's rouge not rogue

promptsource/seqio_tasks/tasks.py
```diff
@@ -32,6 +37,18 @@
     ("tweet_eval", "irony"),
 ]
 
+
+def strip_whitespace(output_or_target, example=None, is_target=False):
+    """
+    Temporary hack. Cached tasks from promptsource all have a leading space
```
craffel, line 43: TBH I think it's probably not temporary...

promptsource/seqio_tasks/load_annotated_prompts.py
```diff
@@ -0,0 +1,109 @@
+import csv
```
craffel, line 1: Sort of a nitpick (so feel free to disregard) but I would include these functions in tasks.py and then just treat this as a diagnostic script.

craffel (changes requested):

promptsource/seqio_tasks/load_annotated_prompts.py
awebson: Sounds good. I will move them in the next commit.

promptsource/seqio_tasks/load_annotated_prompts.py
awebson: Good catch!

promptsource/seqio_tasks/tasks.py
awebson: Sure. We can double check later but can definitely strip both to be safe for now.

awebson (commented):
Pushed new commits that strip both output and target in postprocessor and fixed typos.

craffel:
Thanks!

PR #324, opened 2021-08-03
Switch to indexing format where missing

craffel:
Some prompts had their targets in the format `{% if label == "entailment" %} Yes` etc. It's harder to figure out what the label strings are in this case. This switches them all to indexing notation, like `{{ ["Yes", "Maybe", "No"][label]`. Please scrutinize this carefully to think about whether it changes any of the tasks by mistake!

### DIFF
--- a/promptsource/templates/xnli/en/templates.yaml
+++ b/promptsource/templates/xnli/en/templates.yaml
@@ -3,9 +3,14 @@ subset: en
 templates:
   4e122d26-7e79-49c0-961b-cf8ee134759e: !Template
     id: 4e122d26-7e79-49c0-961b-cf8ee134759e
-    jinja: "Sentence 1: {{premise}}\nSentence 2: {{hypothesis}}\nQuestion: Does Sentence\
-      \ 1 contradict Sentence 2? Yes, No, or {{\"Neutral\"}}? |||\n{% if label ==\
-      \ 0 %} \nNo\n{% elif label == 1 %}\nNeutral\n{% else %}\nYes\n{% endif %}"
+    jinja: 'Sentence 1: {{premise}}
+
+      Sentence 2: {{hypothesis}}
+
+      Question: Does Sentence 1 contradict Sentence 2? Yes, No, or {{"Neutral"}}?
+      |||
+
+      {{["No", "Neutral", "Yes"][label]}}'
     name: Concatenation contraposition
     reference: Concatenation contraposition
     task_template: false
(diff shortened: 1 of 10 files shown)


craffel:
cc @awebson @NohTow

awebson:
Looks good to me. cc @sbmaruf who wrote the ones for WiC.

NohTow:
Agree that it makes the prompt easier to understand. 

I saw that you for `esnli` you did not change `explain_4` and `explain_2` (id: `f64d6196-370b-4501-acb5-e11a5ebf0c5e` for example). I assume it is because the conditioned string is more complex than just a label but it should still be more readable as something like 
`["\nWhy is it always the case that:\n", "\nWhy it is not necessarily the case that:\n ", "\nWhy is it not true that:\n"][label]` 
wouldn't it ? Or am I missing something ?

promptsource/templates/xnli/en/templates.yaml
```diff
@@ -42,16 +51,19 @@ templates:
...
-    jinja: "Sentence 1: {{premise}}\nSentence 2: {{hypothesis}}\nQuestion: Does Sentence\
-      \ 1 entail Sentence 2? Yes or No? |||\n{% if label == 0 %} \nYes\n{% else %}\n\
-      No\n{% endif %}"
+    jinja: 'Sentence 1: {{premise}}
+
+      Sentence 2: {{hypothesis}}
+
+      Question: Does Sentence 1 entail Sentence 2? Yes or No? |||
+
+      {{["Yes", "No", "No"][label]}}'
```
sbmaruf, line 66: `["Yes", "No", "No"]`, "No" twice?

promptsource/templates/xnli/en/templates.yaml
```diff
@@ -22,9 +27,13 @@ templates:
...
-    jinja: "Sentence 1: {{premise}}\nSentence 2: {{hypothesis}}\nQuestion: Does Sentence\
-      \ 1 contradict Sentence 2? Yes or No? |||\n{% if label == 2 %} \nYes\n{% else\
-      \ %}\nNo\n{% endif %}"
+    jinja: 'Sentence 1: {{premise}}
+
+      Sentence 2: {{hypothesis}}
+
+      Question: Does Sentence 1 contradict Sentence 2? Yes or No? |||
+
+      {{["No", "No", "Yes"][label]}}'
```
sbmaruf, line 36: `["Yes", "No", "No"]`, "No" twice?

sbmaruf (commented):
Some of the labels contain "No" twice in the label space.

promptsource/templates/xnli/en/templates.yaml
craffel: This is intentional, the original conditional mapped 0 -> yes, 1 or 2 -> no.

VictorSanh (approved):

craffel:
> Agree that it makes the prompt easier to understand.
> 
> I saw that for `esnli` you did not change `explain_4` and `explain_2` (id: `f64d6196-370b-4501-acb5-e11a5ebf0c5e` for example). I assume it is because the conditioned string is more complex than just a label but it should still be more readable as something like
> `["\nWhy is it always the case that:\n", "\nWhy it is not necessarily the case that:\n ", "\nWhy is it not true that:\n"][label]`
> wouldn't it ? Or am I missing something ?

I was only focusing on the target side - the main goal of this PR was to make it easier to determine what all of the possible target strings are, so conditionals on the input side don't bother me. I can make this change if you prefer, though.

NohTow:
> > Agree that it makes the prompt easier to understand.
> > I saw that for `esnli` you did not change `explain_4` and `explain_2` (id: `f64d6196-370b-4501-acb5-e11a5ebf0c5e` for example). I assume it is because the conditioned string is more complex than just a label but it should still be more readable as something like
> > `["\nWhy is it always the case that:\n", "\nWhy it is not necessarily the case that:\n ", "\nWhy is it not true that:\n"][label]`
> > wouldn't it ? Or am I missing something ?
> 
> I was only focusing on the target side - the main goal of this PR was to make it easier to determine what all of the possible target strings are, so conditionals on the input side don't bother me. I can make this change if you prefer, though.

Oh ok sorry, it makes sense now.
I guess these changes are less useful than ones on the target, so it does not matter that much. Maybe it will help manually querying the model but I think it is ok to let it unchanged for now.

PR #330, opened 2021-08-12
fixed CommitmentBank label indexing bug

awebson:
Upon some error analyses I found a bug in my prompts for CommitmentBank: Its label 1 is contradiction, when usually label 2 is contradiction and label 1 is neutral among 3-way NLI sets. Sorry! Can't believe I didn't catch this earlier. That explains why our performance on CB was so bad even though we were good on other NLI sets.

@VictorSanh @craffel You don't have to rerun any eval now, but please re-cache seqio before you run the next series of evaluations. (And maybe pull #321 as well to fix a small bug in Natural Questions.)

### DIFF
--- a/promptsource/templates/super_glue/cb/templates.yaml
+++ b/promptsource/templates/super_glue/cb/templates.yaml
@@ -9,21 +9,21 @@ templates:
 
       Question: Does Sentence 1 contradict Sentence 2? Yes, No, or Neutral? |||
 
-      {{["No", "Neutral", "Yes"][label]}}'
+      {{["No", "Yes", "Neutral"][label]}}'
     name: does S1 contradict S2?
     reference: Copied from Victor's prompts for XNLI.
     task_template: false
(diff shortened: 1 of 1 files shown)


VictorSanh:
Oooh thanks for fixing that @awebson , lgtm!

Happy to relaunch the caching for just this one and upload it to v0.1 version of the data or wait a lil' and recache everything into a v0.2 after a pass of cleaning (though I'll be off for the next two weeks for the 2nd)

awebson:
Thanks! Yeah waiting is totally fine. CB measures a very specific linguistic phenomenon (factive vs. non-factive clause embedding) so there is no urgency. We will probably need a couple weeks to fix the other metrics in SuperGLUE anyway. Also, for v0.2, I will likely write more prompts for eval sets which we couldn't include for dumb reasons like no one wrote (clean) prompts for them or there was a dataset bug (Winogrande, Blimp, DROP and LAMBADA come to mind). Adding them would be nice held-out evals since we couldn't include them in training.

PR #333, opened 2021-08-23
Cleaning

stephenbach:
This PR adds functionality for a cleaning pass of the templates, including explicitly listing answer choices when applicable.


promptsource/promptsource.py
```diff
@@ -477,12 +485,16 @@ def get_infos(d_name):
                             elif updated_template_name == "":
                                 st.error("Need to provide a template name.")
                             else:
+                                # Parses state.answer_choices
+                                updated_answer_choices = [x.strip() for x in state.answer_choices.split(",")]
```
craffel, line 489: Not sure if this matters but this means that none of the answer choices can have commas in them. I could imagine targets of the form `["No, it's not", "Yes, it is"]`. But, if this is a pain, we can just ignore.

stephenbach: Good point. A few templates with automatically extracted answer choices already have commas, so we should pick something else. We could just use `|||` again.

craffel: We could probably use tabs, but I think tabs are hard to enter in browser interfaces. Maybe newlines? Or yeah, ||| is fine, I'm just worried it will be confusing to use it in two different contexts.

stephenbach: Yeah, I think either of those are clunky if we want a single-line input. Pushed ||| for now, but very easy to change.

promptsource/templates.py
```diff
@@ -46,6 +46,179 @@ def most_frequent(items):
...
+            :param metric: List of strings denoting metrics to use for evaluation
+            :param choices_in_prompt: If True, the answer choices are included in the templates such that models
+                see those choices in the input. Only applicable to classification tasks.
+            :param _do_train: internal parameter. If True, this template should be
+                used for training
+            :param _do_train: internal parameter. If True, this template should be
+                used for evaluation
+            """
+            self.task_format = task_format
+            self.metric = metric
```
VictorSanh, line 215: are we updating this anywhere in the cleaning pass?

VictorSanh (approved):
looks good to me! thank you @stephenbach 

should we also include new tests to make sure `task_template` and `choices_in_prompt` (and any other field) are filled out for the yamls in the pr?

promptsource/templates.py
stephenbach: Currently, no. @awebson and I were thinking it would be a placeholder that we could update as we do more evaluations. We'd need to establish some format for recording the metric(s). Also might require more paper reading to know for sure, so would be slower annotation. Could include if people would find it helpful.

stephenbach:
Thanks!

I was thinking we would just check the metadata manually in PRs. We'd have to copy the "find the changed templates" from the show_new_templates action, since the check_templates action checks all templates, even ones that haven't been updated. Nice to have if somebody gets to it, but I didn't want to block getting started. Let me know if anyone has any quick fixes.

PR #335, opened 2021-09-02
Clean AG News.

stephenbach:


VictorSanh (approved):
LGTM!

I would tend to think that all of these correspond to the original task though... no?
Like imagine we were to apply some data augmentation on the template, the augmented version would still be considered the original task I think (cf "politics" -> "politician")

stephenbach:
Ah yes, I agree. Changed to all original tasks.

PR #341, opened 2021-09-03
Clean Quartz

VictorSanh:
Few points of attention:
- I put `Comment: Answer_choices_key: choices["text"]` as I couldn't select an answer_choices_key beyond depth 1
- Metrics: I put `accuracy` and `sequence_accuracy` as my understanding is that these two are the same


stephenbach (approved):
LGTM!

Noting from our slack conversation: after cleaning we will combine the two accuracy types into one annotation.

stephenbach:
Merging now.

awebson:
Sorry if I missed anything, but what's differences between the "reverse" prompts here and the non-reversed ones? 

Also flagging that need reopen this and add answer_choices_keys after #373 is merged.

VictorSanh:
> Sorry if I missed anything, but what's differences between the "reverse" prompts here and the non-reversed ones?

The order of the choices is swapped in the "reversed" ones. That felt a little weird to me (it might creates some duplicates entries), so happy to remove them if you think so

awebson:
Yeah let's remove those when you add the answer_choices_keys. Having lots of near duplicates artificially decreases the variance of prompt performance.

PR #343, opened 2021-09-03
Display all templates info in prompted dataset viewer mode

VictorSanh:
Displaying all new template metadata info in the viewer mode too

Capture d’écran 2021-09-03 à 6 37 05 PM


stephenbach (approved):
LGTM!

PR #348, opened 2021-09-05
Clean copa.

stephenbach:


craffel (commented):
Shouldn't this have `answer_choices: [choice1, choice2]` on all the templates?

craffel (commented):
Some of the prompts here will discard part of the dataset (for example, all of the examples where question is not "cause" could be discarded if the prompt wraps things in a conditional like `{% if question == "effect" %}`. I think this means they aren't an "original task" prompt. We could probably merge the complementary prompts together so that all examples pass through. WDYT?

stephenbach:
I think we might have consensus in the slack thread that original_task should remain about skills and not differentiate between such variants?

Either way, pushed some changes that updates the `answer_choices_key` format if you wouldn't mind double checking that too.

craffel:
My concern is not whether it's the original task or not, but that if we eval on these prompts we will not be getting numbers over the full dataset, which I think we should avoid. So I think we should probably avoid prompts that avoid a conditional that discards examples, if it's a task we're going to eval on.

craffel (approved):

PR #350, opened 2021-09-05
Clean record.

stephenbach:


stephenbach:
This is interesting situation. This QA dataset is usually evaluated with SQUAD Span metrics, so I marked that. The choices are also listed, so I set `answer_choices_key`. But we shouldn't evaluate with rank classification because there's more than one right answer, so I didn't mark accuracy. cc @craffel @awebson to be aware, and in case this affects evaluation code in some way I'm missing.

awebson:
In principle rank eval could work with multiple correct choices as well, but yeah that would involve even more annoying engineering for very few datasets. Also note that seqio already built in metrics for all *GLUE tasks so I just used `t5.data.glue_utils.get_super_glue_metric`.

stephenbach:
Ah thanks, makes sense. I'm just cautious about making sure we don't accidentally do some evaluation that doesn't make sense.

craffel:
Yeah, I think ReCoRD is complicated enough that it would probably just be simplest to rely on the test server for final correct evaluation.

craffel (approved):

PR #358, opened 2021-09-07
fix: app_review's answer choices with some suggestion/discussion

tianjianjiang:
1. `*` is changed to ★ for streamlit compatibility;
2. `choices_in_prompt` may be a bit ambiguous;
3. Review-rating are positively correlated but imperfect; e.g., a positive review can be either a 5 or a 4;
4. Based on (3), perhaps the metrics can be some rank correlation instead of accuracy?
5. Kinda like to know what will happen if reversing the hypothesis and the premise. Not exactly long distance but suspicious to me;
6. Or even more nitpicking: wouldn't those rating be interpreted as the quality of the review, not the app?

### DIFF
--- a/promptsource/templates/app_reviews/templates.yaml
+++ b/promptsource/templates/app_reviews/templates.yaml
@@ -1,42 +1,30 @@
 dataset: app_reviews
 templates:
   2da8f134-58db-4f9d-b3b0-8c6b50693ab5: !Template
-    answer_choices: null
-    answer_choices_key: null
+    answer_choices:
+    - Not at all
+    - 'No'
+    - Maybe
+    - 'Yes'
+    - Definitely
+    answer_choices_key: star
     id: 2da8f134-58db-4f9d-b3b0-8c6b50693ab5
-    jinja: 'Given this review: {{review}}
+    jinja: 'Given this review: "{{review}}"
 
-      Would you recommend this app to a friend?
+      Would you recommend this app to a friend? {{answer_choices[0]}}, {{answer_choices[1]}},
+      {{answer_choices[2]}}, {{answer_choices[3]}}, or {{answer_choices[4]}}?
 
       |||
 
-      {% if star==5 %}
-
-      Definitely
-
-      {% elif star==4%}
-
-      Yes
-
-      {% elif star==3%}
-
-      Maybe
-
-      {% elif star==2%}
-
-      No
-
-      {% else %}
-
-      Not at all
-
-      {% endif %}'
+      {{answer_choices[star-1]}}'
     metadata: !TemplateMetadata
       _do_eval: false
       _do_train: true
-      choices_in_prompt: false
-      metrics: []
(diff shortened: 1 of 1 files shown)


promptsource/templates/app_reviews/templates.yaml
```diff
@@ -58,39 +46,24 @@ templates:
...
-
-      **
-
-      {% else %}
-
-      *
-
-      {% endif %}'
+    jinja: "What would be the \u2605-rating of this review (\u2605 being the lowest\
+      \ and \u2605\u2605\u2605\u2605\u2605 being the highest) : {{review}}?\n|||{{answer_choices[star-1]}}"
```
tianjianjiang, line 58: I think this kind of expression (or something like "on a scale of 1 to 5") doesn't count as `choices_in_prompt`, but I can be totally mistaken.

promptsource/templates/app_reviews/templates.yaml
```diff
@@ -58,56 +46,44 @@ templates:
...
       |||
 
       {{star}}'
     metadata: !TemplateMetadata
       _do_eval: false
       _do_train: true
-      choices_in_prompt: true
-      metrics: []
-      original_task: null
+      choices_in_prompt: false
```
tianjianjiang, line 83: I don't think this kind of expression counts as `choices_in_prompt` but I can be totally mistaken.

promptsource/templates/app_reviews/templates.yaml
```diff
@@ -58,56 +46,44 @@ templates:
...
+      metrics:
+      - Accuracy
+      - Spearman Correlation
+      original_task: false
```
srush, line 87: Why isn't this the original task?

promptsource/templates/app_reviews/templates.yaml
srush: Your judgement seems right.

srush (changes requested):
This all seems right, except I don't understand what the task is?

promptsource/templates/app_reviews/templates.yaml
tianjianjiang: That's a relief. Thank you!

promptsource/templates/app_reviews/templates.yaml
tianjianjiang: I also wondered and failed to find the full reference.

The paper seems in a different title https://explore.openaire.eu/search/publication?pid=10.5167%2Fuzh-113425 and the server responded 403. (If we're going to be more thorough, I guess I will have to contact the dataset uploader.)
Nevertheless, judging by the abstract, the dataset curators seem using it for categorizing reviews into a custom taxonomy, instead of associating reviews and stars.

srush: Thanks! That's helpful.

srush (approved):

PR #359, opened 2021-09-07
Clean Rotten Tomatoes

zaidalyafeai:


promptsource/templates/rotten_tomatoes/templates.yaml
```diff
@@ -6,15 +6,16 @@ templates:
     - good
     answer_choices_key: null
     id: 10adbcf1-b839-4522-bd76-567f0c760474
-    jinja: '{{text}} Did the reviewer find this movie good or bad? ||| {{ answer_choices
+    jinja: '{{text}} Did the reviewer find this movie {{"good or bad"}}? ||| {{ answer_choices
```
srush, line 9: Ideally it would be {{answer_choices[0]}} or {{answer_choices[1]}} but I think it is close enough.

srush (approved):

PR #369, opened 2021-09-09
Clean PIQA

arnaudstiegler:
Added original task templates (we now have 5):
- finish_sentence_with_correct_choice
- pick_correct_choice_index
- pick_correct_choice_with_choice_given_before_goal
- what_is_the_correct_ending

Addressing the comments in the spreadsheet: 
- choose_the_most_appropriate_solution_reorder_solution -> removed
- Does_this_solution_make_sense_sol1 and Does_this_solution_make_sense_sol2 -> now using jinja for the choices

Misc changes:
- Deleted : Generate a similar but wrong solution -> very vague, and unnatural task IMO

- Changed template  "Does this solution make sense? sol1" -> so that it’s not a duplicate of "Does this solution make sense? sol1"
- Changed template "Correct the solution if false: from sol 1" -> so that it’s not a duplicate of "Correct the solution if false: from sol 2"

### DIFF
--- a/promptsource/templates/piqa/templates.yaml
+++ b/promptsource/templates/piqa/templates.yaml
@@ -110,7 +110,7 @@ templates:
 
       Choices:
 
-      -  {{sol1}}
+      - {{sol1}}
 
       - {{sol2}}
 
--- a/promptsource/templates/piqa/templates.yaml
+++ b/promptsource/templates/piqa/templates.yaml
@@ -1,41 +1,18 @@
 dataset: piqa
 templates:
-  1a1d13ee-2ef5-4c80-89dc-b9e003d4ef45: !Template
+  16e97a16-c958-4956-bfba-279f88dafd5b: !Template
     answer_choices: null
     answer_choices_key: null
-    id: 1a1d13ee-2ef5-4c80-89dc-b9e003d4ef45
-    jinja: 'Given a goal and a correct solution, generate a similar but wrong solution.
+    id: 16e97a16-c958-4956-bfba-279f88dafd5b
+    jinja: 'Goal: {{goal}}
 
-      Goal: {{goal}}
 
-      Correct solution: {{[sol1, sol2][label]}}
+      Which is the correct ending?
 
-      Wrong solution:
+      - {{sol1}}
 
-      |||
+      - {{sol2}}
 
-      {{[sol1, sol2][1 - label]}}
-
-      '
-    metadata: !TemplateMetadata
-      _do_eval: false
-      _do_train: true
-      choices_in_prompt: null
-      metrics: []
-      original_task: null
-    name: Generate a similar but wrong solution
-    reference: ''
-  45773b41-4d13-4ade-9102-afe1543f5776: !Template
-    answer_choices: null
-    answer_choices_key: null
-    id: 45773b41-4d13-4ade-9102-afe1543f5776
-    jinja: 'Given a goal and 2 solutions, choose the most appropriate solution.
-
-      Goal: {{goal}}
-
-      {{"Solution 1"}}: {{sol2}}
-
-      {{"Solution 2"}}: {{sol1}}
 
       Answer:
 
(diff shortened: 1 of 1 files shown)


promptsource/templates/piqa/templates.yaml
```diff
@@ -157,30 +227,33 @@ templates:
...
     answer_choices_key: null
     id: f42cd457-a14b-465a-a139-d7d2407a3bac
-    jinja: 'Given a sentence, correct it if it doesn''t make sense.
+    jinja: 'Sentence: {{goal}} {{sol1[0].lower() + sol1[1:]}}
 
-      Input: {{goal}} {{sol1[0].lower() + sol1[1:]}}
+      If the sentence does not make sense, correct it so that it does make sense.
 
-      Output:
+      Answer:
```
VictorSanh, line 256: Need something in the template to instruct the model to re-copy if the solution makes sense

promptsource/templates/piqa/templates.yaml
```diff
@@ -129,9 +195,11 @@ templates:
...
       _do_train: true
-      choices_in_prompt: null
-      metrics: []
-      original_task: null
+      choices_in_prompt: false
+      metrics:
+      - BLEU
+      - ROUGE
+      original_task: false
     name: 'Correct the solution if false: from sol 2'
```
VictorSanh, line 203: Need something in the template to instruct the model to re-copy if the solution makes sense

promptsource/templates/piqa/templates.yaml
```diff
@@ -107,8 +172,9 @@ templates:
       _do_eval: false
       _do_train: true
       choices_in_prompt: true
-      metrics: []
-      original_task: null
+      metrics:
+      - Accuracy
+      original_task: true
     name: choose the most appropriate solution
```
VictorSanh, line 178: change that template so that Y is `answer_choices[label]` (i.e. outputting Solution 1 and 2)

promptsource/templates/piqa/templates.yaml
```diff
@@ -1,54 +1,88 @@
...
-      _do_eval: false
-      _do_train: false
       choices_in_prompt: true
-      metrics: []
-      original_task: null
-    name: 'choose the most appropriate solution: reorder solution'
+      metrics:
+      - Accuracy
+      original_task: true
+    name: pick_correct_choice_with_choice_given_before_goal
```
VictorSanh, line 54: disambiguate as too whether the model has to output `Solution 1` or the text in Solution 1

VictorSanh (changes requested):
requested a few changes (mostly disambiguation for multiple choice).

also @stephenbach this is an edge case maybe you've already considered: the choices are in multiple keys: `sol1` and `sol2` instead of being under a single key. any way this can be tagged?
Capture d’écran 2021-09-09 à 12 40 24 PM

VictorSanh:
> also @stephenbach this is an edge case maybe you've already considered: the choices are in multiple keys: `sol1` and `sol2` instead of being under a single key. any way this can be tagged?
> Capture d’écran 2021-09-09 à 12 40 24 PM

Note: #373 adds support for that

arnaudstiegler:
@VictorSanh will update the PR so that we can have the answer choices correctly set up thanks to https://github.com/bigscience-workshop/promptsource/pull/373 as well

arnaudstiegler:
Done

promptsource/templates/piqa/templates.yaml
```diff
@@ -60,11 +94,40 @@ templates:
...
+  94c39589-7bfb-4c09-9337-672369459545: !Template
+    answer_choices: null
+    answer_choices_key: '{{sol1}} ||| {{sol2}}'
+    id: 94c39589-7bfb-4c09-9337-672369459545
+    jinja: 'Finish the following sentence with the best choice: {{goal}}
+
+
+      Choices:
+
+      -  {{sol1}}
```
VictorSanh, line 113:
```suggestion
      - {{sol1}}
```

VictorSanh (approved):
lgtm, thank you!

PR #370, opened 2021-09-09
Clean Quoref 

awebson:
Removed prompts that ask for span indices or non-human language (URL in this case). cc @zaidalyafeai


VictorSanh (approved):
thank you! lgtm, merging now

PR #371, opened 2021-09-09
fix cos_e original task flags

awebson:
Note that cos_e is the same QA data as commonsense_QA but with additional human written explanations. We only use cos_e in training. I marked the QA prompts as original task (of commonsense_qa) while generating explanations as non-original (there is no metric/baseline anyway, and the crowdsourced explanations are not very high quality).

### DIFF
--- a/promptsource/templates/cos_e/v1.11/templates.yaml
+++ b/promptsource/templates/cos_e/v1.11/templates.yaml
@@ -22,7 +22,7 @@ templates:
       choices_in_prompt: true
       metrics:
       - Accuracy
-      original_task: false
+      original_task: true
     name: question_description_option_text
     reference: ''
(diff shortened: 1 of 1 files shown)


VictorSanh:
@awebson I agree with that though I am worried we are deviating from what we have been calling `original_task` (i.e. what was this dataset created for?) and introducing edge cases specific to our batch of experiments.

awebson:
Agree with your concern. Although looks like the cos_e [paper](https://arxiv.org/abs/1906.02361) uses predicting the collected explanations as an auxiliary task, while their ultimate task is still accuracy on the commonsense_qa task. So I think we're safe here?

VictorSanh:
ok, that's fair, makes sense. merging now

PR #373, opened 2021-09-09
Expanded answer_choices_key.

stephenbach:
**Note:** Once this PR is merged, open PRs that set `answer_choices_key` should merge `main` and then open and resave such templates.

This expands the format of answer_choices_key to cover more cases, and provides a way to enter them in the UI.

(Note that the dropdown list of features is removed because that doesn't cover all these cases. A more complex UI in the future is possible, but I wanted to get this done quickly since these are mostly corner cases.)

Examples:

If the answer choices are in an iterable in the example, just enter the key name:
For example, given a dataset with format like
```python
{
  "abstractive_explanation": "a house is the only place that is not likely to sell things",
  "answer": "house",
  "choices": [
    "antique shop",
    "house",
    "dark place"
  ],
  "extractive_explanation": "not for sale",
  "id": "d3b479933e716fb388dfb297e881054c",
  "question": "If a lantern is not for sale, where is it likely to be?"
}
```
just enter `choices`.

For nested keys like
```python
{
  "answerKey": "A",
  "choices": {
    "label": [
      "A",
      "B"
    ],
    "text": [
      "scarce",
      "plentiful"
    ]
  },
  "id": "QRQA-10385-4",
  "para": "Many of the worlds people live with water scarcity, and that percentage will increase as populations increase and climate changes."
}
```
enter `choices ; text`.

For choices with separate keys like
```python
{
  "choice1": "The sun was rising.",
  "choice2": "The grass was cut.",
  "idx": 0,
  "label": 0,
  "premise": "My body cast a shadow over the grass.",
  "question": "cause"
}
```
enter `choice1 ||| choice2`.

This also includes code for applying this logic to examples. To get the choices for an example `example` prompted with a template `template`,
```python
if template.get_answer_choices_key() is not None:
    answer_choices = template.get_answer_choices_key().get_answer_choices(example)
else:
    answer_choices = template.get_answer_choices()
```

### DIFF
--- a/promptsource/promptsource.py
+++ b/promptsource/promptsource.py
@@ -344,7 +344,9 @@ else:
                 st.markdown("##### Answer Choices")
                 st.text(", ".join(template.answer_choices) if template.answer_choices is not None else None)
                 st.markdown("##### Answer Choices Key")
-                st.text(template.answer_choices_key.to_str() if template.get_answer_choices_key() is not None else None)
+                st.text(
+                    template.answer_choices_key.to_str() if template.get_answer_choices_key() is not None else None
+                )
                 st.markdown("##### Jinja")
                 splitted_template = template.jinja.split("|||")
                 st.markdown("###### Prompt + X")
--- a/promptsource/promptsource.py
+++ b/promptsource/promptsource.py
@@ -344,9 +344,7 @@ else:
                 st.markdown("##### Answer Choices")
                 st.text(", ".join(template.answer_choices) if template.answer_choices is not None else None)
                 st.markdown("##### Answer Choices Key")
-                st.text(
-                    template.answer_choices_key.to_str() if template.get_answer_choices_key() is not None else None
-                )
+                st.text(template.get_answer_choices_expr())
                 st.markdown("##### Jinja")
                 splitted_template = template.jinja.split("|||")
                 st.markdown("###### Prompt + X")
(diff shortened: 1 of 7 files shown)


srush:
I like this idea, but it is a little confusing to me. 

* How about replacing "choices ; text" with "choices.text" and just evaluating it as a standard JSON key. 
* I don't totally get the semantics of ||| . It merges together options into a set? What if both of the elements are lists? 

Alternatively, I think this might be a nice opportunity to just use jinja again instead of writing our own language. Like what if you put in a jinja expression that generates options separated by |||. 

For example:

Ex1 -> {{ choices |  join("|||") }}  
Ex2 -> {{ choices.text |  join("|||") }}  
Ex3 -> {{ choice1 }} ||| {{choice 2}}

stephenbach:
Oh yes, I like the idea of using Jinja here.

We can go further and combine `answer_choices` and `answer_choices_key` into a single field because this format supports both. (Maybe you were already thinking this!)

I will hold off on actually merging the two fields though because it will be a pain for the open PRs, and easy to do after the fact. More commits coming.

srush:
Oh yeah, that's true.

stephenbach:
Ok, answer_choices_key is now a Jinja expression. It can also be used in place of answer_choices because its output is put in the `answer_choices` Jinja variable.

I left answer_choices in for backwards compatibility while we have so many open PRs, but before release we can remove the separate field and merge into one field.

srush (approved):
lgtm

PR #375, opened 2021-09-09
clean yelp_review_full

VictorSanh:


awebson:
Many prompts here have a period after each answer. Is this intended?

VictorSanh:
> Many prompts here have a period after each answer. Is this intended?

indeed good catch, corrected it

PR #376, opened 2021-09-09
clean ropes and add more original templates

zaidalyafeai:
This dataset is weird, We don't have the start indices of answers, not sure if this will affect evaluation.


awebson:
That's fine. In fact, it's preferred to answer by actual answer text not span indices.

PR #377, opened 2021-09-09
clean glue qqp

VictorSanh:


promptsource/templates/glue/qqp/templates.yaml
```diff
@@ -54,8 +56,9 @@ templates:
...
       _do_train: true
       choices_in_prompt: false
-      metrics: []
-      original_task: null
+      metrics:
+      - Accuracy
+      original_task: true
     name: same thing
     reference: ''
   c0182cd1-c7ac-4abe-829f-4651536af951: !Template
```
awebson, line 64: I propose that we remove this "can answer to Q1 be used to answer Q2" template. Although it's a valid non-original template, it could cause sneaky problems since Q1 and Q2 could be totally non-paraphrases (the originally intended task) while still sharing the same answer. 

If you disagree, feel free to disregard and merge now.

VictorSanh: yes, that's a fair concern.
I went through a few a couple of examples and I couldn't find any of these cases so I think it's fairly rare.

merging now

PR #378, opened 2021-09-09
clean wiki_hop task/dataset

yongzx:
What I have done:
- [X] fix existing issues
- [X] add 4 more original task templates
- [X] set metrics
- [X] fill out answer_choices
- [X] indicate Choices in Prompt

**Existing issues:**
The existing templates `Choose_Best_Object_Candidate`, `Generate_Fact_Triple`, `Generate_Object_Answer`, `Generate_Subject_Answer` are not in natural form. I have fixed them and rename all the templates to lower case and underscores. I include numbers in template names such as `choose_best_object_interrogative_1` to indicate the variants of the interrogative and affirmative forms. 

As for `Indirect_Question_about_Birthplace_Citizenship_Place_of_Death`, the spreadsheet says template bug. I believe the template bug comes from the template being unable to apply to examples that are not asking about birthplaces and citizenship. Therefore, I have removed the template.

**Potential Issues**
The input sequences are long so we might risk getting the prompt instruction truncated. See [slack thread](https://huggingface.slack.com/archives/C023Z96L7B4/p1631125687080900).

### DIFF
--- a/promptsource/templates/amazon_polarity/templates.yaml
+++ b/promptsource/templates/amazon_polarity/templates.yaml
@@ -17,9 +17,10 @@ templates:
       _do_eval: false
       _do_train: true
       choices_in_prompt: true
-      metrics: []
-      original_task: null
-    name: Template 1
+      metrics:
+      - Accuracy
+      original_task: true
+    name: Is_this_review
     reference: ''
   3a48f287-6a4b-4df0-ab2d-2eaf6cb8e53d: !Template
     answer_choices:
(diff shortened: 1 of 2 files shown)


awebson (changes requested):
Clever templates! Although `Choices: {{ candidates | join("|||") }}` is not joining the candidates correctly for your original task prompts.

Also, it seems like `{{ supports | join("\n- ") }}` is also often only joining a subset of all the `supports` paragraphs?

yongzx:
@awebson 

Fixed `Choices: {{ candidates | join("|||") }}` with `Choices: {{ candidates | join("\|\|\|") }}`.
Fixed `{{ supports | join("\n- ") }}` with a for-loop. I don't know why join doesn't work whereas a for-loop works.

awebson:
Great thanks! One final nit: Could use something else other than `|||` for deliminating your choices? It's not natural English, and it takes up 3 tokens when your input is already long. Just use comma or bullet list like you did for the `supports` paragraphs.

yongzx:
@awebson I have fixed the issue for both PRs. Let me know if I should change anything else.

awebson (commented):
Thank you!

PR #379, opened 2021-09-09
Update winogrande to new answer_choices_key format.

stephenbach:


srush (approved):

srush:
Nice. Just to be sure, you are stripping whitespace right?

stephenbach:
I think the answer to your question is yes. Whitespace is stripped from the choices here: https://github.com/bigscience-workshop/promptsource/blob/main/promptsource/templates.py#L146

Lmk if there's other places we should double check.

PR #383, opened 2021-09-10
clean adversarialQA

VictorSanh:


awebson (approved):
Great thanks! Although for the professor_gave_assignment prompt: 

> I don't like cheating, but the exercise is easy for humans and frankly, a waste of time for me. Could you do it for me?

Although it's not exactly toxic or offensive, it's not the sort of behaviors we want to encourage models to be trained on for 30k instances. Personally I think we should delete it. (It's fine to have less than 4 original task template for this for now.) If you disagree, feel free to merge now.

VictorSanh:
> Great thanks! Although for the professor_gave_assignment prompt:
> 
> > I don't like cheating, but the exercise is easy for humans and frankly, a waste of time for me. Could you do it for me?
> 
> Although it's not exactly toxic or offensive, it's not the sort of behaviors we want to encourage models to be trained on for 30k instances. Personally I think we should delete it. (It's fine to have less than 4 original task template for this for now.) If you disagree, feel free to merge now.

yes that's a good point, I removed them.
merging now

PR #384, opened 2021-09-10
Clean up TREC

arnaudstiegler:
- Addressed comments in the spreadsheet
- added answer-choices everywhere
- added original task ticks
- added prompts for original task
- added metrics

### DIFF
--- a/promptsource/templates/trec/templates.yaml
+++ b/promptsource/templates/trec/templates.yaml
@@ -119,7 +119,7 @@ templates:
     - other entity
     answer_choices_key: null
     id: 3aff84f3-e478-4598-abe8-40aa24cec1fa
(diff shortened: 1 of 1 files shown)


stephenbach:
It's probably not a huge deal since we won't evaluate on them, but it would be good for the answer_choices to be only the ones actually listed in the choices for the prompt. So for the fine grain classification, it should be just the options with a map from the label (over all the choices) to the entry in answer_choices.

arnaudstiegler:
Will update them

arnaudstiegler:
@stephenbach should be good now. I added a mapping from fine_label to the fined grained categories

stephenbach (approved):
LGTM! Thanks!

Issue #380, opened 2021-09-09
Streamlit version and `beta_columns`


VictorSanh:
`beta_columns` was moved out of beta in the latest `streamlit` version (at least `0.88.0` but might have happened before). We should replace these `beta_columns` into `columns` (and pin the streamlit version), otherwise there are warnings popping up everywhere.

VictorSanh:
solved

PR #385, opened 2021-09-10
drop beta_ from beta_columns and beta_expander

yongzx:
Fix issue #380 (including `st.beta_expander`).

### DIFF
--- a/promptsource/promptsource.py
+++ b/promptsource/promptsource.py
@@ -365,7 +365,7 @@ else:
                     continue
                 example = dataset[ex_idx]
                 example = removeHyphen(example)
-                col1, _, col2 = st.beta_columns([12, 1, 12])
+                col1, _, col2 = st.columns([12, 1, 12])
                 with col1:
                     st.write(example)
                 if num_templates > 0:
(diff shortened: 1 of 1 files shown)


srush:
If you do it this way, can you pin version v0.88 on the repo and check that everything else works with that.

yongzx:
I can do that. For checking, it would be to check if the templates are loaded properly and if we can add new templates, right?

srush:
Yeah, just check that it loads properly and doesn't throw any errors for streamlit 0.88.

tianjianjiang:
@yongzx @srush Just FYI, I've been using 0.88 with or without `beta_` since last Friday and it seemed fine.

VictorSanh:
i pinned the streamlit version to `0.82.0` to avoid bugs for `request_rerun()` (see #444) in this commit c16363dd30370526871a98bc8660916e9a999f35.

Might be worth it to investigate #444 before upgrading (and pinning) the streamlit version to fix remove these `beta` tags

tianjianjiang:
> @yongzx @srush Just FYI, I've been using 0.88 with or without `beta_` since last Friday and it seemed fine.

A small clarification: I also encountered #444 but it wasn't related to `beta_` stuff.

VictorSanh:
thanks @yongzx ! i tested and it worked for me. merging now, tests are failing because of community datasets (wip)

PR #386, opened 2021-09-10
Dbpedia check

lintangsutawika:


awebson (changes requested):
This was not your but the original prompt author's fault: Could you please rewrite the templates to be more natural English? "Refer to" reads awkwardly with "what category" or "topic". Maybe say something like "this paragraph talks about a company", "the topic of this passage is an artist". Something like that. Thanks!

lintangsutawika:
No problem. Will change now.

promptsource/templates/dbpedia_14/templates.yaml
```diff
@@ -74,15 +46,16 @@ templates:
     - Written Work
     answer_choices_key: null
     id: 824ecb55-ecad-40c2-8033-f2fa0add2ddf
-    jinja: '{{title}} - {{content}} The text refers to what category? ||| {{ answer_choices[label]
+    jinja: '{{title}} - {{content}} What does the paragraph talk about? ||| {{ answer_choices[label]
```
awebson, line 49: Hmm… Wording it in an interrogative sentence like this still reads weird. Like a normal human is expected to answer the entity's name, not a category name. A declarative sentence like what I suggested earlier is better, but that'd require adding determiner agreement "a company" vs. "an athlete" in the targets. Let's just delete this template for now.

awebson (changes requested):
See my comment below. These answer choices just read weird as answers to the questions as they're phrased. Let's remove all templates that don't have choices_in_prompt for now. Alternatively, you can rewrite some templates to include answer choices in their prompts. Don't feel obligated to have 5 original task templates for now.

promptsource/templates/dbpedia_14/templates.yaml
lintangsutawika: What if I add the indefinite article in the answer choices?

awebson (approved):
Thank you!

PR #388, opened 2021-09-10
Cleanup pass on sciq, wiki_bio, scitail

jason-fries:
Added new templates to sciq and scitail to bring us up to 5 templates
Rephrases the "who" template in wiki_bio


awebson (changes requested):
Thanks! For sciq, could you please:

1. Add the answer_choices_key according to #373
2. In multiple choice prompts, replace your answer deliminator with commas and space, or a bullet list. Right now simply using `|` with no space is not natural English and might cause problems for the tokenizer.
3. For your Direction Question prompt, uncheck `choices in prompts`.

For SciTail: Please note that entailment is always annotated as unidirectional, so it's ill-defined to ask what is the entailment label of a sentence pair without specifying the direction (even though they may be pragmatically inferred). Also, we should provide more explanations on what does "entail" even mean. Anyway, SciTail is actually held out from both training and evaluation, so it's probably not worth your time to fix them now, and I'd recommend just removing them from your PR.

awebson (approved):
Thank you!

awebson:
@stephenbach Would you mind taking a look at why is the check_templates test failing here? Otherwise this PR is ready to be merged. Thanks!

stephenbach:
The error is ```ValueError: Template for dataset scitail/snli_format with uuid 90827988-2a8d-4ecb-b8c1-54ad6cd0ebfa has unrecognized variable label.``` Looking at it now.

promptsource/templates/scitail/snli_format/templates.yaml
```diff
@@ -1,17 +1,26 @@
...
-    id: 707f9870-64dc-4f24-b4e3-3f26297ec28c
-    jinja: "{{sentence1}} \n{{sentence2}}\nAre these sentences {{\"neutral or entailment\"\
-      }} to one another?|||\n{{gold_label}}"
+    answer_choices_key: yes ||| no
+    id: 90827988-2a8d-4ecb-b8c1-54ad6cd0ebfa
+    jinja: 'Given that {{sentence1}} Does it follow that {{sentence2}}
+
+      {{ answer_choices | join('' or '') }}?
+
+      |||{% if label == "entails" %}
```
stephenbach, line 12: Looks like it should be `gold_label == "entailment"`

stephenbach (changes requested):
Left comment on line that needs changing.

stephenbach (approved):

PR #389, opened 2021-09-10
add quac and five prompts

zaidalyafeai:
Dataset description: There are three main entities teacher and a student and the context (wiki article). The student has only access to the background (title + first paragraph of the context) while the teacher has access to the full article. The student asks questions to learn more about the article. The teacher answers questions using spans of the context (or no answer) and provides some directions (follow ups) (whether the student's questions should continue in this direction or not). It is kind of QA dataset in a dialogue fashion.

### DIFF
--- a/promptsource/templates/quac/templates.yaml
+++ b/promptsource/templates/quac/templates.yaml
@@ -1,0 +1,90 @@
+dataset: quac
+templates:
+  12c9d007-991c-49ed-82e2-13a7d3147881: !Template
+    answer_choices: null
+    answer_choices_key: null
+    id: 12c9d007-991c-49ed-82e2-13a7d3147881
+    jinja: "Given the dialogue : \n\nStudent: {{questions[0]}}\n\nTeacher: {{answers.texts[0]\
+      \ | choice}}\n\nThe context : {{context}}\n\nAnswer the question: {{questions[1]\
+      \ }}\n|||\n{{answers.texts[1] | choice }}"
+    metadata: !TemplateMetadata
+      choices_in_prompt: false
+      metrics:
+      - Squad
+      original_task: true
+    name: Answer Given Only First Dialouge
+    reference: ''
+  1d4014c2-7cf7-45d1-8f85-a701b6a65118: !Template
+    answer_choices: null
+    answer_choices_key: null
+    id: 1d4014c2-7cf7-45d1-8f85-a701b6a65118
+    jinja: "Given the dialogue: \n{% for i in range(0, questions | length - 1)%}\n\
+      Student: {{questions[i]}}\n\nTeacher: {{answers.texts[i] | choice}}\n{% endfor\
+      \ %}\n\nThe context: {{context}}\n\nAnswer the question: {{questions | last\
+      \ }}\n|||\n{{answers.texts| last | choice }}"
+    metadata: !TemplateMetadata
+      choices_in_prompt: false
+      metrics:
+      - Squad
+      original_task: true
+    name: Answer Given Full Dialouge
+    reference: ''
+  1f28f748-bc9e-4881-89ae-1d561abf2f2b: !Template
+    answer_choices: null
+    answer_choices_key: null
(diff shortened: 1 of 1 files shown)


promptsource/templates/quac/templates.yaml
```diff
@@ -0,0 +1,90 @@
+dataset: quac
+templates:
+  12c9d007-991c-49ed-82e2-13a7d3147881: !Template
+    answer_choices: null
+    answer_choices_key: null
+    id: 12c9d007-991c-49ed-82e2-13a7d3147881
+    jinja: "Given the dialogue : \n\nStudent: {{questions[0]}}\n\nTeacher: {{answers.texts[0]\
```
srush, line 7: Can you make it more clear in the prompt that that this only is given part of the information that it needs to get the right answer.

promptsource/templates/quac/templates.yaml
```diff
@@ -0,0 +1,90 @@
...
+      \ | choice}}\n\nThe context : {{context}}\n\nAnswer the question: {{questions[1]\
+      \ }}\n|||\n{{answers.texts[1] | choice }}"
+    metadata: !TemplateMetadata
+      choices_in_prompt: false
+      metrics:
+      - Squad
```
srush, line 13: Just want to check that Squad the metric they use? For some reason I thought this was more complex.

promptsource/templates/quac/templates.yaml
```diff
@@ -0,0 +1,90 @@
...
+      original_task: true
```
srush, line 14: I don't think this is the task?

promptsource/templates/quac/templates.yaml
```diff
@@ -0,0 +1,90 @@
...
+    name: 'Answer Converation '
+    reference: ''
+  2c052ef8-adfa-497b-adb1-9e942ad998e0: !Template
+    answer_choices: null
+    answer_choices_key: null
+    id: 2c052ef8-adfa-497b-adb1-9e942ad998e0
+    jinja: "I read an article : {{context}} \n\nThen the following conversation: \n\
```
srush, line 64: "Then the following conversation *occurred*

promptsource/templates/quac/templates.yaml
```diff
@@ -0,0 +1,90 @@
...
+      \ | last}}\n\nYou can use this dialogue to find the answer faster:\n{% for i\
+      \ in range(0, questions | length - 1)%}\nStudent: {{questions[i]}}\n\nTeacher:\
+      \ {{answers.texts[i] | choice}}\n{% endfor %}\n|||\n{{answers.texts | last |\
+      \ choice }}"
+    metadata: !TemplateMetadata
+      choices_in_prompt: false
+      metrics:
+      - Squad
+      original_task: true
+    name: Use Dialouge as Hint
```
srush, line 89: Dialogue

srush (changes requested):

promptsource/templates/quac/templates.yaml
zaidalyafeai: According to the leaderboard http://quac.ai/ they use F1, HEQQ and HEQD. Since our answer spans have vaguer boundaries than the shorter ones in SQuAD, exact match is not a useful metric.

> HEQ measures the percentage of examples
> for which system F1 exceeds or matches human
> F1. We compute two variants: (1) the percentage
> of questions for which this is true (HEQ-Q), and
> (2) the percentage of dialogs for which this is true
> for every question in the dialog (HEQ-D).

promptsource/templates/quac/templates.yaml
zaidalyafeai: I think this template doesn't make sense. Will add better ones in the next commit.

awebson:
Also, this dataset contains many "CANNOTANSWER" in both inputs and targets. Could you please replace them with "Cannot answer"? Thanks!

srush (approved):

PR #395, opened 2021-09-10
re-align _do_train/_do_eval tags with before

VictorSanh:
I messed up in #383 by copying-pasting too fast and not being careful enough with the _do_train and _do_eval tags. This fixes it.

### DIFF
--- a/promptsource/templates/adversarial_qa/dbert/templates.yaml
+++ b/promptsource/templates/adversarial_qa/dbert/templates.yaml
@@ -12,7 +12,7 @@ templates:
       {{question}}'
     metadata: !TemplateMetadata
       _do_eval: false
-      _do_train: false
+      _do_train: true
       choices_in_prompt: false
       metrics:
       - BLEU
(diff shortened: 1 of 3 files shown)


srush (approved):

awebson (commented):
Oh you don't have to worry about the _do_train and _do_eval flags for now. Stephen and I went back and forth about how to best incorporate some metadata helper for experiments, but eventually we decided to not include any internal flags in promptsource and instead just rely on matching the UUID of each prompt to whatever experiment CSV files I need for loading to seqio. 

I think even though you said you copied and pasted prompts here, the UUIDs are still unique? So I think we're good here.

VictorSanh:
> Oh you don't have to worry about the _do_train and _do_eval flags for now. Stephen and I went back and forth about how to best incorporate some metadata helper for experiments, but eventually we decided to not include any internal flags in promptsource and instead just rely on matching the UUID of each prompt to whatever experiment CSV files I need for loading to seqio.

ok cool! we should remove these flags from the yamls down the road since we are not using them there (and especially since these are specific to our batch of experiments)

> I think even though you said you copied and pasted prompts here, the UUIDs are still unique? So I think we're good here.

Yup the uuids are still unique (manually changed them to be unique)

PR #397, opened 2021-09-10
clean qa srl

VictorSanh:
- I put `Other` as metric for the original task. The exact metric is described in Section 5 of the [paper](https://aclanthology.org/D15-1076.pdf)


srush (approved):
These are good prompts!

PR #398, opened 2021-09-10
Clean glue/{{mnli|mnli-matched|mnli-mismatched}}

Iwontbecreative:
Straightforward cleaning following the cleaning instructions, mostly replacing written labels with answer_choices[x].

I also looked at the templates on `multi_nli` (same task, but P4 probably because not in Glue) and imported one from there.


awebson:
Thanks for the PR! Resolved in #640.

PR #399, opened 2021-09-10
fix unused variable in `add_task`

VictorSanh:
First variable is unused in `add_task`.
It worked because `dataset_name` was defined outside of the function (L153).


VictorSanh:
I am confused that the code checks didn't cache this though...

craffel:
Oh gosh. I guess it worked since dataset_name is defined in the outer scope. Thanks for catching.

PR #402, opened 2021-09-11
Add Drop

sbmaruf:
As requested, total 10 prompt added (2 was there earlier).

Please let me know if there are any issues. 

Though I should re-check earlier, one issue,

- For QA prompts, the Metrics we are adding is `Sequence Accuracy`. There should be another metric `F1` score which is not available.  I found one metric named, `COQA F1` which seemed COQA task dependent.


awebson (approved):

sbmaruf:
@awebson 
- [The near duplicate prompts:](https://github.com/bigscience-workshop/promptsource/pull/402#:~:text=removed%20near%20duplicate%20prompts) These are added because in future it would help us to evaluate the std-dev over similar prompts.
- [Removed extremely open-ended generations](https://github.com/bigscience-workshop/promptsource/pull/402/commits/5543b74f193c74a701d3a265ebf44f72b103a7dc) : these are not original task prompt. Do you want to also clean this from [nq_open](https://github.com/bigscience-workshop/promptsource/blob/main/promptsource/templates/kilt_tasks/nq/templates.yaml#L85). from my understanding, I don't see any issue with adding them as `original_task: false`. In a large pre-training, this could anchor knowledge base with specific topics.
- [removed prompts that exceed max seq length](https://github.com/bigscience-workshop/promptsource/pull/402/commits/8bfa0d095e02d6da8f22ef9bc0e3c6016c9c574c) : this one is not related to this pull. This one is related to trivia-qa. Btw how do you know these prompts exceeds `max_seq_length`?

PR #405, opened 2021-09-11
Experiment D4 train/eval mixtures

awebson:
`seqio_tasks/tasks.py` will now read from:

1. Our master datasets [google sheet](https://docs.google.com/spreadsheets/d/1ELEM-BY5E-5Tqy34jguT91ChwAguoWfyDEN0OzOhjSk/edit#gid=1315886539) downloaded as a CSV, which marks which datasets are train or eval sets.
2. Per-template metadata from promptsource, which marks each template's metric and whether it's an original task template. To-do for @craffel: This will also enable rank eval by using either `answer_choices` or `answer_choice_keys` in template.metadata.

@VictorSanh technically this is ready to be cached:
```
seqio.MixtureRegistry.add(
    "clean_tasks",  # rename as train_mixture?
    [task for task in train_mixture if task not in TASK_BLACKLIST],
    default_rate=functools.partial(seqio.mixing_rate_num_examples, maximum=500_000),
)
```
Victor and I talked about it [on Slack](https://huggingface.slack.com/archives/C020GCD4YTV/p1631313745099900) although double checking with @craffel: Am I allowed to rename the mixture name passed to  seqio.MixtureRegistry? I'm afraid that might mess up your job commands or something. Although I might make more mixtures to allow the ablation study we discussed: Train on D4 sets, checkpoint, and then train on GPT-3 sets, checkpoint, and then finally train on SuperGLUE sets.

### DIFF
--- a/promptsource/seqio_tasks/tasks.py
+++ b/promptsource/seqio_tasks/tasks.py
@@ -29,7 +29,7 @@ GET_METRICS = {
     "COQA F1": mt.coqa_f1,
     "Edit Distance": mt.edit_distance,
     # "Mean Reciprocal Rank": mt.accuracy,  # NOTE not in T5?
-    'Other': mt.accuracy,
+    "Other": mt.accuracy,
     # Missing support for mean_multiclass_f1 etc. which need a num_classes parameter
 }
 
@@ -147,30 +147,30 @@ def add_task(dataset_name, subset_name, template_name, task_name=None, split_map
 
 
 train_sets: List[Dict] = []
-eval_sets:List[Dict] = []
+eval_sets: List[Dict] = []
 do_train: List[Tuple[str, Optional[str]]] = []
 do_eval: List[Tuple[str, Optional[str]]] = []
 experiment_path = pkg_resources.resource_filename(__name__, "experiment_D4.csv")
 with open(experiment_path) as exp_file:
     reader = csv.DictReader(exp_file)
     for row in reader:
-        if row['skip']:
+        if row["skip"]:
             continue
-        if row['subset'] == '':
-            row['subset'] = None  # to match promptsource.Template object
-        if row['do_train'] == 'TRUE':
+        if row["subset"] == "":
+            row["subset"] = None  # to match promptsource.Template object
+        if row["do_train"] == "TRUE":
             train_sets.append(row)
-            do_train.append((row['HF_name'], row['subset']))
-        if row['do_eval'] == 'TRUE':
+            do_train.append((row["HF_name"], row["subset"]))
+        if row["do_eval"] == "TRUE":
             eval_sets.append(row)
-            do_eval.append((row['HF_name'], row['subset']))
+            do_eval.append((row["HF_name"], row["subset"]))
 
 train_or_eval = do_train + do_eval
-print(f'Number of training datasets = {len(train_sets)}')
-print(f'Number of evaluation datasets = {len(eval_sets)}')
+print(f"Number of training datasets = {len(train_sets)}")
+print(f"Number of evaluation datasets = {len(eval_sets)}")
 
 all_templates = promptsource.templates.TemplateCollection()
-all_templates.remove('anli')  # Need to special-case ANLI due to weird split conventions
+all_templates.remove("anli")  # Need to special-case ANLI due to weird split conventions
 train_mixture: List[str] = []  # dataset_subset_template
 eval_mixture: List[str] = []
 for dataset_name, subset_name in all_templates.keys:
@@ -206,13 +206,13 @@ for anli_round in ("r1", "r2", "r3"):
         eval_mixture.append(task_name)
 
 # print(train_mixture)
-print(f'Number of training templates = {len(train_mixture)}')
+print(f"Number of training templates = {len(train_mixture)}")
 # print(eval_mixture)
-print(f'Number of evaluation templates = {len(eval_mixture)}')
+print(f"Number of evaluation templates = {len(eval_mixture)}")
 # for i in seqio.TaskRegistry.names():
 #     print(i)
-print(f'Number of SeqIO registered templates = {len(seqio.TaskRegistry.names())}')
-print('^ includes non-original task templates which are excluded from the eval mixture')
+print(f"Number of SeqIO registered templates = {len(seqio.TaskRegistry.names())}")
+print("^ includes non-original task templates which are excluded from the eval mixture")
 # raise SystemExit
 
 
--- a/promptsource/seqio_tasks/tasks.py
+++ b/promptsource/seqio_tasks/tasks.py
@@ -8,16 +8,30 @@ import datasets
 import seqio
 import t5
 import tensorflow as tf
-
+from t5.data.glue_utils import get_glue_metric, get_super_glue_metric
+from t5.evaluation import metrics as mt
 import promptsource.templates
 from promptsource.seqio_tasks import utils
 
 
-# Tasks deemed as clean/useful
-# annotated_tasks = load_annotated_prompts.load_annotated_prompts()
-# CLEAN_TASKS = [t["dataset_subset_template"] for t in annotated_tasks if not t["skip_train"]]
-# CLEAN_EVAL_TASKS = [t["dataset_subset_template"] for t in annotated_tasks if t["do_eval"]]
-# EVAL_METRICS = {t["dataset_subset_template"]: t["metrics"] for t in annotated_tasks if t["do_eval"]}
+GET_METRICS = {
+    "BLEU": mt.bleu,
+    "ROUGE": mt.rouge,
+    "Span Squad": mt.span_squad,
+    "Squad": mt.squad,
+    "Trivia QA": mt.trivia_qa,
+    "Accuracy": mt.accuracy,
+    "Sequence Accuracy": mt.sequence_accuracy,
+    "Pearson Correlation": mt.pearson_corrcoef,
+    "Spearman Correlation": mt.spearman_corrcoef,
+    "MultiRC": mt.multirc_f1_over_all_answers,
+    "AUC": mt.auc,
+    "COQA F1": mt.coqa_f1,
+    "Edit Distance": mt.edit_distance,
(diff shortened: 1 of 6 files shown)


promptsource/seqio_tasks/tasks.py
```diff
@@ -115,7 +114,7 @@ def add_task(dataset_name, subset_name, template_name, task_name=None, split_map
     preprocessors = [
         seqio.preprocessors.tokenize,
         seqio.preprocessors.append_eos,
-        seqio.CacheDatasetPlaceholder(required=False),
+        # seqio.CacheDatasetPlaceholder(required=False),  # TODO
```
craffel, line 117: Why is this commented out? This just tells seqio where to cache.

promptsource/seqio_tasks/tasks.py
```diff
@@ -148,16 +147,51 @@ def add_task(dataset_name, subset_name, template_name, task_name=None, split_map
...
+        if row["do_train"] == "TRUE":
+            train_sets.append(row)
+            do_train.append((row["HF_name"], row["subset"]))
+        if row["do_eval"] == "TRUE":
+            eval_sets.append(row)
+            do_eval.append((row["HF_name"], row["subset"]))
+
+train_or_eval = do_train + do_eval
+print(f"Number of training datasets = {len(train_sets)}")
+print(f"Number of evaluation datasets = {len(eval_sets)}")
```
craffel, line 171: Sort of a nitpick but can you use logging for this, or remove?

promptsource/seqio_tasks/tasks.py
```diff
@@ -170,6 +204,17 @@ def add_task(dataset_name, subset_name, template_name, task_name=None, split_map
...
+
+# print(train_mixture)
+print(f"Number of training templates = {len(train_mixture)}")
+# print(eval_mixture)
+print(f"Number of evaluation templates = {len(eval_mixture)}")
+# for i in seqio.TaskRegistry.names():
+#     print(i)
+print(f"Number of SeqIO registered templates = {len(seqio.TaskRegistry.names())}")
+print("^ includes non-original task templates which are excluded from the eval mixture")
+# raise SystemExit
```
craffel, line 217: Same comment on print statements, and please remove commented-out code.

promptsource/seqio_tasks/tasks.py
```diff
@@ -148,16 +147,51 @@ def add_task(dataset_name, subset_name, template_name, task_name=None, split_map
...
+    if (dataset_name, subset_name) not in train_or_eval:
+        all_templates.remove(dataset_name, subset_name)
         continue
 
-    for template_name in all_templates.get_dataset(dataset_name, subset_name).all_template_names:
+    dataset = all_templates.get_dataset(dataset_name, subset_name)
+    for template_name in dataset.all_template_names:
         add_task(dataset_name, subset_name, template_name)
 
+        DST_name = utils.get_task_name(dataset_name, subset_name, template_name)
```
craffel, line 186: I think to be PEP compatible this should be dst_name

promptsource/seqio_tasks/tasks.py
```diff
@@ -1,44 +1,38 @@
...
+    "Trivia QA": mt.trivia_qa,
+    "Accuracy": mt.accuracy,
+    "Sequence Accuracy": mt.sequence_accuracy,
+    "Pearson Correlation": mt.pearson_corrcoef,
+    "Spearman Correlation": mt.spearman_corrcoef,
+    "MultiRC": mt.multirc_f1_over_all_answers,
+    "AUC": mt.auc,
+    "COQA F1": mt.coqa_f1,
+    "Edit Distance": mt.edit_distance,
+    # "Mean Reciprocal Rank": mt.accuracy,  # NOTE not in T5?
```
craffel, line 32: Which datasets use mrr?

promptsource/seqio_tasks/tasks.py
craffel: It also looks like train_sets and eval_sets are only populated to print their length. I think you can just log the length of do_train/do_eval and not crete those lists.

craffel:
Oh, and feel free to change the mixture name, it's easy to modify the commands.

promptsource/seqio_tasks/tasks.py
awebson: I think none so far. But @stephenbach added it as an option [here](https://github.com/bigscience-workshop/promptsource/blob/26841476fc58682e9749bbd5cf77fb2f98129f22/promptsource/promptsource.py#L509), so I thought I'd put it here as a commented out placeholder.

awebson:
Two main ideas are implemented now: 

First, it now supports setting a max number of examples not just per template but also a per dataset cap, where cap = min(dataset_train_size, 500k) / num_templates. Each template's cap is stored in a dictionary, which is then wrapped in a lambda replacing seqio.mixing_rate_num_examples. (But not for the eval mixture, which doesn't need any cap per our Slack discussion.)

Second, three training mixtures are provided: `d4_train, gpt_train, sglue_train`, which are intended to be trained in that order with checkpoints between. This provides (1) a preliminary ablation study of which mixture contributes the most transfer benefits and (2) a fail-safe, in case D4 holds out too many datasets and our performance is terrible. Of course, we could also train on `d4_train`, do eval, and if our performance is great, just leave the other two training mixtures alone. Feel free to do whatever makes the most sense in terms of TPU availability.

Lastly, there will probably be some seqio debugging which I'm afraid I will have to leave that to you. For example, this error existed before I made any of the above changes:
```
  File "/home/webson/Research/promptsource/promptsource/seqio_tasks/tasks.py", line 302, in <module>
    default_rate=functools.partial(seqio.mixing_rate_num_examples, maximum=500_000),
  File "/home/webson/Research/prompt_env/lib/python3.7/site-packages/seqio/dataset_providers.py", line 1423, in add
    return super().add(name, Mixture, name, tasks, default_rate, **kwargs)
  File "/home/webson/Research/prompt_env/lib/python3.7/site-packages/seqio/dataset_providers.py", line 143, in add
    provider = provider_cls(*provider_args, **provider_kwargs)
  File "/home/webson/Research/prompt_env/lib/python3.7/site-packages/seqio/dataset_providers.py", line 1182, in __init__
    "All Tasks in a Mixture must have the same output features."
ValueError: All Tasks in a Mixture must have the same output features.
```

VictorSanh:
> Lastly, there will probably be some seqio debugging which I'm afraid I will have to leave that to you. For example, this error existed before I made any of the above changes:
> 
> ```
>   File "/home/webson/Research/promptsource/promptsource/seqio_tasks/tasks.py", line 302, in <module>
>     default_rate=functools.partial(seqio.mixing_rate_num_examples, maximum=500_000),
>   File "/home/webson/Research/prompt_env/lib/python3.7/site-packages/seqio/dataset_providers.py", line 1423, in add
>     return super().add(name, Mixture, name, tasks, default_rate, **kwargs)
>   File "/home/webson/Research/prompt_env/lib/python3.7/site-packages/seqio/dataset_providers.py", line 143, in add
>     provider = provider_cls(*provider_args, **provider_kwargs)
>   File "/home/webson/Research/prompt_env/lib/python3.7/site-packages/seqio/dataset_providers.py", line 1182, in __init__
>     "All Tasks in a Mixture must have the same output features."
> ValueError: All Tasks in a Mixture must have the same output features.
> ```

I am running into the same error the caching is crashing... please let me know when you know where it is coming from

craffel:
> First, it now supports setting a max number of examples not just per template but also a per dataset cap, where cap = min(dataset_train_size, 500k) / num_templates. Each template's cap is stored in a dictionary, which is then wrapped in a lambda replacing seqio.mixing_rate_num_examples. (But not for the eval mixture, which doesn't need any cap per our Slack discussion.)

I had thought that at the meeting we agreed to hold off on this until after the paper was out, since we didn't want to deviate too crazy much from the recipe that worked for D3. Do we have highly variable num_templates across different datasets? If not, then this would seem to mainly change our max mixing rate from 500k to 500k/typical num_templates.

> Lastly, there will probably be some seqio debugging which I'm afraid I will have to leave that to you. For example, this error existed before I made any of the above changes:

This would likely mean that the score_eval mixture is getting tasks that aren't score_eval tasks. I'll debug.

craffel:
Re: the "All Tasks in a Mixture must have the same output features." exception - this is actually a completely misleading error message caused by the fact that 
`[task for task in seqio.TaskRegistry.names() if task.endswith("_score_eval")]`
is currently `[]`, since none of the score eval tasks are getting created. I think this is because, in `add_task`, a score eval task is only created when the following conditional passes:
```
    labels = get_label_strings(template)
    if labels:
       ...
```
I assume that `get_label_strings` is no longer the correct method to use, now that we've added all the functionality for annotating answer choices. Do you want to update that function accordingly, or should I?

awebson:
Since I don't entirely understand the plumbing needed for score eval, feel free to go ahead and change `get_label_strings` and anything else that's needed for score eval. Thank you!

promptsource/seqio_tasks/tasks.py
```diff
@@ -144,19 +150,77 @@ def add_task(dataset_name, subset_name, template_name, task_name=None, split_map
...
-    for template_name in all_templates.get_dataset(dataset_name, subset_name).all_template_names:
+    dataset = all_templates.get_dataset(dataset_name, subset_name)
+    num_templates = len(dataset.all_template_names)
+    train_size = gsheet[(dataset_name, subset_name)]["train_size"]
+    if train_size == "":
+        train_size = 0
+    else:
+        train_size = int(train_size)
+    train_size = min(train_size, MAX_EXAMPLES_PER_DATASET)
+    cap = train_size // num_templates
```
awebson, line 201:
```suggestion
    cap = min(train_size, MAX_EXAMPLES_PER_DATASET)
```

awebson, line 201:
```suggestion
    if train_size > MAX_EXAMPLES_PER_DATASET:
        cap = MAX_EXAMPLES_PER_DATASET // num_templates
    else:
        cap = train_size
```

craffel:
I think everything is merged in, and I just tested it and it seems to be fine. So I'm going to merge.

Issue #355, opened 2021-09-06
Linebreaks are rarely showing


VictorSanh:
Single linebreaks are not faithfully displayed on the streamlit interface (viewer mode or sourcing mode).
Example from glue/mrpc:
```jinja
{{sentence1}}
{{sentence2}}
```
will not faithfully display the linebreak between sentence1 and sentence2

stephenbach:
Fixed in #532.

PR #406, opened 2021-09-11
fix: confusion of yaml newline #355

tianjianjiang:
When having a `'` or `"` quoted string, 2 `\n` means 1 newline.
When using block literal (`|`), `\n` and visual newline are the same.

See the summary table of https://stackoverflow.com/a/21699210 and the demo of https://yaml-multiline.info

There's also a decision to be made, please check the code review comments.


promptsource/promptsource.py
```diff
@@ -55,7 +55,7 @@ def reset_template_state():
 
 def show_jinja(t, width=WIDTH):
     wrap = textwrap.fill(t, width=width, replace_whitespace=False)
-    out = highlight(wrap, DjangoLexer(), HtmlFormatter())
+    out = highlight(wrap, DjangoLexer(), HtmlFormatter(lineseparator="<br>"))
```
tianjianjiang, line 58: This is for visual.
Without it, `prompt + X` of some templates in the viewer mode will have different styles per paragraph due to the `\n` vs `<br>` situation.

promptsource/templates.py
```diff
@@ -375,7 +375,7 @@ def write_to_file(self) -> None:
         # We only create the folder if a template is written
         if not os.path.exists(self.folder_path):
             os.makedirs(self.folder_path)
-        yaml.dump(self.format_for_dump(), open(self.yaml_path, "w"))
+        yaml.dump(self.format_for_dump(), open(self.yaml_path, "w"), default_style="|")
```
tianjianjiang, line 378: This is where we may need a consensus.
`default_style` means it will change everything when updating templates.
While it will be consistent and easier to read for the jinja part, the rest may become more complex.

promptsource/templates/glue/mrpc/templates.yaml
```diff
@@ -1,35 +1,49 @@
...
-    reference: ''
-  44c2e6d9-facf-4959-8400-38e0eb8dd3a8: !Template
-    answer_choices:
-    - 'no'
-    - 'yes'
-    answer_choices_key: null
-    id: 44c2e6d9-facf-4959-8400-38e0eb8dd3a8
-    jinja: 'I want to know whether the following two sentences mean the same thing.
-
+      {% endif %}
```
tianjianjiang, line 21: The block scalar's literal style `|(-)` doesn't need single or double quotes, and newlines will guarantee WYSIWYG.

If we want one the jinja to act in this manner, then we will need to modify more places like this: https://stackoverflow.com/a/20863889

promptsource/templates/glue/mrpc/templates.yaml
```diff
@@ -1,35 +1,49 @@
...
+      null
+    "id": |-
+      3b88d2c4-0aeb-4c6d-9ccc-653a388250a5
+    "jinja": |-
+      {% if label == 1 %}
       Paraphrase the following sentence: {{sentence1}}
 
       |||
 
       {{sentence2}}
```
tianjianjiang, line 20: The block scalar's literal style `|(-)` doesn't need single or double quotes, and will guarantee WYSIWYG for newlines.

If we want one the jinja to act in this manner, then we will need to modify more places like this: https://stackoverflow.com/a/20863889

promptsource/promptsource.py
```diff
@@ -55,13 +55,13 @@ def reset_template_state():
...
     st.write(out, unsafe_allow_html=True)
 
 
 def show_text(t, width=WIDTH):
     wrap = [textwrap.fill(subt, width=width, replace_whitespace=False) for subt in t.split("\n")]
-    wrap = "\n".join(wrap)
+    wrap = "<br>".join(wrap)
```
tianjianjiang, line 64: If we use components with IDs instead of st.text()/write(), then we can use `st.markdown("<style> some_id : { white-space: pre-wrap; } </style>", unsafe_allow_html=True)` instead of the `<br>` here.
But I guess we don't want to modify too many places for now.

promptsource/promptsource.py
```diff
@@ -54,14 +54,15 @@ def reset_template_state():
 
 
 def show_jinja(t, width=WIDTH):
-    wrap = textwrap.fill(t, width=width, replace_whitespace=False)
-    out = highlight(wrap, DjangoLexer(), HtmlFormatter())
+    wrap = [textwrap.fill(subt, width=width, replace_whitespace=False) for subt in t.split("\n")]
+    wrap = "\n".join(wrap)
```
tianjianjiang, line 58: Double newlines sometimes got ignored (by textwrap?) so we have to trick it.

tianjianjiang:
```suggestion
    wrap = textwrap.fill(t, width=width, replace_whitespace=False, drop_whitespace=False)
```

This is probably equivalent but nicer.

tianjianjiang (commented):
self-review

promptsource/promptsource.py
tianjianjiang: No, I was wrong. They are not the same and the special treatment still works better than `drop_whitespace=False`,

promptsource/templates/glue/mrpc/templates.yaml
srush: I really dislike the way the new format looks, it seems to break the nice parts of Yaml. I think we should try the more complex version to only change the jinja lines.

promptsource/promptsource.py
srush: I'm a bit confused. There are two question here right, what format to have for the yaml and how to print. But they are independent right?

srush (changes requested):

promptsource/promptsource.py
tianjianjiang: > There are two question here right, what format to have for the yaml and how to print. But they are independent right?

I was probably overthinking it but didn't explain what I considered explicitly, my apologies.
In short, I think we can ignore the yaml part and only be focused on UI representation. Unless #355 also wanted to have a unified yaml format. I suppose #355 doesn't really care about it.

promptsource/templates/glue/mrpc/templates.yaml
tianjianjiang: I guess a third option is as https://github.com/bigscience-workshop/promptsource/pull/406#discussion_r707334059 mentioned that we simply don't enforce yaml format.

stephenbach:
Sorry to interject: would we want to convert all existing templates as part of this PR? Either way, since we're getting ready to launch a training run, we might want to save PRs that affect the template data itself until later.

tianjianjiang:
@stephenbach 
> Sorry to interject: would we want to convert all existing templates as part of this PR? Either way, since we're getting ready to launch a training run, we might want to save PRs that affect the template data itself until later.

No problems at all. As I explained here at https://github.com/bigscience-workshop/promptsource/pull/406#discussion_r707334059, it is probably the best that we don't touch any yaml template. I've already reverted those changes.

tianjianjiang:
I'm going to close this for #511 for the time being.

PR #404, opened 2021-09-11
fix: XSum with "2nd grader" replaced and BLUE added and naming changed

tianjianjiang:
- Mostly for https://github.com/bigscience-workshop/promptsource/pull/360#issuecomment-916532724
- Naming and some wording changed
  - Hopefully name changes may hint structural/orthographic differences
     1. ~Instructive punctuations `:` and `?`;~
     2. ~Custom separators `===` and `"""`~;
     3. Placeholder ~`<DOC>`~ `DOC`.

Issue #326, opened 2021-08-04
Markdown interpretation mess up examples visualisation in prompted dataset viewer mode


NohTow:
In prompted dataset viewer mode, if the template contains two `_`, it is interpreted as markdown syntax and begin writing the example in italic until the second `_`.

This comportement has been observed on `winogrande_winogrande_xs_question_sentence_answer` and `winogrande_winogrande_xs_sentence_question_answer`. A picture of the result is attached.

I don't know how promptsource will behave in case the example contains only one `_`, maybe it will write in italic until the end, maybe it won't interpret it as markdown syntax at all.
cc @VictorSanh 
![image](https://user-images.githubusercontent.com/38869395/128205623-9d521f29-709e-41e3-9345-06c349f337f5.png)

stephenbach:
Fixed in #532.

PR #408, opened 2021-09-12
fix: `template.apply()` with `truncation=False`

tianjianjiang:
See https://github.com/bigscience-workshop/promptsource/pull/404#issuecomment-917533222
----
Weakly related: #326 for it will need a new special treatment of escaping markdown syntax.

### DIFF
--- a/promptsource/app.py
+++ b/promptsource/app.py
@@ -374,9 +374,9 @@ else:
                 if num_templates > 0:
                     with col2:
                         try:
-                            prompt = template.apply(example, highlight_variables=True)
+                            prompt = template.apply(example, truncate=False, highlight_variables=True)
                         except (TemplateSyntaxError, TypeError):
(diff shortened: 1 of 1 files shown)


srush:
This code looks good to me. Are we not doing any truncation though? Does this just move the problem downstream to the model.

srush (approved):

tianjianjiang:
> Are we not doing any truncation though? Does this just move the problem downstream to the model.

As far as I can tell from slack discussions, 1024 is the hard limit anyway?

For the streamlit usability, I think I will send another small PR to add word count next to the title of the output, so at least we will be informed. Or do we actually run a tokenizer to get an even more accurate count?

tianjianjiang:
Since there is already a conflict and the original issue seems non-critical, I'm closing this PR.

PR #410, opened 2021-09-12
Wiqa check

lintangsutawika:
removed very open ended generative prompts


awebson (changes requested):
For the "effect with string answer" template, sometimes the `answer_label` contains underscores, e.g., "no_effect". This is not natural English that we want models to generate. Could you please remove those underscores?

awebson (approved):

PR #411, opened 2021-09-12
fix: process pool's fork safety issue since macOS High Sierra

tianjianjiang:
Long story short: if not replacing `Pool` with `ThreadPool`, 99% of the time the helicopter view will hang on recent macOS.

Code review comments will provide some details.

### DIFF
--- a/promptsource/app.py
+++ b/promptsource/app.py
@@ -1,6 +1,8 @@
 import argparse
 import textwrap
-from multiprocessing import Manager, Pool
+from itertools import repeat
+from multiprocessing import Manager
+from multiprocessing.pool import ThreadPool
 
 import pandas as pd
 import plotly.express as px
@@ -138,11 +140,11 @@ if mode == "Helicopter view":
     all_infos = manager.dict()
     all_datasets = list(set([t[0] for t in template_collection.keys]))
 
-    def get_infos(d_name):
+    def get_infos(d_name, all_infos):
         all_infos[d_name] = get_dataset_infos(d_name)
 
-    pool = Pool(processes=len(all_datasets))
-    pool.map(get_infos, all_datasets)
+    pool = ThreadPool(processes=len(all_datasets))
+    pool.starmap(get_infos, zip(all_datasets, repeat(all_infos)))
     pool.close()
     pool.join()
 


promptsource/promptsource.py
```diff
@@ -103,11 +105,11 @@ def show_text(t, width=WIDTH):
     all_infos = manager.dict()
     all_datasets = list(set([t[0] for t in template_collection.keys]))
 
-    def get_infos(d_name):
+    def get_infos(d_name, all_infos):
```
tianjianjiang, line 108: This isn't mandatory "here" but better be. Normally a function must be defined before Pool init. However, in this case it is almost impossible, because the Pool context is managed by streamlit. Assuming we actually can control the context, then the Pool will encounter `AttributeError` for the implicit/global variable `all_infos`.

promptsource/promptsource.py
```diff
@@ -103,11 +105,11 @@ def show_text(t, width=WIDTH):
...
         all_infos[d_name] = get_dataset_infos(d_name)
 
-    pool = Pool(processes=len(all_datasets))
-    pool.map(get_infos, all_datasets)
+    pool = ThreadPool(processes=len(all_datasets))
```
tianjianjiang, line 111: A safer choice on recent macOS is to replace `fork` with `spawn`. Unfortunately, we can't really change `Pool`'s start method to `spawn` without breaking streamlit. So the only convenient choice is `ThreadPool` (which is also good for I/O bound in this case).

stephenbach:
Hey @tianjianjiang! Thanks for working on this! Shall we revisit now that the paper submission has passed? If you can resolve conflicts, I can review. Thanks!

tianjianjiang:
> Hey @tianjianjiang! Thanks for working on this! Shall we revisit now that the paper submission has passed? If you can resolve conflicts, I can review. Thanks!

@stephenbach You're welcome and sure! I will see what I can do ASAP.

tianjianjiang:
> > Hey @tianjianjiang! Thanks for working on this! Shall we revisit now that the paper submission has passed? If you can resolve conflicts, I can review. Thanks!
> 
> @stephenbach You're welcome and sure! I will see what I can do ASAP.

One of obstacles that I forgot: https://github.com/bigscience-workshop/promptsource/issues/515#issuecomment-940592388

tianjianjiang:
> Ok, we're downgrading to Python 3.7 in #547. But ok if this doesn't get into the first release. We're pinning tonight (Eastern US time).

Not a pressing issue, but a side note: I realize that even when using python 3.7.12, the fork errors sometimes still happen. So I use what this PR does locally.

tianjianjiang:
Close this in favor of #733.

Please kindly note that #733 still use multi-process instead of multi-thread, which may still be an issue if we want to upgrade `streamlit`.

PR #417, opened 2021-09-13
Fix cosmos.

stephenbach:


VictorSanh (changes requested):
these modifications lgtm
I am not sure though for the other templates (`693c47c6-f17c-417a-af70-bc20e71b4ed4`, `71325300-1f16-4a68-97c7-a03457f00cc7`, ...): they have both `answer_choices` and `answer_choices_key`, and it seems like we only need `answer_choices` which is correct

VictorSanh:
edit: can be separate PR as for #418 to unlock Colin

VictorSanh:
per slack discussion, merging now and there will a separate PR

PR #422, opened 2021-09-14
Use new answer_choices template functionality for rank eval

craffel:
I tested this on a few tasks; we might find that it doesn't work on certain tasks (either because answer_choices/answer_choices_key is annotated incorrectly in the template, or because I have a bug in my code).

### DIFF
--- a/promptsource/seqio_tasks/tasks.py
+++ b/promptsource/seqio_tasks/tasks.py
@@ -198,20 +198,20 @@ seqio.MixtureRegistry.add(
 
 seqio.MixtureRegistry.add(
     "clean_tasks",
-    [task for task in CLEAN_TASKS if task not in TASK_BLACKLIST],
+    [task for task in seqio.TaskRegistry.names() if task in CLEAN_TASKS and task not in TASK_BLACKLIST],
     default_rate=functools.partial(seqio.mixing_rate_num_examples, maximum=500_000),
 )
 
 
 seqio.MixtureRegistry.add(
     "clean_eval_tasks",
-    [task for task in CLEAN_EVAL_TASKS if task not in TASK_BLACKLIST],
+    [task for task in seqio.TaskRegistry.names() if task in CLEAN_EVAL_TASKS and task not in TASK_BLACKLIST],
     default_rate=functools.partial(seqio.mixing_rate_num_examples, maximum=500_000),
 )
 
 seqio.MixtureRegistry.add(
     "anli_eval_tasks",
-    [task for task in CLEAN_EVAL_TASKS if task.startswith("anli")],
+    [task for task in seqio.TaskRegistry.names() if task in CLEAN_EVAL_TASKS and task.startswith("anli")],
     default_rate=functools.partial(seqio.mixing_rate_num_examples, maximum=500_000),
 )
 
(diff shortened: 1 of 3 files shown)


promptsource/seqio_tasks/tasks.py
```diff
@@ -42,26 +41,12 @@ def strip_whitespace(output_or_target, example=None, is_target=False):
...
-        labels = eval(label_string_match.group(2))
-        after_label = label_string_match.group(3)
-        labels = [before_label + label + after_label for label in labels]
-        return labels
-
-
 def maybe_get_class_id_postprocessor(template):
-    labels = get_label_strings(template)
-    if labels is not None:
+    if template.answer_choices:
```
stephenbach, line 45: The intended usage was that the answer choices would always be obtained through `get_answer_choices_list(example)`, but this function doesn't have access to a particular example. The issue with this is that we planned to merge answer_choices and answer_choices_key, since they can be represented in the same syntax (a jinja expression). So this access to a list without a specific example was intended to go away before release. Is that a problem?

craffel: So, the problem is that there really are kind of two cases for answer choices:
1. They are static across the entire dataset
2. They vary on a per-example basis (and possibly even vary in length)

(1) is actually I think a lot more common than (2). We need to make the assumption that answer_choices is fixed-length to do things like "map label string to an integer" (e.g. "Yes" -> 0, "No" -> 1), which is necessary for certain metrics. If we force everything into the general case (2) we won't be able to do this. And for case (2), we don't care about doing the things that we want to do with (1), since we can't e.g. map label strings to integers if the label strings differ per-example. So, if we do get rid of answer_choices, it'd be important to still allow things for (1). Not sure the best way to support that.

promptsource/seqio_tasks/tasks.py
```diff
@@ -125,21 +110,22 @@ def add_task(dataset_name, subset_name, template_name, task_name=None, split_map
...
-            inputs_fn=lambda ex: tf.fill((len(labels),), ex["inputs"]),
-            targets_fn=lambda ex: labels,
-            is_correct_fn=lambda ex: tf.equal(labels, tf.strings.strip(ex["targets"])),
+            inputs_fn=lambda ex: tf.fill((len(ex["answer_choices"]),), ex["inputs"]),
+            targets_fn=lambda ex: ex["answer_choices"],
+            is_correct_fn=lambda ex: tf.equal(ex["answer_choices"], tf.strings.strip(ex["targets"])),
             weight_fn=lambda ex: 1.0,
         )
+
+        num_classes = len(template.answer_choices) if template.answer_choices else None
```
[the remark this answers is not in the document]
craffel: Same comment as above.

promptsource/seqio_tasks/tasks.py
stephenbach: Ah darn, sorry, it seemed like a good idea at the time to combine them all into one representation on the annotation side. What if I added a function to templates like `get_fixed_answer_choices_list()` that either returns the fixed answer choices or None if it's not fixed? I would just see if the expression parses without access to any example, otherwise return None.

craffel: That would work great I think.

craffel: Updated code base on #423

stephenbach (approved):
LGTM!

PR #426, opened 2021-09-14
Randomize sciq tasks.

stephenbach:
It's not very pretty.... but not sure how else to do it using the choice function.


VictorSanh:
Yeah agree... Should we add a new filter instead as explained [here](https://stackoverflow.com/a/31608030)? 
That would look something like:
```jinja
{% set answer_choices_shuffled = answer_choices | shuffle %}
```

stephenbach:
Yeah, that's probably worthwhile. But it should call choice so that everything can be seeded for determinism (or something equivalent).

PR #427, opened 2021-09-15
fix adversarial qa for test subset which doesn't contain choices.text

VictorSanh:
Test sets don't contain target annotations, so i wrapped the original tasks under `{% if metadata.split != "test" %} ... {% endif %}` to solve the bugs that arise when trying to access the annotations for the test sets


stephenbach (approved):
LGTM! Merging now.

PR #428, opened 2021-09-15
fix ropes for test split which doesn't contain answers.text

VictorSanh:
Test sets don't contain target annotations, so i wrapped the original tasks under {% if answers.text %} ... {% endif %} to solve the bugs that arise when trying to access the annotations for the test sets.
Same as #427


stephenbach (approved):
LGTM! Merging now.

PR #430, opened 2021-09-15
Cleaning samsum

NohTow:
I cleaned the samsum dataset, mostly by checking the true task checkbox, except for one prompt which was the inverse task (generate dialogue from the summary). I added two prompts so we have at least 5 true task templates.

Also added the metric: ROUGE, since it is the original paper metric.

I also flagged it as very open-ended generation, but I'm unsure of this. I think it is true for the inverse task, but just summarizing is maybe not too open-ended.

### DIFF
--- a/templates/samsum/templates.yaml
+++ b/templates/samsum/templates.yaml
@@ -1,45 +1,114 @@
 dataset: samsum
 templates:
-  3ac01292-4a54-4546-b4e6-c225ae114213: !Template
-    id: 3ac01292-4a54-4546-b4e6-c225ae114213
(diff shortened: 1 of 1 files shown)


stephenbach:
Not sure what's going on... the YAML looks ok, but git says there's a merge conflict because the base branch is too far behind. Can you try rebasing to the latest commit?

NohTow:
I think I missed something when I created this branch.
Was easier to just recreate one and paste the YAML file.
I opened another PR (https://github.com/bigscience-workshop/promptsource/pull/436), you can close this one whenever you want.
I am deeply sorry about the duplicate.

PR #432, opened 2021-09-15
prompt duorc

VictorSanh:
Per discussion on Slack, theare are a bunch of R-rated films though we couldn't find any NC17 films.
The recommendation is that these are fine to include in training


stephenbach (commented):
Looks good. Just a couple of wording edits. Will then copy to paraphrase RC.

stephenbach (approved):
Merging now.

PR #435, opened 2021-09-15
simplify and clean tydiqa templates

VictorSanh:
unfortunately the two splits are not the same...

### DIFF
--- a/promptsource/templates/tydiqa/primary_task/templates.yaml
+++ b/promptsource/templates/tydiqa/primary_task/templates.yaml
@@ -1,127 +1,176 @@
 dataset: tydiqa
 subset: primary_task
 templates:
+  16f11e56-a78d-4e33-bba1-586f9947baf7: !Template
+    answer_choices: null
+    answer_choices_key: Yes ||| No ||| None
+    id: 16f11e56-a78d-4e33-bba1-586f9947baf7
+    jinja: '{% if language == "english" %}
+
+      I wonder {{question_text}}.
+
+      Help me answer this question with "{{answer_choices[0]}}" or "{{answer_choices[1]}}"
+      or "{{answer_choices[2]}}" if none of this two answer apply.
+
+      Here''s what I found on the internet:
+
+      Topic: {{document_title}}
+
+      Article: {{document_plaintext}}
+
+      |||
+
+      {{annotations.yes_no_answer[0] | capitalize}}
+
+      {% endif %}'
+    metadata: !TemplateMetadata
+      choices_in_prompt: true
+      metrics:
+      - Accuracy
+      original_task: true
+    name: heres_what_I_found
+    reference: ''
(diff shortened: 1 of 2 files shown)


promptsource/templates/tydiqa/primary_task/templates.yaml
```diff
@@ -1,127 +1,174 @@
...
       \    {% endif %}\n{% endif %}"
     metadata: !TemplateMetadata
       _do_eval: false
       _do_train: false
-      choices_in_prompt: null
-      metrics: []
+      choices_in_prompt: false
+      metrics:
+      - Accuracy
       original_task: true
```
craffel, line 139: The prompt `open_domain_qa_without_choices` has `original_task: false`, but it seems it is no less original-tasky than this one. Should one or the other be changed?

promptsource/templates/tydiqa/primary_task/templates.yaml
```diff
@@ -1,127 +1,174 @@
...
       _do_eval: false
       _do_train: false
-      choices_in_prompt: null
-      metrics: []
-      original_task: true
-    name: binary_question_without_context_1
+      choices_in_prompt: true
+      metrics:
+      - Accuracy
+      original_task: false
```
craffel, line 156: Same here about `original_task`

promptsource/templates/tydiqa/primary_task/templates.yaml
```diff
@@ -1,127 +1,174 @@
...
+      \ == \"YES\" or annotations.yes_no_answer[0] == \"NO\" %} \nBased on the following\
+      \ text snippet, {{question_text}} \n{{document_plaintext}}\n||| \n{{annotations.yes_no_answer[0]\
+      \ | capitalize}}\n    {% endif %}\n{% endif %}"
     metadata: !TemplateMetadata
       _do_eval: false
       _do_train: false
-      choices_in_prompt: null
+      choices_in_prompt: false
       metrics: []
       original_task: true
```
craffel, line 172: Same here about `original_task`

craffel (commented):
Some questions about `original_task`. Also, can `answer_choices_key` be something like `{{annotations.yes_no_answer[0] | capitalize}}` or would that not work? Otherwise LGTM, please merge away.

VictorSanh:
> Some questions about `original_task`. 

My rationale for `original_task`: if it uses the text of the article then it's original task, otherwise it's closer to opendomain qa and thus didn't tag it as original task. 

> Also, can `answer_choices_key` be something like `{{annotations.yes_no_answer[0] | capitalize}}` or would that not work? 

`{{annotations.yes_no_answer[0] | capitalize}}` would definitely work but I wouldn't know whether it corresponds to Yes or No, and thus don't know which other option to list.

> Otherwise LGTM, please merge away

Merging now

PR #436, opened 2021-09-15
Cleaned samsum dataset

NohTow:
I cleaned the samsum dataset, mostly by checking the true task checkbox, except for one prompt which was the inverse task (generate dialogue from the summary). I added two prompts so we have at least 5 true task templates.

Also added the metric: ROUGE, since it is the original paper metric.

I also flagged it as very open-ended generation, but I'm unsure of this. I think it is true for the inverse task, but just summarizing is maybe not too open-ended.

cc @stephenbach, this PR should be fine. Sorry about that.

### DIFF
--- a/promptsource/templates/samsum/templates.yaml
+++ b/promptsource/templates/samsum/templates.yaml
@@ -88,7 +88,7 @@ templates:
     answer_choices: null
     answer_choices_key: null
     id: 9f571a72-6813-4307-9aae-753ca0f737c5
-    jinja: 'Write a dialogue that match this summary: {{summary}} |||
+    jinja: 'Write a dialogue that matches this summary: {{summary}} |||
 
       {{dialogue}}'
     metadata: !TemplateMetadata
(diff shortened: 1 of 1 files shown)


promptsource/templates/samsum/templates.yaml
```diff
@@ -44,42 +44,71 @@ templates:
...
+      - ROUGE
+      original_task: true
     name: Generate a summary for this dialogue
     reference: ''
   9f571a72-6813-4307-9aae-753ca0f737c5: !Template
     answer_choices: null
     answer_choices_key: null
     id: 9f571a72-6813-4307-9aae-753ca0f737c5
-    jinja: 'Write a dialogue. Summary: {{summary}} |||
+    jinja: 'Write a dialogue that match this summary: {{summary}} |||
```
stephenbach, line 91:
```suggestion
    jinja: 'Write a dialogue that matches this summary: {{summary}} |||
```

stephenbach (approved):
LGTM! Thank you!

PR #438, opened 2021-09-15
prompt kilt/hotpotqa

VictorSanh:
cc @awebson

### DIFF
--- a/promptsource/templates/kilt_tasks/hotpotqa/templates.yaml
+++ b/promptsource/templates/kilt_tasks/hotpotqa/templates.yaml
@@ -64,6 +64,29 @@ templates:
       original_task: false
     name: formulate
     reference: ''
+  9211f663-51f9-428e-ba27-158480eee083: !Template
+    answer_choices: null
+    answer_choices_key: null
+    id: 9211f663-51f9-428e-ba27-158480eee083
+    jinja: '{% if output %}
+
+      FINAL EXAM
+
+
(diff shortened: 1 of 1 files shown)


awebson:
As we cast Hotpot as a closed-book QA, the questions are neither "complex" nor "elaborate" as some of the prompts here describe. I don't know @craffel if you think it's still worth rewriting them now?

craffel:
> As we cast Hotpot as a closed-book QA, the questions are neither "complex" nor "elaborate" as some of the prompts here describe. I don't know @craffel if you think it's still worth rewriting them now?

Eh I think it's fine to describe them in this way, since they do require multi-step reasoning even if the context isn't given. And after all, didn't we decide that "human"-y prompting was not the most pertinent bit here? But, yes, we also could just have a prompt that is literally {{question}} ||| {{answer}} or some such thing.

awebson:
I guess to what extent do we want to enforce humanly/semantically plausible prompt is a separate debate that we can't resolve now. I don't have strong feelings in this particular case though, so feel free to merge unless you find other bugs in these templates.

craffel:
Yes, I guess I am saying: these are reasonably humanly/semantically plausible!

craffel:
The only other prompts I would add would be something like
```
FINAL EXAM

Question 1. {{question}}
|||
{{answer}}
```
or silly things like that. Otherwise LGTM.

PR #439, opened 2021-09-15
print/test all splits in show_templates

VictorSanh:
This PR prints/tests all splits in the sanity test `show_templates`.
This should catch errors we encountered with #427 and #428 early on (i.e. before the PRs are merged and we try caching)


stephenbach (approved):
LGTM!

PR #441, opened 2021-09-16
Wino bias 5 first templates

Iwontbecreative:
Adds 5 templates for the subsets of wino_bias.

The template is a bit complex due to how the dataset is structured in hf.datasets.

### DIFF
--- a/promptsource/templates/wino_bias/type1_anti/templates.yaml
+++ b/promptsource/templates/wino_bias/type1_anti/templates.yaml
@@ -1,0 +1,97 @@
+dataset: wino_bias
+subset: type1_anti
+templates:
+  51578f5e-4913-41d9-b109-054b24a2f0b5: !Template
+    answer_choices: null
+    answer_choices_key: null
+    id: 51578f5e-4913-41d9-b109-054b24a2f0b5
+    jinja: 'What does "{{tokens[coreference_clusters[2] | int : coreference_clusters[3]
+      | int + 1] | join(" ")}}" refer to in the following sentence?
+
+      {{tokens | join(" ")}} ||| {{tokens[coreference_clusters[0] | int : coreference_clusters[1]
+      | int + 1] | join(" ")}}'
+    metadata: !TemplateMetadata
+      _do_eval: false
+      _do_train: false
+      choices_in_prompt: false
+      metrics:
+      - Other
+      original_task: true
+    name: refers_to
+    reference: ''
+  51ae7473-969e-4219-b213-ff857f8fc487: !Template
+    answer_choices: null
+    answer_choices_key: null
+    id: 51ae7473-969e-4219-b213-ff857f8fc487
+    jinja: '{{tokens | join(" ")}}
+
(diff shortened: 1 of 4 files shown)


stephenbach:
Thanks! These are really nice, especially with the tricky format. What is the metric denoted by "Other" in this case?

Iwontbecreative:
The original metric is F1 (mimics the ontonotes format). This did not seem to be available (there is COQA F1 but AFAICT that is different).

stephenbach (approved):
Great, thanks! Merging now.

cc @awebson @craffel this is an eval set, but AFAIK we don't have eval code for plain old F1?

awebson:
[It](https://github.com/google-research/text-to-text-transfer-transformer/blob/f9771263e02a6c01232eac272ebf472c646ce253/t5/evaluation/metrics.py#L216) needs a parameter that specifies the number of classes, which promptsource doesn't have an UI for users to enter. Not sure what's the priority to support to this. I [used to](https://github.com/bigscience-workshop/promptsource/blob/239c348f425e29526bad9c57d70d9d34237b78e2/promptsource/seqio_tasks/preview_annotated_prompts.py#L30) just hardcode the number of classes for eval sets but that code is outdated now.

stephenbach:
I think we could handle that by calling [get_fixed_answer_choices_list](https://github.com/bigscience-workshop/promptsource/blob/239c348f425e29526bad9c57d70d9d34237b78e2/promptsource/templates.py#L150)? It will return None if there isn't a hardcoded set of answer choices. So the length of that list would do it? If so, I can add F1 to the metrics list (any others, or more nuanced versions of F1?) and also add it to this dataset.

awebson:
Oh you're right about using `len(answer_choices)`. Thanks!

The main variant is a token-based F1 popularized by Squad, which I think we already support. There are others like [F1 with invalids](https://github.com/google-research/text-to-text-transfer-transformer/blob/f9771263e02a6c01232eac272ebf472c646ce253/t5/evaluation/metrics.py#L287) for SuperGLUE ReCoRD, but honestly I don't work in QA enough to know how prevalent they are. I think let's leave those for now unless @craffel am I missing something?

stephenbach:
Thanks! Just to clarify, here is the current list of available annotations:
```
# Metrics from here:
                        # https://github.com/google-research/text-to-text-transfer-transformer/blob/4b580f23968c2139be7fb1cd53b22c7a7f686cdf/t5/evaluation/metrics.py
                        metrics_choices = [
                            "BLEU",
                            "ROUGE",
                            "Span Squad",
                            "Squad",
                            "Trivia QA",
                            "Accuracy",
                            "Sequence Accuracy",
                            "Pearson Correlation",
                            "Spearman Correlation",
                            "MultiRC",
                            "AUC",
                            "COQA F1",
                            "Edit Distance",
                        ]
                        # Add mean reciprocal rank
                        metrics_choices.append("Mean Reciprocal Rank")
                        # Add generic other
                        metrics_choices.append("Other")
```
Should this be expanded now that we can check whether there's a fixed number of classes? Perhaps `F1` should be added? And then which one(s) should this dataset get?

PR #445, opened 2021-09-18
New prompts for some superglue subsets

VictorSanh:
MultiRC
- Wrap "@placeholder" in escaping
- Replaced `Span Squad` with `Squad` as per [slack discussion](https://huggingface.slack.com/archives/C023Z96L7B4/p1631106461072000?thread_ts=1631100879.070900&cid=C023Z96L7B4)

COPA
- I wrote 7 original templates that apply the whole set. Not all of the existing templates applied to the whole set (which makes them invalid for reporting eval number unless we combine them in a fancy way...)

RECORD
- wrote more prompts

BoolQ:
- Corrected bug on test set (`label=-1` is equivalent to no label)
- wrote more propmts

### DIFF
--- a/promptsource/templates/anli/templates.yaml
+++ b/promptsource/templates/anli/templates.yaml
@@ -20,7 +20,7 @@ templates:
   179eb863-3ece-4e6f-af0f-fcb46d997306: !Template
     answer_choices:
     - 'Yes'
-    - 'Maybe'
+    - Maybe
     - 'No'
     answer_choices_key: null
     id: 179eb863-3ece-4e6f-af0f-fcb46d997306
@@ -36,7 +36,7 @@ templates:
   5459237b-97de-4340-bf7b-2939c3f7ca19: !Template
     answer_choices:
     - 'Yes'
-    - 'Maybe'
+    - Maybe
     - 'No'
     answer_choices_key: null
     id: 5459237b-97de-4340-bf7b-2939c3f7ca19
@@ -54,7 +54,7 @@ templates:
   620aa3fc-d5eb-46f5-a1ee-4c754527aa97: !Template
     answer_choices:
     - 'True'
-    - 'Neither'
+    - Neither
     - 'False'
     answer_choices_key: null
     id: 620aa3fc-d5eb-46f5-a1ee-4c754527aa97
@@ -66,7 +66,8 @@ templates:
       _do_eval: true
       _do_train: false
       choices_in_prompt: true
-      metrics: []
+      metrics:
+      - Accuracy
       original_task: true
     name: GPT-3 style
     reference: 'Same as reported in Figure G7 of the GPT-3 paper, except that there
@@ -74,7 +75,7 @@ templates:
   9b613182-c6ab-4427-9221-3d68f6d62765: !Template
     answer_choices:
     - 'Yes'
-    - 'Maybe'
+    - Maybe
     - 'No'
     answer_choices_key: null
     id: 9b613182-c6ab-4427-9221-3d68f6d62765
@@ -92,7 +93,7 @@ templates:
   a850110d-f1a3-49b4-949a-d3bfe9f81344: !Template
     answer_choices:
     - 'Yes'
-    - 'Maybe'
+    - Maybe
     - 'No'
     answer_choices_key: null
     id: a850110d-f1a3-49b4-949a-d3bfe9f81344
@@ -108,7 +109,7 @@ templates:
   bab86d5a-4f9c-40db-b619-a7b7d5cae681: !Template
     answer_choices:
     - 'True'
-    - 'Inconclusive'
+    - Inconclusive
     - 'False'
     answer_choices_key: null
     id: bab86d5a-4f9c-40db-b619-a7b7d5cae681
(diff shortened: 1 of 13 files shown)


promptsource/templates/anli/templates.yaml
```diff
@@ -1,27 +1,38 @@
...
-    id: 391acb43-732a-4c09-8c10-e02d226d8854
-    jinja: 'Sentence 1: {{premise}}
-
-      Sentence 2: {{hypothesis}}
-
-      Question: Does Sentence 1 contradict Sentence 2? Yes, No, or Neutral? |||
-
-      {{answer_choices[label]}}'
+    id: 0cc3ae39-3997-4686-8c93-5d51457efa1f
+    jinja: '{{premise}} Using only the above description and what you know about the
```
craffel, line 10: Should there be newlines after premise here and elsewhere? Not sure - wondering if we have any indication of whether it matters.

promptsource/templates/anli/templates.yaml
```diff
@@ -1,27 +1,38 @@
...
+      choices_in_prompt: true
+      metrics:
+      - Accuracy
+      original_task: true
+    name: MNLI crowdsource
+    reference: Adapted from Williams et al. 2018's instructions to crowdsourcing workers.
+  179eb863-3ece-4e6f-af0f-fcb46d997306: !Template
+    answer_choices:
+    - 'yes'
+    - maybe
```
craffel, line 23: any idea why sometimes the answer_choices have quotes and sometimes they don't?

promptsource/templates/anli/templates.yaml
```diff
@@ -1,27 +1,38 @@
...
+    - 'no'
+    answer_choices_key: null
+    id: 179eb863-3ece-4e6f-af0f-fcb46d997306
+    jinja: 'Given {{premise}} Should we assume that "{{hypothesis}}" is true? Yes,
+      no, or maybe? ||| {{ ["yes", "maybe", "no"][label] }} '
```
craffel, line 28: use answer_choices here?

promptsource/templates/anli/templates.yaml
```diff
@@ -69,55 +78,178 @@ templates:
     - 'No'
     answer_choices_key: null
     id: 9b613182-c6ab-4427-9221-3d68f6d62765
-    jinja: '{{premise}} Based on the previous passage, is it true that {{hypothesis}}
+    jinja: '{{premise}} Based on the previous passage, is it true that "{{hypothesis}}"?
       Yes, no, or maybe? ||| {{ answer_choices[label] }}'
```
craffel, line 82: Answer choices have capital first letters - not sure if this matters.

promptsource/templates/anli/templates.yaml
awebson: Sadly we just didn't have enough templates in D3 to tell whether this matters. I threw in a mix of with and without newlines this time so hopefully we will figure out soon.

promptsource/templates/anli/templates.yaml
```diff
@@ -69,55 +78,178 @@ templates:
...
-    - must be false
+    - 'yes'
+    - maybe
+    - 'no'
     answer_choices_key: null
-    id: e0e8b914-c32c-4c85-8b8f-0f61ae79e2b3
-    jinja: Given that {{premise}}, it {{"must be true, might be true, or must be false"}}
-      that {{hypothesis}}? ||| It {{ answer_choices[label] }}.
+    id: bcd90047-3a2b-426b-b065-8a418f1317b8
+    jinja: 'Given that {{premise}} Therefore, it must be true that "{{hypothesis}}"?
```
craffel, line 132: I think it would be slightly more grammatically correct to say "it therefore must be true..."

promptsource/templates/anli/templates.yaml
awebson: I had the same question. Checked YAML's spec and it says quotes don't matter unless there are colons in the string. So probably doesn't matter for now; should be enforced from streamlit's side.

promptsource/templates/anli/templates.yaml
awebson:
```suggestion
      no, or maybe? ||| {{ answer_choices[label] }} '
```

promptsource/templates/super_glue/multirc/templates.yaml
```diff
@@ -52,6 +95,27 @@ templates:
...
+  7bf537ea-ff8d-44c7-8fc9-305b35e3be66: !Template
+    answer_choices: null
+    answer_choices_key: No ||| Yes
+    id: 7bf537ea-ff8d-44c7-8fc9-305b35e3be66
+    jinja: '{{paragraph}}
+
+
+      Decide whether "{{answer}}" is a valid answer to the following question: {{question}}
+
+      Answer yes or no.
```
craffel, line 107: Capitalization? Does it matter?

promptsource/templates/anli/templates.yaml
awebson: I think they're equally correct? "It therefore…" sounds pedantic and Google Ngram says "therefore it" is twice as frequent. https://books.google.com/ngrams/graph?content=therefore+it%2Cit+therefore&year_start=1800&year_end=2019&corpus=26&smoothing=3#

promptsource/templates/anli/templates.yaml
awebson: Good catch. Thanks!

awebson (approved):

stephenbach (approved):
LGTM!

craffel (approved):
Only question is whether we're fixing the label = -1 things now or later.

awebson:
I don't think we need to worry about the test splits for now. There are more urgent to-dos for the paper…

PR #446, opened 2021-09-19
Add mixture of one single original-task prompt per dataset

craffel:


awebson (approved):
Thanks! Looks good to me. I'm not sure which one is more preferable: selecting the first original task template or selecting a random OG template. There might be some side effect as I don't know if promptsource sorts the template alphabetically or implicitly by date added. But probably okay for now.

PR #448, opened 2021-09-20
clean Winogrande

awebson:
@VictorSanh please take a second look. Note that the underscores won't show up in streamlit.


VictorSanh (approved):
lgtm!

PR #467, opened 2021-09-27
Fix detection of fixed answer choices.

stephenbach:
Now a "fixed" set of answer choices is defined as an answer choices expression that doesn't use any variables.

### DIFF
--- a/promptsource/templates.py
+++ b/promptsource/templates.py
@@ -8,8 +8,7 @@ from typing import Dict, List, Optional, Tuple
 import pandas as pd
 import pkg_resources
 import yaml
-from jinja2 import BaseLoader, Environment
-from jinja2.exceptions import UndefinedError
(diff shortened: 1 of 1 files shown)


craffel:
We might want to test this by instantiating the datasets that eventually produced -1 and confirming that the examples are no longer -1.

stephenbach:
I’ve confirmed that the function returns the right results on some different templates. I think to test further, @VictorSanh could you recache and check?

craffel:
No need to recache! Postprocessing is not cached.

craffel:
Looks like this works.

Without this fix: https://colab.research.google.com/drive/1ZvleQkCbAvErJ-U3jAQh-JuXZ3pQ1lwf?usp=sharing

With this fix: https://colab.research.google.com/drive/1ULEQz2wh5welxrLZoZ78iqDr6aMh_mPe?usp=sharing

Merging.

PR #478, opened 2021-09-29
Fix conditional target in SuperGlue

VictorSanh:
Fix plumbing for conditional targets.
When there is no target (for instance on test sets), we should return the prompted input but not the prompted target (which is now by default `<NO LABEL>` on seqio/caching side).
That wasn't the case before as the whole template was wrapped into an if condition, and thus the whole test set (for instance) was just ignored...

This will be needed for SuperGLUE


stephenbach (approved):
LGTM!

PR #479, opened 2021-09-30
Copy rc subset of trivia_qa to unfiltered

craffel:


awebson (approved):
The spreadsheet/CSV columns have been updated a bunch since its last merge so I will update it in a separate PR.

craffel:
Sorry are you saying I should not modify the spreadsheet here or I should? I'd like to push this quickly so that Victor can cache the new tasks if possible.

craffel:
Test is failing because the download is timing out.

awebson:
Sorry for being unclear. Yeah I updated the spreadsheet in #480 so you don't have to.

VictorSanh:
> Test is failing because the download is timing out.

the test split of trivia qa will run forever and eventually the test will time out (because of some conditional template we didn't take into account in the tests yet). feel free to ignore if you are sure

PR #481, opened 2021-10-01
added a prompt for LAMBADA

awebson:
By design, LAMBADA asks a model to predict the last word of a paragraph, so the one obvious prompt (which simply omits the last word of the paragraph) is possibly the only prompt needed for this dataset.

@VictorSanh no need to cache it immediately, but do cache it before you launch the next eval runs if possible.

### DIFF
--- a/promptsource/templates/lambada/templates.yaml
+++ b/promptsource/templates/lambada/templates.yaml
@@ -1,0 +1,15 @@
+dataset: lambada
+templates:
+  d5707bd9-d3cc-4535-b4c1-5c2aee8cb8c7: !Template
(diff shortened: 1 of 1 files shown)


VictorSanh:
lgtm, i can cache.
do you think we'll get decent results considering the truncation strategy which keeps the 1024 first tokens?

awebson:
Correct me if I'm wrong but I think most paragraphs are relatively short?

VictorSanh:
cached and uploaded.
indeed the longest sequence across all splits is of 915 tokens. we should be good

Issue #477, opened 2021-09-29
Bias/fairness quantitative measurements


VictorSanh:
Two goals:
- have quantitative measurements for the paper's Broader impact section
- reflect these in the model cards when we release checkpoints

4 held out evaluation sets identified by the Evaluation WG:
- jigsaw_toxicity_pred
- crows_pairs
- winogender (AXG in SuperGLUE)
- winobias

At the current state, I see
`crows_pairs` and `winobias` prompted, `jigsaw_toxicity_pred` has an opened PR (#451) we need to check, `winogender` needs to be prompted.

Workflow:
- [x] prompt those that were not prompted yet
- [x] making sure these were actually cached (@VictorSanh can have this caching step done fairly quickly)
- [x] evaluation (normal and score rank evaluation) or maybe they have some special evaluation?
- [x] when we know which checkpoints exactly to eval, get the final numbers to report

VictorSanh:
@StellaAthena

awebson:
Pasted from Slack discussion for reference:

> currently in HF datasets https://www.kaggle.com/c/jigsaw-toxic-comment-classification-challenge/data ; what the Eval WG proposed https://www.kaggle.com/c/jigsaw-unintended-bias-in-toxicity-classification

The former might have been added to our dataset spreadsheet while the latter just got added to HF. Still, I get a download error for both though.

Also, Winogender and Winobias can use the prompts from Winogrande.

StellaAthena:
@awebson I am able to download both from Kaggle without issue.

> Also, Winogender and Winobias can use the prompts from Winogrande.

It looks like crow pairs is the only one that needs prompting then. I can take care of that today.

StellaAthena:
Actually, it looks like [crows-pairs](https://github.com/bigscience-workshop/promptsource/blob/main/promptsource/templates/crows_pairs/templates.yaml) has a template already.

Iwontbecreative:
> Pasted from Slack discussion for reference:
> 
> > currently in HF datasets https://www.kaggle.com/c/jigsaw-toxic-comment-classification-challenge/data ; what the Eval WG proposed https://www.kaggle.com/c/jigsaw-unintended-bias-in-toxicity-classification
> 
> The former might have been added to our dataset spreadsheet while the latter just got added to HF. Still, I get a download error for both though.

[...]

PR #482, opened 2021-10-01
Update winogender (super_glue/axg) to match fixed RTE.

stephenbach:
Part of #477.


stephenbach:
This PR is unnecessary because winogender has the labels availlable even in the test split. So its templates are the same as RTE, except for the check for label == -1.

PR #484, opened 2021-10-01
Add unanswerable variants for squad

craffel:


craffel:
@stephenbach what is up with the id001 stuff?

stephenbach:
Not sure why these anchor things would suddenly turn on. Perhaps you upgraded pyyaml and it changed the default? I'm using 5.4.1. You?

stephenbach:
Oh actually this might just be an artifact from copying a reference to the same Metadata object and serializing to the same YAML in that copy script. You should be able to delete them manually. If it still loads, it's fine.

PR #485, opened 2021-10-01
Update wsc273

stephenbach:
Per @awebson, this copies and adapts the current templates in wsc.fixed for wsc273. In addition to converting from yes/no questions to either/or ones, I made the following changes:
 - Deleted two templates that were not natural to convert
 - Change "who is/are" to "who or what is/are" since not all referents are people in this dataset.

Part of #477.


stephenbach:
Closing and resubmitting because I accidentally branched off another PR instead of main.

PR #490, opened 2021-10-03
Add 10 continuation-style ReCoRD prompts

craffel:


awebson:
Thank you for adding these! Is there a way to preview whether these complex Answer Choices Key expressions works?

promptsource/templates/super_glue/record/templates.yaml
```diff
@@ -31,6 +31,41 @@ templates:
...
     reference: ''
+  147656b2-2dad-4028-96c4-f19d57cd1344: !Template
+    answer_choices: null
+    answer_choices_key: '{% for entity in entities[:-1] %} {{ query | replace("@placeholder",
+      entity) }} ||| {% endfor %} {{ query | replace("@placeholder", entities[-1])
+      }}'
+    id: 147656b2-2dad-4028-96c4-f19d57cd1344
+    jinja: "Summary:\n\n- {{ passage.split(\"@highlight\")[1:] | join(\"\\n- \") }}\
+      \ \n\nArticle:\n\n{{ passage.split(\"@highlight\")[0] }}\n ||| {% if ( answers\
+      \ | length ) > 0 %}{{ query | replace(\"@placeholder\", answers[0]) }} {% endif\
```
VictorSanh, line 42: is there a reason why using `answers[0]` rather than `answers | choice`?

promptsource/templates/super_glue/record/templates.yaml
```diff
@@ -45,6 +80,39 @@ templates:
...
     reference: ''
+  64013fb3-1afd-4e5a-8777-b164ca3b8e18: !Template
+    answer_choices: null
+    answer_choices_key: '{% for entity in entities[:-1] %} {{ query | replace("@placeholder",
+      entity) }} ||| {% endfor %} {{ query | replace("@placeholder", entities[-1])
+      }}'
+    id: 64013fb3-1afd-4e5a-8777-b164ca3b8e18
+    jinja: "{{ passage.split(\"@highlight\")[0] }}\n\nSummary:\n\n- {{ passage.split(\"\
+      @highlight\")[1:] | join(\"\\n- \") }} \n\n ||| {% if ( answers | length ) >\
+      \ 0 %}- {{ query | replace(\"@placeholder\", answers[0]) }} {% endif %}"
```
VictorSanh, line 91: move "-" in the target to the input so that the targets only contains the actual target

promptsource/templates/super_glue/record/templates.yaml
VictorSanh: didn't repeat over templates

promptsource/templates/super_glue/record/templates.yaml
```diff
@@ -61,6 +129,39 @@ templates:
...
     reference: ''
+  94577b75-2eac-4eae-b367-3b413c4188c6: !Template
+    answer_choices: null
+    answer_choices_key: '{% for entity in entities[:-1] %} {{ query | replace("@placeholder",
+      entity) }} ||| {% endfor %} {{ query | replace("@placeholder", entities[-1])
+      }}'
+    id: 94577b75-2eac-4eae-b367-3b413c4188c6
+    jinja: "After reading the article, write another sentence to add to it.\n{{ passage\
+      \ | replace(\"@highlight\", \"\\n- \") }} \n ||| {% if ( answers | length )\
+      \ > 0 %}- {{ query | replace(\"@placeholder\", answers[0]) }} {% endif %}\n"
```
VictorSanh, line 140: same comment about moving the "-" from target to input

promptsource/templates/super_glue/record/templates.yaml
```diff
@@ -61,6 +129,39 @@ templates:
...
+    name: Add sentence after (continuation choices)
+    reference: ''
+  9579b54e-4f0f-4e43-8907-af57112cc857: !Template
+    answer_choices: null
+    answer_choices_key: '{% for entity in entities[:-1] %} {{ query | replace("@placeholder",
+      entity) }} ||| {% endfor %} {{ query | replace("@placeholder", entities[-1])
+      }}'
+    id: 9579b54e-4f0f-4e43-8907-af57112cc857
+    jinja: "Please read the following news article and write another sentence to add\
+      \ to it.\n\n{{ passage | replace(\"@highlight\", \"\\n- \") }} \n ||| {% if\
```
VictorSanh, line 155: add a "-" at the end of input

promptsource/templates/super_glue/record/templates.yaml
```diff
@@ -168,6 +286,39 @@ templates:
...
+    name: GPT-3 style (continuation choices)
+    reference: Brown et al. 2020
+  dfa3052f-ede8-42c2-b99a-bc5762c4fdc6: !Template
+    answer_choices: null
+    answer_choices_key: '{% for entity in entities[:-1] %} {{ query | replace("@placeholder",
+      entity) }} ||| {% endfor %} {{ query | replace("@placeholder", entities[-1])
+      }}'
+    id: dfa3052f-ede8-42c2-b99a-bc5762c4fdc6
+    jinja: "Article:\n\n{{ passage.split(\"@highlight\")[0] }}\n\nHighlights:\n\n\
+      - {{ passage.split(\"@highlight\")[1:] | join(\"\\n- \") }} \n\n ||| {% if (\
```
VictorSanh, line 312: same about "-_

awebson:
Unrelated: 
```
{{ passage |
replace("@highlight", "\n- ") }}
```
works 

but `- {{ passage.split("@highlight")[1] | join("\n- ") }}` gives weird results where the second line is divorced from its bullet point. Maybe just a streamlit rendering issue.

VictorSanh (commented):
lgtm, left a few minor comments

stephenbach:
> Thank you for adding these! Is there a way to preview whether these complex Answer Choices Key expressions works?

@awebson Not currently, but I could hack something into the dataset viewer if helpful.

awebson:
@stephenbach Thanks. I trust that Colin made sure the Answer Choices Key expressions works so I don't need a new dataset viewer just for me to review.

awebson (approved):
Agree with Victor's comment that the targets shouldn't include the bullet points. Otherwise seems good.

FLAN also used these continuation style prompts for coreference, which I thought is kinda clever, but they said it's "less effective". But it does work great for GPT-3. So let's see how our models fare…

craffel:
GPT-3 includes the bullet point - the point I think is that you are continuing the list from above. In cases where that's true, I left the bullet point in the target (and actually added it to answer_choices, not the input, because you're right that it should be part of the text that is scored). Also changed all answers[0] to answers | choice.

PR #491, opened 2021-10-03
Community datasets

stephenbach:
This PR allows for including datasets from HF community users in the list of selections. Rather than open the floodgates, there's a hardcoded list of users to include. All of their datasets will be included, regardless of metadata.

### DIFF
--- a/promptsource/utils.py
+++ b/promptsource/utils.py
@@ -95,7 +95,11 @@ def render_features(features):
 
 
 def filter_english_datasets():
-    """Filter English datasets based on language tags in metadata"""
+    """
+    Filter English datasets based on language tags in metadata.
+
+    Also includes the datasets of any users listed in INCLUDED_USERS
+    """
     english_datasets = []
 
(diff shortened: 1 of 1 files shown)


craffel:
LGTM as long as you have indeed confirmed that the desired datasets have shown up!

craffel (approved):

stephenbach:
Yup! Just note that to actually load the datasets, they first have to be installed outside of the streamlit interface by calling
```python
import datasets
datasets.load_dataset('Zaid/coqa_expanded')
datasets.load_dataset('Zaid/quac_expanded')
```

craffel:
Hmm so if there's a task that depends on them, we need to manually install the datasets before we load that task probably?

stephenbach:
Yes, I think that's true for any dataset that requires manual download. Is there something else that would work better?

craffel:
I see, I don't totally understand the manual download distinction but I will just ignore this for now in order to get things done!

PR #495, opened 2021-10-04
trivia_qa - unfiltered : conditional target

VictorSanh:


VictorSanh:
> Might as well fix the one additional non-original task too, right?

this one would have no input but a target since the answers are on the input side. not sure that make sense to only have a target with no input...

craffel:
Looks like the test timed out since it's a big dataset, I think we can merge.

PR #496, opened 2021-10-04
Coqa quac expanded

stephenbach:


craffel:
Looks like the tests don't like the Hub dataset - I assume this is because they'd have to download and prepare locally first?

stephenbach:
Yes, that's right. Will revisit how to organize this more nicely after the deadline.

craffel:
This looks good, we might also want a GPT-3 style one with no "prompt" other than the Q: at the end.

craffel:
(but feel free to merge)

PR #498, opened 2021-10-05
Community names and rename app

stephenbach:
This PR does two things:
- Correctly parse dataset names from the community. Since they have slashes and are in subdirectories, we can't tell which is a subset and which is part of a user's datasets without looking at our list of users to include.
- Rename the main module to app.py. This is to enable the correct handling of absolute imports. It seems that streamlit, at least up to 0.88, can neither handle relative imports nor absolute imports if the module has the same name as the package. (Can anyone reproduce? Perhaps I'm missing something, but this is the only fix I've found.) We had hacked around it up to this point by using absolute imports to submodules, but this breaks when we want modules to start importing from each other. Hence, why these two changes are coupled.

### DIFF
--- a/promptsource/app.py
+++ b/promptsource/app.py
@@ -9,9 +9,17 @@ from jinja2 import TemplateSyntaxError
 from pygments import highlight
 from pygments.formatters import HtmlFormatter
 from pygments.lexers import DjangoLexer
+
 from promptsource.session import _get_state
 from promptsource.templates import Template, TemplateCollection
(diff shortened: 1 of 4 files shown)


stephenbach:
Do not merge yet. This breaks the UI, but if used outside streamlit, should be ok.

stephenbach:
Ok, this is fixed now. But we really need to fix the imports. I thinks the issue is that streamlit doesn't like absolute imports if the package and module have the same name. But I will leave that for another day.

VictorSanh:
ok, lgtm otherwise.
merging now

PR #502, opened 2021-10-06
head_qa QA

srulikbd:
-updating all as original task templates
-"Does the output contain only the answer?" - deleting the "answer number" from the answer pharase.


stephenbach:
Thanks! Would you mind pulling main? As for resolving the conflicts, you can take your version, just delete the _do_train and _do_eval lines in all the metadata objects.

PR #519, opened 2021-10-08
remove legacy stuff

zaidalyafeai:


stephenbach:
Thanks! Can you also give everything a read through and check for other things to remove? I think @VictorSanh 's points were just examples. I notice also `Join the Hackaprompt and help writing templates!` in the README, for instance.

And for the `"If you want to prompt a new dataset or add more prompts to an old dataset, "
        + "open an issue [promptsource](https://github.com/bigscience-workshop/promptsource/issues)."` line in the UI, should we point them to the contributing guide? I don't think they need to open an issue if they're going to do the work.

zaidalyafeai:
Yes, sure. Is there a list of datasets to be added to `promptsource` later ? I was thinking that there might be datasets that we don't want to include, so we encourage contributors to start an issue first. We can also remove such instructions from the UI because we  can write them in the README anyways.

PR #523, opened 2021-10-08
Issues/510: Add an argument for read-only

tae898:
At the moment, streamlit does not handle python script arguments gracefully.
Thus, for read-only mode, you have to type one of the below two:

`streamlit run promptsource/app.py -- -r`
`streamlit run promptsource/app.py -- --read-only`

Check https://github.com/streamlit/streamlit/issues/337 for more information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,12 @@ There are 3 modes in the app:
 
 <img src="assets/promptsource_app.png" width="800">
 
+## Running (read-only)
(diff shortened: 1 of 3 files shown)


tae898:
> Beautiful!
> 
> I had been meaning to change the sidebar name to "Promptsource" anyway, so did that in the suggestions. If that's ok with you, please accept them and I will merge. Thanks!

alright sure! thanks!

PR #524, opened 2021-10-08
catch get_dataset_infos error for community datasets 

VictorSanh:
Some community datasets don't have `dataset_infos.json`.
This will return an error in the helicopter view

Also drastically reduced the number of workers in multiprocessing (which I think was causing some issues).

### DIFF
--- a/promptsource/app.py
+++ b/promptsource/app.py
@@ -1,6 +1,6 @@
 import argparse
 import textwrap
-from multiprocessing import Manager, Pool, cpu_count
+from multiprocessing import Manager, Pool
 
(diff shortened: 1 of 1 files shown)


tianjianjiang:
> @VictorSanh @tianjianjiang Does this PR also address the issue in #411? Is there anything from #411 that should be included here?

@stephenbach +CC @VictorSanh I will have to test it again to be sure. However, last time I tested it, `cpu_count()` wouldn't fix #411, and `cpu_count()` was much slower than `len(en_datasets)`, probably because it was I/O bound.

PR #530, opened 2021-10-11
add seqio mixtures for bias/fairness eval

VictorSanh:
cf #447 
could you have a quick look @awebson please?


VictorSanh:
thanks @awebson !

@stephenbach could you have a look at PR #529 ? the prompts we originally wrote are not for the jigsaw subset split the eval WG has decide on. We need the `toxicity pred` subset, not the `unintended bias` subset....

PR #531, opened 2021-10-11
same pull request as before

srulikbd:
this one https://github.com/bigscience-workshop/promptsource/pull/502


stephenbach:
Thanks! The tests are failing because these templates don't have metadata objects. This should be created by the editor automatically. Not sure why they're not here. Maybe all the metadata was taken out instead of just _do_eval and _do_train? Let me know if I can help with anything.

stephenbach:
@zaidalyafeai No I don't think so! 🙂  I think it's just something that came up from branching off a commit earlier than your changes, and then trying to get it to match.

VictorSanh:
i wasn't able to push on this PR so i took over in #688

PR #532, opened 2021-10-12
fix rendering issues: remove highlighting and show raw text for templates

zaidalyafeai:


stephenbach:
Do we still need pygments for anything? It should either be cut from imports or added to requirements.txt.

zaidalyafeai:
@stephenbach not sure how `pygments` is related. I think it is used for the jinja syntax highlighting?

PR #536, opened 2021-10-12
Try load_dataset

craffel:


stephenbach:
Maybe easier to just get rid of all the error handling. It's not adding much. So replace utils.get_dataset with
```python
def get_dataset(path, conf=None):
    "Get a dataset from name and conf."
    builder_instance = get_dataset_builder(path, conf)
    if builder_instance.manual_download_instructions is None and builder_instance.info.size_in_bytes is not None:
        builder_instance.download_and_prepare()
        return builder_instance.as_dataset()
    else:
        return datasets.load_dataset(path, conf)
 ```
The "fail==True" check is only adding a little "Skip for now" message, and if you take it out, the errors are actually pretty informative:
image

stephenbach:
Replaced by #538.

PR #550, opened 2021-10-15
align vocabulary in app with the paper

VictorSanh:
Most of the changes are motivated by how we defined things in the paper: `We define a \textit{prompt} as consisting of an input template and target template, along with a collection of associated metadata. The templates are functions mapping a data example into natural language for the input and target sequences.`

### DIFF
--- a/promptsource/app.py
+++ b/promptsource/app.py
@@ -61,6 +61,10 @@ state = _get_state()
 # Initial page setup
 #
 st.set_page_config(page_title="Promptsource", layout="wide")
+st.sidebar.markdown(
+    "<center><h3><a href='https://github.com/bigscience-workshop/promptsource'>💻Github - Promptsource</h3></a></center>",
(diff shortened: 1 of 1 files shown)


stephenbach:
Actually there's still a few places in the sourcing form that refer to templates, including the box for the Jinja itself, the delete button, and the reference field, plus most of the help text and error messages.

VictorSanh:
I made the repo link less prominent. Wanted to put it as a footer in the sidebar, but couldn't find a way to do that easily in streamlit. Let me know if you have something else in mind.

Capture d’écran 2021-10-17 à 3 25 54 PM

Template/Prompt -> Fixed the few missing places, sorry for that. Kept the difference between template and prompt in particular in "Choices in template" as technically, it is the function that transforms the example (the template) that lists the choices

stephenbach:
Looks great, thanks!

PR #565, opened 2021-10-27
Add metrics to the ade-corpus-v2 dataset

tae898:

### DIFF
--- a/promptsource/templates/ade_corpus_v2/Ade_corpus_v2_classification/templates.yaml
+++ b/promptsource/templates/ade_corpus_v2/Ade_corpus_v2_classification/templates.yaml
@@ -19,8 +19,8 @@ templates:
     answer_choices: No ||| Yes
     id: 78c4ce65-dd66-46ed-878d-11f4eca5e544
     jinja: "Read the below text and answer the question.\n\nText: {{text}} \n\nQuestion:\
-      \ Is the above text related to adverse drug effect (ADE)? Please answer it with\
-      \ either \"Yes\" or \"No\".\n\n|||\n{{answer_choices[label]}}"
+      \ Is the above text related to adverse drug effect (ADE)? Your answer should\
+      \ be either \"Yes\" or \"No\".\n\n|||\n{{answer_choices[label]}}"
     metadata: !TemplateMetadata
       choices_in_prompt: true
       metrics:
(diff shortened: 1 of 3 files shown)


promptsource/templates/ade_corpus_v2/Ade_corpus_v2_classification/templates.yaml
```diff
@@ -2,28 +2,28 @@ dataset: ade_corpus_v2
...
+      metrics:
+      - Accuracy
       original_task: true
-    name: baseline
+    name: binary-classification
     reference: ''
   78c4ce65-dd66-46ed-878d-11f4eca5e544: !Template
-    answer_choices: Yes, it is related to adverse drug effect. ||| No, it is not related
-      to adverse drug effect.
+    answer_choices: Yes ||| No
```
stephenbach, line 16: This doesn't follow the required format of responding A or B if those are used to indicate choices. More generally, these just seem like weird ways of phrasing the question. Maybe both the classification tasks should just use Yes and No as the targets.

stephenbach (changes requested):
Thanks! I think this dataset still needs some work.

- For the classification subset, I left a note about a minor formatting thing.
- For the other two, I think more work is needed. A lot of the prompts don't use the contextual information, resulting in prompts like
```Input

What drug has a dosage of 1500 mg/m2?
Target

methotrexate
```
This doesn't really make sense, because there's no context from which to answer the question. I think making these more like `squad` would be a lot better.

tae898:
Thank you for the nice feedback again. I've made some changes. Please let me know if further changes are necessary.

stephenbach:
Thanks! These look great now! Just wondering, why are no prompts marked original for the relation subsets?

tae898:
> Thanks! These look great now! Just wondering, why are no prompts marked original for the relation subsets?

AFAIK, the original authors created this dataset to train a machine-learning based classifier. The annotations from the relation subsets were used to compare the annotations between the human annotators, but they were not used to train an automatic classifier.

stephenbach:
Ah, gotcha. I think it would be fine to mark them as original task, because it's still the intended task, even if they originally used it more for testing humans.

tae898:
> Ah, gotcha. I think it would be fine to mark them as original task, because it's still the intended task, even if they originally used it more for testing humans.

alright!

stephenbach (approved):
Thanks for all your hard work on this PR! Merging now!

PR #568, opened 2021-10-27
Add metrics to the docred dataset

tae898:

### DIFF
--- a/promptsource/templates/docred/templates.yaml
+++ b/promptsource/templates/docred/templates.yaml
@@ -58,7 +58,7 @@ templates:
       choices_in_prompt: false
       metrics:
       - Accuracy
-      original_task: true
+      original_task: false
     name: ner-type
     reference: ''
   412b482e-185b-48da-8aef-4a93a42e779d: !Template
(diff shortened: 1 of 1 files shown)


stephenbach (changes requested):
Great! A few thoughts:
- Let's change the metrics to Other, because it requires parsing the output and counting individual matches. Accuracy should only be used when counting exact matches to the target.
- There's a bug in rendering at least some of the prompts, where the last token of a sentence (usually a period) doesn't have a space before the first token of the next sentence.
- In the "find all numbers" prompt, I would clarify that it's asking just for numbers expressed in words (not numerals).
- For ner-type, please take out the example (remember to ask for its label instead), so that it remains zero-shot. There's also a lot of blank lines. (Maybe a big code block?)
- Also for ner-type, answer choices should not be set, because that's not the list of all possible outputs.
- relation is also not zero shot.

stephenbach:
Thanks! I think the whitespace thing is important to fix because it affects tokenization. Perhaps there's a way to concatenate all the sentences together before joining. At the very least, you could just put a space each sentence in the for loop I think?

stephenbach:
Sorry, I missed your question about the numbers. It looked to me like it doesn't include numerals?

tae898:
> Thanks! I think the whitespace thing is important to fix because it affects tokenization. Perhaps there's a way to concatenate all the sentences together before joining. At the very least, you could just put a space each sentence in the for loop I think?

Alright I added a white space between setnences.

tae898:
> Sorry, I missed your question about the numbers. It looked to me like it doesn't include numerals?

There are some:
image

image

stephenbach (approved):
Looks great! Thanks! I think I was seeing days and years, which aren't tagged. Thanks for checking.

PR #569, opened 2021-10-27
Add metrics to the tmu_gfm dataset

tae898:

As for the regression tasks, since there isn't a metric for a regression task (e.g., MSE, MAE, etc.), I marked them as "Other".


stephenbach:
Looks great! Are none of these original tasks though?

tae898:
> Looks great! Are none of these original tasks though?

The metric values (fluency, grammar, and meaning) that were annotated by humans were used by the author of this paper to come up with their own metric. So yeah, I think I can say that the three metric values can also be considered as original tasks.

stephenbach:
Great, thanks!

PR #570, opened 2021-10-27
Add metrics to the turk dataset

tae898:

### DIFF
--- a/promptsource/templates/turk/templates.yaml
+++ b/promptsource/templates/turk/templates.yaml
@@ -34,7 +34,7 @@ templates:
 
 
       One of the texts above is more verbose than the other. Which one is the verbose
-      one?
+      one? Your answer should be either "Text A" or "Text B".
 
       |||
 
@@ -49,7 +49,7 @@ templates:
 
 
       One of the texts above is more verbose than the other. Which one is the verbose
-      one?
+      one? Your answer should be either "Text A" or "Text B".
 
       |||
 
@@ -78,7 +78,7 @@ templates:
       choices_in_prompt: false
       metrics:
       - BLEU
-      - Squad
+      - ROUGE
       original_task: false
     name: make-verbose
     reference: ''
@@ -94,7 +94,7 @@ templates:
 
 
       One of the texts above is a simplification of the other. Which one is the simplified
-      one?
+      one? Your answer should be either "Text A" or "Text B".
 
       |||
 
@@ -109,7 +109,7 @@ templates:
 
 
       One of the texts above is a simplification of the other. Which one is the simplified
-      one?
+      one? Your answer should be either "Text A" or "Text B".
 
       |||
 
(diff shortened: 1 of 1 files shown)


tae898:
> Thanks! Minor thing: are none of these original tasks?

Now, I marked the simplification tasks as original tasks.

> More major thing: a lot of the simplifications are duplicates of the original, resulting in prompts like this: image Is there a way to filter out duplicates, and return a prompt from the remaining ones (if any are left)?

Yeah unfortunately the data is not so clean. Among the eight possible simplified answers, the target is chosen randomly with the below code:

```jinja
{{simplifications | choice}}
```

I don't think there is a way to select only simplified texts...

stephenbach:
It's not pretty, but you can get such a list called `real_simplifications` by inserting this code at the top of the template:
```jinja2
{% set real_simplifications = [] %}{% for text in simplifications %}{% if text != original %}{% set tmp = real_simplifications.append(text) %}{% endif %}{% endfor %}
```
Keep it all on one line to avoid inserting a bunch of new lines.

tae898:
> It's not pretty, but you can get such a list called `real_simplifications` by inserting this code at the top of the template:
> 
> ```
> {% set real_simplifications = [] %}{% for text in simplifications %}{% if text != original %}{% set tmp = real_simplifications.append(text) %}{% endif %}{% endfor %}
> ```
> 
> Keep it all on one line to avoid inserting a bunch of new lines.

That's a very good suggestion. Your suggestion only works when there is an exact string match, which doesn't look like it happens often. So I made a modification to it:

```jinja
{% set real_simplifications = [] %}{% for text in simplifications %}{% if text|length < original|length %}{% set tmp = real_simplifications.append(text) %}{% endif %}{% endfor %}
``` 

I kinda like this solution. If you also like it, I can add this in the `CONTRIBUTING.md` as well.

tae898:
I had to add a small trick to get it working:

```jinja
{% set real_simplifications = [] %}{% for text in simplifications %}{% if text|length < original|length %}{{real_simplifications.append(text) | default("", True)}}{% endif %}{% endfor %}
```

stephenbach:
Great! I think we still need to wrap the templates in an if statement to check if the len of real_simplifications in non-zero?

tae898:
> Great! I think we still need to wrap the templates in an if statement to check if the len of real_simplifications is non-zero?

That's a good point. Although I haven't seen such a case, it'd be to safe to add such a thing. 

I can't really code until Monday (CET timezone). If you want to merge this fast, you can add the if statement and merge the PR.

stephenbach:
No rush! Sorry, I'm swamped too. :) We could also just run a quick check and see if any are actually zero. If not, should be fine.

tae898:
@stephenbach done!

stephenbach:
Looks great! I verified that it outputs a blank result when there are no real simplifications. Thanks!

PR #575, opened 2021-10-27
Clean coqa

nihalnayak:
This PR cleans COQA prompts. Added additional prompts to ensure there are five original task prompts.


VictorSanh:
looking into this now

VictorSanh:
ok, i think it was something with the download (server wasn't responsive during the whole download...), I just relaunched the test and it worked

nihalnayak:
@VictorSanh I think if the templates look okay, then the PR can be merged.

VictorSanh:
should we look into this again? seems like a duplicate

PR #577, opened 2021-10-27
Clean mathqa

nihalnayak:
This PR cleans the mathqa dataset. Added more original prompts to the task. The only weird thing about this dataset was the options/answer_choices. I've made sure it uses `answer_choices` but it doesn't feel right. Please let me know if I need to fix something here.

### DIFF
--- a/promptsource/templates/math_qa/templates.yaml
+++ b/promptsource/templates/math_qa/templates.yaml
@@ -1,18 +1,16 @@
 dataset: math_qa
 templates:
   091ba88e-d208-4a3a-ada7-d9698aeb5568: !Template
-    answer_choices: 'a ||| b ||| c ||| d '
+    answer_choices: a ||| b ||| c ||| d  ||| e
     id: 091ba88e-d208-4a3a-ada7-d9698aeb5568
-    jinja: '{% set mapping = {"a": 0, "b": 1, "c": 2, "d": 3} %}
-
-      You will receive full credit if you solve the following word problem:
+    jinja: 'You will receive full credit if you solve the following word problem:
 
       {{Problem}}
 
 
       {{options}}|||
 
-      {{answer_choices[mapping[correct]]}}'
+      {{correct}}'
     metadata: !TemplateMetadata
       choices_in_prompt: true
       metrics:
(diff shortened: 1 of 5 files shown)


promptsource/templates/math_qa/templates.yaml
```diff
@@ -30,25 +54,54 @@ templates:
 
       {{options}}|||
 
-      {{correct}}'
+      {{answer_choices[mapping[correct]]}}'
```
VictorSanh, line 57: you can use `{{correct}}` directly (also the mapping is missing "e" but since I suggest using `correct` directly, you don't need the mapping!

promptsource/templates/math_qa/templates.yaml
```diff
@@ -1,26 +1,50 @@
...
+      choices_in_prompt: false
+      metrics:
+      - BLEU
+      - ROUGE
+      original_task: false
+    name: annotated_formula
+    reference: The template generates the annotated formula
   6312d599-8ca4-4bc8-a76f-81f2e36727bd: !Template
-    answer_choices: null
+    answer_choices: a ||| b ||| c ||| d
```
VictorSanh, line 43:
```suggestion
    answer_choices: a ||| b ||| c ||| d ||| e
```

VictorSanh: same in other places

promptsource/templates/math_qa/templates.yaml
```diff
@@ -30,25 +54,54 @@ templates:
...
+      metrics:
+      - Accuracy
+      original_task: true
+    name: pick_the_correct
+    reference: The template picks the correct the answer choice
+  a313a5f8-53cd-4b76-abb6-fea2ac4e9ef4: !Template
+    answer_choices: a ||| b ||| c ||| d
+    id: a313a5f8-53cd-4b76-abb6-fea2ac4e9ef4
+    jinja: "{% set mapping = {\"a\": 0, \"b\": 1, \"c\": 2, \"d\": 3} %}\nOne of the\
+      \ five choices are correctly answer the math problem. Can you choose the right\
```
VictorSanh, line 96:
```suggestion
      \ five choices correctly answers the math problem. Can you choose the right\
```

promptsource/templates/math_qa/templates.yaml
```diff
@@ -62,25 +115,40 @@ templates:
...
-      metrics: []
-      original_task: null
-    name: Template_4
-    reference: ''
+      choices_in_prompt: false
+      metrics:
+      - BLEU
+      - ROUGE
+      original_task: false
+    name: linear_formula
```
VictorSanh, line 141: this prompt is extremely weird. it doesn't give the model the possible operations (multiply, add, const, etc...), so I'd expect to have a 0 perf... same for `annotated_formula`.
either we remove it, or you can write instructions on which operations to use...

promptsource/templates/math_qa/templates.yaml
```diff
@@ -30,25 +54,54 @@ templates:
...
+      \ one? \n\n{{options}}\n\nProblem: {{Problem}}\n|||\n{{answer_choices[mapping[correct]]}}"
+    metadata: !TemplateMetadata
+      choices_in_prompt: true
+      metrics:
+      - Accuracy
+      original_task: true
+    name: first_choice_then_problem
+    reference: First give the list of choices and then describe the problem
   a3c2ec72-4af5-42aa-9e8e-ef475fa7c039: !Template
     answer_choices: null
```
VictorSanh, line 106: The categories present in the dataset are the following:
```
>>> Counter(d["train"]["correct"])
Counter({'c': 6653, 'b': 6308, 'd': 6085, 'a': 6010, 'e': 4781})
>>> Counter(d["train"]["category"])
Counter({'general': 13273, 'physics': 7063, 'gain': 5120, 'geometry': 2117, 'other': 1814, 'probability': 450})
>>> Counter(d["test"]["category"])
Counter({'general': 1301, 'physics': 702, 'gain': 558, 'geometry': 200, 'other': 172, 'probability': 52})
>>> Counter(d["validation"]["category"])
Counter({'general': 2006, 'physics': 1056, 'gain': 743, 'geometry': 308, 'other': 288, 'probability': 74})
```

Could you update this field?

VictorSanh (changes requested):
Mostly lgtm!
Left a few comments:
- missing "e" in a few answer_choices
- you can use `{{correct}}` directly without using the mapping
- I don't expect model to me able to outputting the formulas because it's not natural language so I would remove them

VictorSanh:
thank you @nihalnayak , merging now. i updated the spreadsheet

PR #579, opened 2021-10-27
Clean squadshifts

nihalnayak:
This PR contains prompts for the Squadshift dataset. Removed a few templates related to counting. All the subsets have the same prompts as they only differ in their source, i.e., Amazon, NYT, and new_wiki. 

EDIT: temporarily removed reddit subset of the dataset.


VictorSanh (approved):
lgtm! thank you @nihalnayak
pushed a few edits:
- `non-answers` -> `answers`
- conservatively remove CBQA template as it is way too open-ended
- remove title when the title wasn't super informative (`Amazon review 123`)

PR #586, opened 2021-10-29
gutenberg_time

yongzx:
Completed the following:
- [X] add templates (≥ 5 original templates)
- [X] add metrics
- [X] indicate answer choices if exists (which even includes the phrase of "between 0 and 23" in prompts)
- [X] add answer choices
- [X] remove templates whose instructions/questions come after the long paragraph (risk of them getting cut off).
- [X] rename some templates.


VictorSanh (approved):
lgtm!
added a few corrections per slack conversation on choices in prompt

PR #587, opened 2021-10-29
Cleaned humicroedit

yongzx:
On top of just cleaning the existing prompts, here are some decisions I've made:
- [X] In subtask 1, I judge if the task follows the original task based on its usage of regression metrics (in this case, "Other). 
- [X] In subtask 1, I remove some templates that have minimal changes in surface-form wordings.
- [X] In subtask 2, if the prompt requires the model to type out the full sentence (even though it is a multiple choice answer), I would indicate that answer choices not in template. This is in prompt `only_edited_sentences_text`. In this case, I set the metrics as BLEU and ROUGE.
- [X] rename templates.

Discussion
For the regression tasks in subtask 1, I do not put a tick for the "Choices in Template?" even though I include the text of "0.0 (not funny) to 3.0 (very funny)" because to me, those are not answer *choices*.

### DIFF
--- a/promptsource/templates/humicroedit/subtask-1/templates.yaml
+++ b/promptsource/templates/humicroedit/subtask-1/templates.yaml
@@ -1,27 +1,6 @@
 dataset: humicroedit
 subset: subtask-1
 templates:
-  43af1016-6d25-434d-b9b0-893706cda5d6: !Template
-    answer_choices: null
-    id: 43af1016-6d25-434d-b9b0-893706cda5d6
-    jinja: 'Please give a score between 0 and 3 denoting the funniness of the following
-      sentence. Your score should be something like {{"1.5"}}, where {{"0.0 means
-      not funny and 3.0 means funny"}}.
-
-
-      {{ original.replace(original[original.index("<"):original.index(">")+1], edit)
-      }}
-
-      |||
-
-      {{ (((5 * meanGrade) | round) / 5) }}'
-    metadata: !TemplateMetadata
-      choices_in_prompt: false
-      metrics:
-      - Other
-      original_task: true
-    name: example_score_edited_sent_affirmative
-    reference: ''
   692750f4-b4a2-4344-bc4d-e05daef47c25: !Template
     answer_choices: null
     id: 692750f4-b4a2-4344-bc4d-e05daef47c25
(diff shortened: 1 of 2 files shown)


stephenbach:
Thanks. A few thoughts:
- For subtask 1, some ask how funny a sentence is, and some ask how much funnier a sentence is compared with the original. It seems like only of these could be the correct interpretation of the annotations? You could just delete the ones that don't fit (whichever that is)
- For metrics, the paper just says they compute "correlations." I haven't looked closely, but I think marking them as pearson correlation is fine.
- Agree that choices in prompt should be false, but at least one prompt has it marked true.
- In subtask 2, why not make the options the answer choices when requiring to output the sentence? I would use accuracy as metric and mark choices in template as true. Even though they're long, it's still a multiple choice question.

yongzx:
@stephenbach 
1. For subtask 1, I believe I only use funny instead of funnier (which is for subtask-2). The reason of my choice is that the prompted output is the funniness of the edited sentence. Should I remove the prompts that compare edited sentence to the original sentence in subtask 1?
2. I just reread the paper and realized that for the regression task, they construct it as the classification task. I would make the changes (from regression to classification task, so the metrics is simply accuracy.) I believe the correlation metrics doesn't apply here because they compute the correlation between the funniness and Glove distances of the replaced words.
3. Thanks for pointing that out!
4. I can simply remove that template in question (which uses BLEU/ROUGE as metrics) because I believe if I were to transform it back to a multiple-choice prompt (i.e., output A/B/C instead of the sentence), it wouldn't be much different than other prompts on the surface level.

stephenbach:
- For subtask 1, prompts like
```
Please give a score between 0 (not funny) and 3 (very funny) denoting the funniness of replacing "{{ original[original.index("<")+1:original.index("/>")]
```
indicate to me it's how funny is the difference. I guess it's subjective, but it reads to me like a relative comparison.
- You could add some classification ones and leave (some of) the regression ones as non-original tasks.
- 👍 
- Sounds good.

yongzx:
- Fixed by removing non-comparative templates. 
- Apologies for this. I just double-check the SemEval task ([here](https://competitions.codalab.org/competitions/20970#learn_the_details-evaluation)) and they use RMSE for the evaluation. The original task is a regression task. 
- Fixed (the template in question is a non-comparative template in subtask 1 so it's removed)
- Fixed by removing `only_edited_sentences_text`. The rest of the templates for task2 use accuracy as the metric, as in the original task.

stephenbach:
Great, thanks!

PR #589, opened 2021-10-29
cleaned acronym identification

teelinsan:
Cleaned acronym_identification dataset

### DIFF
--- a/promptsource/templates/acronym_identification/templates.yaml
+++ b/promptsource/templates/acronym_identification/templates.yaml
@@ -42,7 +42,6 @@ templates:
     metadata: !TemplateMetadata
       choices_in_prompt: false
       metrics:
-      - AUC
       - Other
       original_task: true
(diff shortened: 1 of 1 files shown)


stephenbach:
Thanks! This seems like a really interesting/useful dataset. A few thoughts:
- Should we just present the tokens as-is like text, rather than inserting commas? Even though the dataset comes pre-tokenized, leaving them more like natural language seems better for generalizing to new tasks.
- I would just set all the metrics to Other. AUC is assuming something like accuracy where it's checking if the output matches the target, but this dataset needs special parsing.

teelinsan:
Thanks for the code review!
- I think that using space is more like natural language, but it might be difficult to debug (difficult to see the difference between a single or double space). Changed commas with space in the last commit for now.
- I agree, removed AUC from the dataset

stephenbach (approved):
Looks great! Thanks!

PR #590, opened 2021-10-29
cleaned multi_nli

teelinsan:
Cleaned multi_nli dataset


awebson:
These are duplicates from an outdated collection of NLI prompts. The updated ones are under ANLI and SuperGLUE CB. I will copy them over to MNLI and close this for now. Thanks for the PR though!

PR #593, opened 2021-10-29
clean art

zaidalyafeai:


stephenbach:
Thanks! I think all of the ones that ask to output a sentence should have answer choices set, since it's either or?

stephenbach:
Also changing assignment to @awebson since this is an NLI one.

zaidalyafeai:
Thanks, Stephen. I fixed the templates.

promptsource/templates/art/templates.yaml
```diff
@@ -1,34 +1,37 @@
...
+    answer_choices: '{{hypothesis_1 | trim(''.?!'') }} ||| {{hypothesis_2 | trim(''.?!'')
+      }}'
     id: 151d0e97-d7d2-47f2-86b4-6777587b16f2
-    jinja: "We know that:\n\n{{ observation_1  | trim('.?!') }},\n\nand:\n\n{{ observation_2\
-      \ }} \n\nWhat is more likely?\n\nFirst option: \n\n{{ hypothesis_1  | trim('.?!')\
-      \ }}, \n\nor second option:\n\n{{ hypothesis_2  | trim('.?!') }}?\n|||\n{{ [hypothesis_1,\
-      \ hypothesis_2][label-1]}}"
+    jinja: "We know that:\n\n{{ observation_1 }},\n\nand:\n\n{{ observation_2 }} \n\
+      \nWhat is more likely?\n\nFirst option: \n\n{{ answer_choices[0] }}, \n\nor\
+      \ second option:\n\n{{ answer_choices[1] }}?\n|||\n{{ answer_choices[label-1]\
```
awebson, line 9:
```suggestion
      \nWhich one is more likely?\n\nThe first option: \n\n{{ answer_choices[0] }}, \n\nor\
      \ the second option:\n\n{{ answer_choices[1] }}?\n|||\n{{ answer_choices[label-1]\
```

awebson (changes requested):
Thanks Zaid! I suggest rephrase or simply remove the following ones:

1. The "reversed" ones are not meaningfully different from each other and can artificially lower the performance variance. 
2. In `is_hyp_1 true` and `is_hyp_2 true`, you asked which one is "true". But unlike standard NLI, AbductiveNLI is not about whether the premise is strictly true, but about which option is more likely (as you noted). 
3. In `choose_hypothesis_likely`, "which version is more accurate" sounds kinda vague. "accurate" in what? (I know you mean "more likely" here, but someone who hasn't read the AbductiveNLI paper most likely won't know.)

zaidalyafeai:
I don't feel that removing templates is a good practice. Can't we just exclude/blacklist them if they are not relative to the current use cases? Maybe we can use them as bad examples for the coming paper?

zaidalyafeai:
Between, I didn't create such templates, so I can't defend why they were written in such format.

awebson:
Well, there is a difference between prompts that read a little awkward and those that are misleading descriptions of the task in its original paper. I'm afraid the ones I pointed out are closer to the latter. 

@stephenbach Aside from the "Reference" field where you can put arbitrary string, I don't think promptsource has a metadata field for marking which ones "should be used" and which ones are saved for special purposes? 

For example I'd love to write a bunch of slight variations of one prompt for minimal pair experiments, but I want to clearly mark them with some "special experiment X" string field and be excluded by default. (Otherwise we'd artificially deflate the variance of our prompt performance.) But then I don't know if we actually want people to store their own experimental prompts and merge them into this repository?

zaidalyafeai:
My suggestion is to keep such prompts and maintain a list of prompts that we feel are not good (and why). Then we can use those prompts in the analysis paper. After that, we can decide if we want to remove/keep such prompts.

zaidalyafeai:
Changed accurate to likely. For 1. and 2. I can remove their associated templates if this is the final decision?

awebson:
Yup thanks for the "accurate" to "likely" change. And yeah removing 1 and 2 sounds good. 

I asked @stephenbach for a "final decision" but he hasn't responded yet. If we were to maintain a separate list in a Google Sheet or something, we'd have to ship/advertise that list to users of promptsource, who'd then have to implement stuff on their own to manually exclude prompts in the bad list? In which case we should just include a experiment tag + filter option in promptsource itself.

zaidalyafeai:
@awebson for the experimental prompts, I suggest doing that in a forked version of promptsource for the reasons you mentioned. For the time being, I will remove 1. and 2. since we don't have a mechanism of filtering.

awebson (approved):
Great! Thank you!

PR #595, opened 2021-10-29
MQP cleaning

Iwontbecreative:
Mainly adding answer_choices to prompts and shortening them since they were sometimes overly complex.


stephenbach (approved):
Beautiful! Thanks!

PR #600, opened 2021-10-29
Improve Scicite templates, add metrics/choices

Iwontbecreative:
Added some diversity to templates as they were very similar.


stephenbach (approved):
Beautiful!

PR #604, opened 2021-10-30
add cleaned templates for esnli dataset

MagedSaeed:
This pull request updates the cleaned templates for the esnli dataset

### DIFF
--- a/promptsource/templates/esnli/templates.yaml
+++ b/promptsource/templates/esnli/templates.yaml
@@ -1,7 +1,7 @@
 dataset: esnli
 templates:
   16206cff-dd01-408e-b0b4-daabb750b38b: !Template
-    answer_choices: null
+    answer_choices: '{{explanation_1}} ||| {{explanation_2}} ||| {{explanation_3}}'
     id: 16206cff-dd01-408e-b0b4-daabb750b38b
     jinja: 'Explain why the relation between the following two sentences can be described
       as {{ ["an entailment", "neutral", "a contradiction"][label] }}.
@@ -14,7 +14,7 @@ templates:
 
       |||
 
-      {{explanation_1}}'
+      {{ answer_choices |select("!=","") |list |choice }}'
     metadata: !TemplateMetadata
       choices_in_prompt: false
       metrics:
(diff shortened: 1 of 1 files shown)


awebson:
Thanks for the PR! I removed the extra bullet point dashes in target templates, as well as correcting a common mistake of "sentences contradict each other": In NLI, entailment and contradiction are always unidirectional, not bidirectional. 

Also, all explanation generation templates only use explanation_1, whereas three are given for the validation set. Can you update your templates to randomly choose one among (nonempty) three?

MagedSaeed:
Hey @awebson 

Thanks for your comments and the commits you did. Regarding the issue you raised about the explanations, I found that the train split of the dataset contains only `explanation_1`. Other explanations are empty. However, I checked the other splits and found them nonempty.

image

It is not that straightforward to shuffle the list and pick the nonempty explanations without creating custom template tags/filters in the template.py, I guess. What do you think?

MagedSaeed:
@awebson

Any comments how to move forward? Should I move forward with random nen empty selection? Waiting for your inputs.

awebson:
Sorry for the delay! We got occupied by the paper rebuttals. 

I think it's possible to write a template that randomly choose from 3 options? @stephenbach what's the best way to do this in jinja? The tricky part is that we need to (separately?) check if 3 string variables are non-empty (see the data schema below), join them into a list, and then do a random `choice`?

```
{
  "premise": "string",
  "hypothesis": "string",
  "label": [
    "entailment",
    "neutral",
    "contradiction"
  ],
  "explanation_1": "string",
  "explanation_2": "string",
  "explanation_3": "string"
}
```

MagedSaeed:
Thanks for the reply @awebson I will try my best and let you know soon.

MagedSaeed:
After discussing it with @zaidalyafeai, the following snippet works like a charm!

```
{{ answer_choices |select("!=","") |list |choice }}
```

MagedSaeed:
@awebson The PR is updated.

awebson:
Beautiful. Thank you!

PR #605, opened 2021-10-31
add cleaned templates for snli dataset

MagedSaeed:
this PR adds the cleaned templates for snli dataset


awebson:
Thanks for the PR! Resolved in #640.

PR #608, opened 2021-10-31
Fix sem eval 2010 task 8

srulikbd:
fixing:
-templates names
-metrics


VictorSanh:
thanks for working on this @srulikbd !
I removed the changes for another file.
I moved stuff to answer_choices to avoid overcrowding the template, I fixed the metric to the official one, and removed some weird additional `'`
merging now

PR #612, opened 2021-11-01
delete-duplicate-boolq

sbmaruf:
Please review `answer_choices` in label.


VictorSanh:
Could you expand on the differences with superglue/boolq? templates and dataset

sbmaruf:
It seems like there are two version of boolq template.
- boolq
- superglue/boolq
Earlier I cleaned the `boolq` but `superglue/boolq` seems already cleaned and more prompts are there.
In the new commit, I am just importing (syncing) prompts from `superglue/boolq` into `boolq`. 
Please review.

VictorSanh:
@stephenbach any thoughts on modifying things that were stabilized in v0.1?

sbmaruf:
Please note that,
`boolq` was not cleaned earlier. `superglue/boolq` was cleaned earlier and used in P3. 
Here I'm just syncing the previous  uncleaned`boolq` dataset from `superglue/boolq`.

stephenbach:
Yes, it's just updating the other version of the dataset, so good to do.

@sbmaruf I get the following output from check_equal.py:
```
/Users/bach/anaconda3/envs/promptsource/bin/python /Users/bach/promptsource/check_equal.py
('boolq', None) with name GPT-3 Style has mismatched jinja
('boolq', None) with name GPT-3 Style has mismatched answer_choices
('boolq', None) with name GPT-3 Style has mismatched answer_choices_key
('boolq', None) with name I wonder… has mismatched jinja
('boolq', None) with name after_reading has mismatched jinja
('boolq', None) with name based on the following passage has mismatched jinja
('boolq', None) with name based on the previous passage has mismatched jinja
('boolq', None) with name could you tell me… has mismatched jinja
('boolq', None) with name exam has mismatched jinja
('boolq', None) with name exercise has mismatched jinja
('boolq', None) with name valid_binary has mismatched jinja
('boolq', None) with name yes_no_question has mismatched jinja
('boolq', None) with name yes_no_question has mismatched answer_choices
('boolq', None) with name yes_no_question has mismatched answer_choices_key
```

Can we just make them identical with copy_templates.py?

stephenbach:
Sorry, I might have run the script on the wrong branch. One minute please.

VictorSanh:
actually to avoid confusion and smooth out future maintenance, how about we just keep one version (the one under super_glue/boolq for instance)?

sbmaruf:
Actually, both of the dataset data class structure is not exactly the same. 
for `super_glue/boolq`, 

```
{
  "question": "string",
  "passage": "string",
  "idx": "int32",
  "label": [
    "False",
    "True"
  ]
}
```

and for `boolq`,

```
{
  "question": "string",
  "answer": "bool",
  "passage": "string"
}
```

Will it work if the data class structure is same? @stephenbach

stephenbach:
Yes, that would be even easier :P There are a few other places where we didn't follow this rule, but we could establish it now. The only argument against I can think of would be if they are not exactly the same datasets, perhaps even processed in a slightly different way, and someone wants one and not the other.

sbmaruf:
> actually to avoid confusion and smooth out future maintenance, how about we just keep one version (the one under super_glue/boolq for instance)?

-- I also think it's a good option.

VictorSanh:
> Yes, that would be even easier :P There are a few other places where we didn't follow this rule, but we could establish it now. The only argument against I can think of would be if they are not exactly the same datasets, perhaps even processed in a slightly different way, and someone wants one and not the other.

yes that makes sense. in this case, that's exactly the same right?

sbmaruf:
What I also worry that, just because of 1 or 2 duplicate datasets, people will question our claim of "annotating X number of prompts".

stephenbach:
Sgtm. Happy to just delete it.

sbmaruf:
Closing the pull request.

stephenbach:
@sbmaruf Maybe you could submit PR/reopen to delete prompts? Thanks!

sbmaruf:
@stephenbach  
Should I delete the `promptsource/templates/boolq` folder?

VictorSanh:
> Should I delete the `promptsource/templates/boolq` folder?

yes that sounds right

sbmaruf:
Done!

PR #613, opened 2021-11-01
Fever::clean-and-new-prompt

sbmaruf:
Two important things to consider here,

**NOT ENOUGH INFO related issue**

Earlier the prompt is written in the following format,
```
"{{claim}}", I have heard. Correct?
|||
{{
{"SUPPORTS": "Yes",
 "REFUTES": "No",
"NOT ENOUGH INFO": ""
}[label]
}}
```
In this format, `NOT ENOUGH INFO` will replace empty label. I assumed these samples will automatically filtered away from the dataset.

So I wrote a new prompt, 

```
"{{claim}}", I have heard. Is this Correct? Yes, No or Not Sure?
|||
{{
{"SUPPORTS": "Yes",
 "REFUTES": "No",
"NOT ENOUGH INFO": "Not Sure"
}[label]
}}
```

**Answer Choice**
Though the prompts are not written with `answer_choices`, I wrote the expected class names in this field.

### DIFF
--- a/promptsource/templates/fever/v1.0/templates.yaml
+++ b/promptsource/templates/fever/v1.0/templates.yaml
@@ -19,7 +19,7 @@ templates:
     jinja: "I've heard that {{claim}} Is this correct?\n|||\n{{\n{\"SUPPORTS\": \"\
       Yes\",\n \"REFUTES\": \"No\",\n\"NOT ENOUGH INFO\": \"\"\n}[label]\n}}"
     metadata: !TemplateMetadata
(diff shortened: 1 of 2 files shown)


VictorSanh:
yes you are right, i removed the `"NOT ENOUGH INFO": ""` cause it messes up with the unlabled splits (we need to put a conditional target in the case of unlabeled sets, but in the case we don't know how to turn a 3 way classification into a 2 way classification....)

sbmaruf:
Looks good! @VictorSanh

PR #618, opened 2021-11-01
cleaned hotpot_qa

yongzx:
Tasks for all subsets.
- [X] rename templates to be more meaningful
- [X] indicate original tasks
- [X] add metrics (see Discussion)
- [X] fix whitespace controls in the answer generation
- [X] make the output contain only answers, except `generate_answer_and_explanations`. 

Discussion
In the paper, the metrics for evaluating original tasks are F1 and EM. However, in the dataset presented in promptsource, there are no answer choices (as they are part of the {{ question }}). In this case, does the use of COQA F1 and Other correct metrics? Should I indicate that answer choices exist in the prompts?

### DIFF
--- a/promptsource/templates/hotpot_qa/distractor/templates.yaml
+++ b/promptsource/templates/hotpot_qa/distractor/templates.yaml
@@ -1,65 +1,86 @@
 dataset: hotpot_qa
 subset: distractor
 templates:
+  0eff7d44-d78a-47b5-9526-8c38f74be939: !Template
+    answer_choices: null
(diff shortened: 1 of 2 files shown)


VictorSanh (approved):
lgtm! thank you for working on this @yongzx !

I changed coqa f1 to squad which encompasses em and f1 ( a lot of qa datasets use that basically)
I changed to BLEU/ROUGE when applicable.
I left the `generate_answer_and_explanations` to simply Others (because it technically requires parsing of the target...

PR #619, opened 2021-11-01
Sel qa PR and thoughts

Iwontbecreative:
I took a look at the current templates for `Sel_qa` and made some minor fixes. 

A major issue is that the original task is not template-friendly: The original task ranks up to ~15-20 passages, where multiple answers can be valid. 2 main issues: 1. the prompt is too long 2. not really a great way to check for multiple answers and their order. The current templates are reformulations and easier tasks (e.g: picking one of two passages), as well as predicting some metadata. I did not add original task prompts as I do not believe they would work without our framework.

Two other issues I noticed with the current prompts:
1. The dataset is a bit noisy, esp. when it comes to the metadata that one must predict. This is an issue for the templates "about-topic-vs-random", "is-he-talking-about",  E.g: 
Input: What is the topic of the question "When did the Nazis take over Poland?"? Is it sport or art?
Target: art
2. Some of the answers one has to rank are the empty string, which means the 1 vs 1 comparisons can be very easy sometimes. This is a problem for the templates: "which-answer-1st-vs-random" and "would-make-sense-qu-rand", "makes-sense-0".  E.g: 
Input: The next question was "How many viewers did "Family Guy" premier to?" Which is
the correct answer? 1: "" or 2: ""Family Guy" officially premiered after Fox's
broadcast of Super Bowl XXXIII on January 31, 1999, with "Death Has a Shadow".
Target: 2

First issue is not easily fixable aside from dropping those templates. Second issue is fixable by excluding empty strings from the possible distractors. The template code is already quite complex (see which-answer-1st-vs-random), so I thought this was not worth it for this task.

### DIFF
--- a/promptsource/templates/selqa/answer_selection_analysis/templates.yaml
+++ b/promptsource/templates/selqa/answer_selection_analysis/templates.yaml
@@ -2,29 +2,29 @@ dataset: selqa
 subset: answer_selection_analysis
 templates:
   39f5f57c-50b9-40b3-bb4f-3f0e4fec7776: !Template
-    answer_choices: null
+    answer_choices: No ||| Yes
     id: 39f5f57c-50b9-40b3-bb4f-3f0e4fec7776
     jinja: '{% set rand_index = range(0,10)|choice %} He asked me "{{ question }}"
       Is he talking about {{ ["MUSIC", "TV","TRAVEL","ART","SPORT","COUNTRY","MOVIES","HISTORICAL
       EVENTS","SCIENCE","FOOD"][rand_index]|lower}}? ||| {% if topic == rand_index
-      %}Yes{% else %}No{% endif %}'
+      %}{{answer_choices[1]}}{% else %}{{answer_choices[0]}}{% endif %}'
     metadata: !TemplateMetadata
-      choices_in_prompt: null
-      metrics: []
+      choices_in_prompt: false
+      metrics:
+      - Accuracy
       original_task: false
     name: is-he-talking-about
     reference: ''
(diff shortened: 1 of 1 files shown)


VictorSanh:
lgtm! thanks for working on this

1/ this is a broader discussion. let's separate this discussion from prompting datasets
2/ how frequent are these cases? i would tend to think that there is no need to introduce more noise than there is already and wrapping a condition `{% if smthg != "" and smtgh2 != "" %}` around is not much of a big deal

Iwontbecreative:
> 2/ how frequent are these cases? i would tend to think that there is no need to introduce more noise than there is already and wrapping a condition `{% if smthg != "" and smtgh2 != "" %}` around is not much of a big deal


@VictorSanh Sadly the code to do this is more complicated. You have to build a new list of non empty answer indexes to select the distractors from, which is not very Jinja-friendly. I did it for 2 out of 3 templates, see the updated commit.
The last one "which-answer-1st-vs-rando" is already over-complicated as is so I left it.

VictorSanh:
> I did it for 2 out of 3 templates, see the updated commit.

I was actually thinking more about wrapping the whole template so that is there aren't two real choices, then the template won't be applied to this particular example: `{if condition} template {% endif %}` If I understand correctly, it's not doing that currently in your last commit. am i missing something?

Iwontbecreative:
Oh, I see. My last commit actually fixes (2) for all templates but which-answer-1st-vs-rando . Your solution can be used for that last template. Does that work? Tests will still fail because it is not happy about the unused _ variable in the templates.

VictorSanh:
great that works, overwriting the test, i couldn't find a way to write that in another way
merging now

PR #621, opened 2021-11-02
Remove neural_code_search

JetRunner:
This PR proposes to remove `neural_code_search`.

The reasons are:
- Current templates (code/comment generation) fall into the category of "very open-ended generation".
- There is no original task that can be added. The original task is to search code from a corpus, which cannot be formulated into a prompting task.


VictorSanh:
yes looks good to me. To me the main motivation for removing it is that it is a non-natural language dataset

PR #623, opened 2021-11-03
Clean templates for scan/filler_num_x, x=1,2,3

abheesht17:
**Important Points:**
- [x] **The task is explicitly stated. However, I have reservations about the answer format. The answer is a string and has terms like "I_TURN_RIGHT", "I_TURN_LEFT", etc. which isn't exactly "natural language". Should we change this? Should we modify the input string to "...in SCAN format"?**
- [x] All "Original Task" checkbox entries have been double-checked. **Templates with names ending with `_with_hint` have not been marked as "Original Task.**
- [x] The Choices in Prompt template looks correct. **Now, for templates with names ending with `_with_hint`, should I tick the "Choices in Prompt" checkbox?**
- [x] I have **added BLEU and ROUGE scores** as metrics since this is a generation task.
- [x] I have added three templates ending with `_opposite`, which translates SCAN actions into NL commands. 

**Common Points:**
- [x] The dataset is appropriate. It does not have any harmful content.
- [x] The templates do not have any bugs. Tried with multiple examples.
- [x] The templates do not have any other issues like span indices (since it is not a QA task), negated answers (no real scope of negation here). The passages are short; there is no "long distance" issue either.
- [x] This is a generation task, and it does not have multiple correct answers.
- [x] There are 8 original task templates.
- [x] We don't need to define `answer_choices` for this task.
- [x] The output contains only the answer.
- [x] Key terms have been correctly escaped.
- [x] Templates have been named appropriately.

### DIFF
--- a/promptsource/templates/scan/filler_num1/templates.yaml
+++ b/promptsource/templates/scan/filler_num1/templates.yaml
@@ -4,35 +4,28 @@ templates:
   21cbc219-c6ec-46d8-8cfc-e039e0429746: !Template
     answer_choices: null
     id: 21cbc219-c6ec-46d8-8cfc-e039e0429746
-    jinja: 'Please translate correctly the following commands in natural language
-      into the corresponding SCAN actions.
-
-
-      {{ commands }}
-
-      |||
-
-      {{ actions }} '
+    jinja: "{% set scan_lst_of_actions = actions.split(' ') %}\n{% set lst_of_actions\
+      \ = [] %}\n{% for item in scan_lst_of_actions %}\n    {{ lst_of_actions.append(item.lower()[2:]\
+      \ | replace(\"_\", \" \")) | default(\"\", True) }}\n{% endfor %}\n\n{% set\
+      \ actions = lst_of_actions | join(\",\") %}\n\nPlease translate correctly the\
+      \ following commands in natural language into the corresponding sequence of\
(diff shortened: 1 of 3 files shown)


stephenbach (changes requested):
Thanks for working on this interesting dataset! Thanks also for laying out the important points. I think that since it's a funky dataset, we should think a little bit more about how to prompt it. Some thoughts:
1. I agree that the format should be natural language. I don't think saying "in SCAN format" is sufficient though. In a zero-shot evaluation, that wouldn't mean anything. Can we make the instructions more explicit, like make the output "Turn left" "Turn right" etc.?
2. I don't think whether there's a hint or not matters for original task. The exact output format is part of the prompting. Describing what that is I don't think should affect it.
3. Likewise, choices in prompt shouldn't be affected by the hint. That's only if it lists the possible outputs (not the pieces the output should be composed of).
4. Metric should either be accuracy (if you want the model to generate the exact sequence) or Other if the original paper parses the output and gives partial credit (say you make two right turns instead of three). But BLEU and ROUGE won't capture this.
5. I think the opposite prompts are great, just same issue about making all pieces natural language.

stephenbach (approved):
Great, thanks!

I just added a space between all the directions and deleted the prompts that still use the non-natural language instructions. We have plenty without them.

PR #627, opened 2021-11-03
Clean discofuse-x datasets, x=sports, wikipedia

abheesht17:
- `answer_choices` field added for certain templates.
- For certain templates which require generation, I have added BLEU and ROUGE scores as metrics. For the Original Task templates, I have added "Other"; the paper has "Exact Match" and "SARI" as metrics. For classification tasks, I have added "Accuracy" and "AUC". I wanted to add "F1" as well but since it was named "CoQA F1", I did not. Should I?
- Changed names of a few templates.
- Ticked the "Original Task" checkbox for appropriate templates, same for "Choices in Prompt".


VictorSanh:
thanks for working on this @abheesht17 
merging now, the test are still failing because the dataset is too big

PR #628, opened 2021-11-03
cleaned ecthr_cases dataset

teelinsan:
Cleaned ecthr_cases dataset


VictorSanh (approved):
thanks for working on this @teelinsan 
to be sure, when it was easy to do, I moved the instruction to the top in case the input was truncated

PR #630, opened 2021-11-03
clean kelm

yongzx:
I've completed the following:
- [X] added metrics (for `sentence_to_kb`, which is not an original task, I choose BLEU and ROUGE. My reasoning is that it is reasonable to compare n-gram words between the generated KB and the correct KB.)
- [X] add the description of KB triples in the prompt.
- [X] reformat the template names


VictorSanh:
thanks for working on this @yongzx !

Minor wording modification
I removed pearson correlation cause in the paper they needed some human annotations to compute this correlation (you need floats to compute the pearson correlation, so it's not out of the box)

PR #592, opened 2021-10-29
clean movie rationales

zaidalyafeai:

PR #631, opened 2021-11-03
clean limit

yongzx:
I've done the following for the cleaning:
- [X] add metrics
- [X] add more templates so we have more than 5 templates.
- [X] add answer choices for yes/no prompts

### Discussion about metrics
- The original tasks (`find_entities_*`) are sequence tagging tasks. The prompt output here is a list of entities in motion. Given this [discussion about Accuracy](https://github.com/bigscience-workshop/promptsource/pull/592/files#diff-65f41d4038388a2817ae686969f206cb05c4c61a9efeebf91221f533ed895328R89), I believe evaluating output requires parsing. Therefore, I choose Others as the metric. The correct metric would be Precision, Recall, and F1.
- However, I use Accuracy as a metrics for the two prompts: first_moving_entity and last_moving_entity. My reasoning is that exact string match can be applied here.
- For the regression task (`count_entities*`), I use Other metrics because the appropriate metrics should be the mean-squared error. Should I use accuracy instead?

### DIFF
--- a/promptsource/templates/limit/templates.yaml
+++ b/promptsource/templates/limit/templates.yaml
@@ -6,7 +6,8 @@ templates:
     jinja: '{{sentence}}
 
 
-      What is the last entity in motion mentioned in the sentence if any?
+      What is the last entity in motion mentioned in the sentence if any? Otherwise,
+      respond with "No entity in motion".
 
       |||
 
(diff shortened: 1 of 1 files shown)


VictorSanh:
thanks for working on this @yongzx , agree with all the points you mentioned 
merging now

PR #632, opened 2021-11-04
clean mocha

yongzx:
Completed the following tasks:
- [X] include metrics
- [X] move if-else statement to target output for classify_similarity prompt (so the prompt is not empty)
- [X] use answer_choices
- [X] make instructions more explicit (such as indicating what similarity score of 1.0 means)


stephenbach (approved):
Thanks! I took out the multiple questions aspect because they seem rare and unnecessarily complicated. Also took out the candidates from the question answering task. I don't see how they help.

PR #633, opened 2021-11-04
clean wiki_split

yongzx:
Completed the following:
- [X] add metrics
- [X] rename templates to become more meaningful
- [X] include up to 5 templates for the original task.
- [X] remove unnecessary quotation marks.

### Discussion
The original task requires outputting two sentences. I included the instruction of separating them with a newline character in the input prompt. Is this the correct practice?


stephenbach (approved):
Thanks! I took out the specific instruction to output a newline, since T5 treats new lines as whitespace.

PR #634, opened 2021-11-04
clean winograd_split

yongzx:
Did the following changes to the split `wsc285`:
- [X] reuse prompts from subtask wsc273 to increase the number of templates for original tasks to >5.
- [X] remove a template that jointly identifies pronoun and its reference and the `identify_pronoun` prompt because they can be problematic if the text contains multiple pronouns

### Discussion
It seems that subtask wsc273 has been cleaned, so I reuse the prompts from it.


stephenbach (approved):
Great! I took out the extra template just to make the two subsets match. I'm not sure it was a clear task anyway. Thanks!

PR #635, opened 2021-11-04
clean ambig_qa/light

yongzx:
I have completed the following:
- [X] increase templates for original tasks 
- [X] rename templates
- [X] include metrics and answer choices

### Discussion and Reasoning:
- I use Other metrics for the original task of answer prediction because the evaluation setup requires parsing the gold target output into a set of answers, and check if the model's output matches any of the gold answer.
- I use BLEU and Edit distance for the original task of question disambiguation, but I am not too sure if it is the right one. The evaluation setup wants to find the maximum BLEU and minimum Edit distance (as it is evaluated on a set of strings instead of a single string). See equation for correctness score in the original paper [here](https://arxiv.org/pdf/2004.10645.pdf).

### DIFF
--- a/promptsource/templates/ambig_qa/light/templates.yaml
+++ b/promptsource/templates/ambig_qa/light/templates.yaml
@@ -21,7 +21,7 @@ templates:
       metrics:
       - BLEU
       - Edit Distance
-      original_task: true
+      original_task: false
     name: is_question_ambiguous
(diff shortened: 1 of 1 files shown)


stephenbach (approved):
Thanks! Just changed some of the metadata. I think only the question answering with all answers is the original task.

PR #637, opened 2021-11-04
Clean up `nlu_evaluation_data`

JetRunner:
In this PR, `nlu_evaluation_data` is cleaned.

- Add tags for original task and choices
- Fix some formats in the templates
- Use `answer_choices` for choices
- Add metric


stephenbach (approved):
Thanks! I redid a lot of these because the answer choices did not match the targets.

PR #639, opened 2021-11-05
Clean up `mdd`

Edward-Sun:
This PR cleans the three subtasks from the `math_dataset` dataset. Added more original prompts to the task.

Specifically, for the `task1_qa` subtask, I currently set the metrics for question-generation-style non-original tasks to BLEU and ROUGE. Please let me know if there are other better metrics.
Also, some movie entities and actor entities in the `train` split of `task1_qa` are not well-capitalized. I currently cannot find a jinja-based way to solve this problem.


stephenbach (approved):
Awesome, thanks! Just some minor tweaks.

PR #641, opened 2021-11-09
clean narrativeqa

zaidalyafeai:


stephenbach:
Do you think we should just remove the prompts that depend on the full text? They're entire books. While they might be useful to some prompting research someday, it seems like it's more likely to cause problems because the books are truncated by default?

zaidalyafeai:
If I remember correctly, the main contribution of the dataset is the longer context compared to other datasets. What do you mean by truncated by default?

stephenbach:
You're right, it's the point of the dataset. In both promptsource and when creating seqio data, these will be truncated to fix within the context window. So the questions that require reading the full text are unsolvable in this format. I'm wondering what's more likely: someone actually wants to use prompting for summarizing entire books, or someone will accidentally train/evaluate on the truncated prompts?

zaidalyafeai:
If someone chooses to evaluate on this dataset, then their purpose is mainly to evaluate prompting on long context. I think it does not make sense to assume a fixed context window because models can get wider. Was the choice of context window dependent on T5? Are we only targeting similar models?

stephenbach (approved):
Sounds good. Not targeting any particular model, just wondering. Merging now.

PR #644, opened 2021-11-15
not sure about the metrics

srulikbd:
I didn't find specific one in the original paper 
http://www.lrec-conf.org/proceedings/lrec2014/pdf/363_Paper.pdf


stephenbach:
This needed to be redone (#648). Please be sure to set the answer choices.

Issue #567, opened 2021-10-27
Community datasets refer to the wrong source files 


zaidalyafeai:
Replace 

[https://github.com/huggingface/datasets/blob/master/datasets/Zaid/quac_expanded/Zaid/quac_expanded.py](https://github.com/huggingface/datasets/blob/master/datasets/Zaid/quac_expanded/Zaid/quac_expanded.py)
by 
[https://huggingface.co/datasets/Zaid/quac_expanded/blob/main/quac_expanded.py](https://huggingface.co/datasets/Zaid/quac_expanded/blob/main/quac_expanded.py)

Affected datasets `Zaid/quac_expanded`, `Zaid/coqa_expanded` and `craffel/openai_lambada`.

zaidalyafeai:
https://github.com/bigscience-workshop/promptsource/pull/646

PR #646, opened 2021-11-15
fix broken source links for community datasets

zaidalyafeai:
Fix this issue https://github.com/bigscience-workshop/promptsource/issues/567.

### DIFF
--- a/promptsource/app.py
+++ b/promptsource/app.py
@@ -329,7 +329,10 @@ else:
         split_dataset_key = dataset_key.split("/")
         possible_user = split_dataset_key[0]
         if len(split_dataset_key) > 1 and possible_user in INCLUDED_USERS:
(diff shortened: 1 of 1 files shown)


stephenbach (changes requested):
Thanks! I guess for readability and also avoiding some weird cases where usernames match dataset names, I'd rather it be something like
```python
split_dataset_key = dataset_key.split("/")
possible_user = split_dataset_key[0]
if len(split_dataset_key) > 1 and possible_user in INCLUDED_USERS:
    source_link = "https://huggingface.co/datasets/%s/blob/main/%s.py" % (
                dataset_key,
                split_dataset_key[-1],
            )
else:
    source_link = "https://github.com/huggingface/datasets/blob/master/datasets/%s/%s.py" % (
        dataset_key,
        dataset_key,
    )
```

zaidalyafeai:
Thanks, fixed.

stephenbach (approved):
Great, thanks!

PR #647, opened 2021-11-16
templates names changes

srulikbd:
there are 4 different metrics. they are explained in the original repository:
https://github.com/taoyds/spider/tree/master/evaluation_examples

Different Evaluation Metrics
Our evaluation script contains:

SQL Hardness Criteria
Exact Matching without Values
Partial Matching without Values
Execution Accuracy with Value Selection


stephenbach:
Ah, sorry about this, but this is one that should be removed, because it has non-natural language. We had said this in slack and meetings, but not explicitly in instructions. (I've fixed that!)

PR #650, opened 2021-11-20
clean_commonsense_qa

manandey:


stephenbach (changes requested):
Really nice! Two minor things:
- Can we change the metrics for `answer_to_question` to ROUGE and BLEU? They're not perfect, but that's what we've been using for question generation.
- The `select_incorrect_answers` has a bug: the values in answer choices do not match the possible targets. (Answer choices is one answer each, but the targets are 4/5 answers.) Rather than fix it, you could just delete this prompt. Negated tasks like that are problematic anyway. (See cleaning doc for more.)

Thanks!

manandey:
Thanks a lot for reviewing the PR @stephenbach! Fixed the issues.

stephenbach (approved):
Great, thanks!

PR #651, opened 2021-11-20
clean_xquad

manandey:


stephenbach (changes requested):
The question answering ones have a couple issues:
- answer_choices should not be just the correct answer. I don't believe answer choices should be set for these.
- The metric for them should be Squad.

manandey:
Thanks, @stephenbach! Hope everything is fixed now.

stephenbach (approved):
Perfect!

PR #652, opened 2021-11-20
Clean climate fever

shanyas10:
There are kind of redundant templates: each statement (5 in total) have been used to create prompts that are essentially same in structure. Wasn't sure if we should let them be as is


stephenbach:
Thanks! Yes, I think it would be better to have 5 more varied prompts than so many with slight variations. Also, the phrasings of many of the questions are a little confusing. I think instead of asking whether the statements refute or support "on climate change" it should ask whether they support the claim?

shanyas10:
Thanks for revieweing @stephenbach . I've changed the structure for a few prompts and deleted the redundant ones.

stephenbach (approved):
Thanks! Just pushed a commit to make the prompts for grammatical, and also fix a few missing instructions about "not enough info".

PR #653, opened 2021-11-20
clean_xquad_r

manandey:


stephenbach:
Same dataset as `xquad`, so can just copy over templates when #651 merged.

manandey:
Thanks, @stephenbach! Since #651 is merged, copied the same here.

stephenbach (approved):
Perfect, thanks!

PR #654, opened 2021-11-20
clean_quora

manandey:


stephenbach (approved):
Awesome, thanks! Just made very minor tweaks to capitalization and spacing based on my preferences, but they weren't really necessary to change. :)

PR #655, opened 2021-11-20
clean_poem_sentiment

manandey:


stephenbach (approved):
Awesome! Mostly just some minor tweaks. I also took out a prompt that was asking for how the reader feels, which I don't think is necessarily the same thing as the sentiment of the statement.

PR #658, opened 2021-11-20
clean_banking77

shanyas10:
Declared the intents as the answer_choices for each template. Is there a way to declare them once, globally?


stephenbach:
Unfortunately no way to declare them globally since not all prompts share the same answer choices. Also, these need to be expressed more as natural language. (The labels are not words because of the underscores.) See #637 for a similar dataset.

shanyas10:
> Unfortunately no way to declare them globally since not all prompts share the same answer choices. Also, these need to be expressed more as natural language. (The labels are not words because of the underscores.) See #637 for a similar dataset.

I've replace underscores with spaces wherever they're used in prompt or target text. Will that not work? I can change the choices themselves if that's the case.

stephenbach:
Same thing as in #637. Changing the choices in the target creates a bug. The choices always need to match the targets exactly.

shanyas10:
ah ok, got it!

shanyas10:
@stephenbach I changed the choices as suggested. Also corrected the `choices_in_prompt` field for some prompts

stephenbach (approved):
Thanks, just a few minor fixes. Also changed the number indices in the choices to dashes, because the style guideline is to not use letters or numbers as indices if the target is the text itself (as opposed to the index).

PR #660, opened 2021-11-21
Clean freebase_qa, craigslist_bargains, health_fact, hyperpartisan_news_detection

jason-fries:
For this task I:
- Added accuracy metrics to all prompts
- Assigned "original task" to the minimal QA prompt template (vs. the ones that add additional contextual information) 
Each template makes some style changes (e.g., capitalizing answers in-prompt) that I don't if we should leave as-is or remove for simpler templates.


stephenbach:
Thanks! Let's make them all original task. It's more about the broad skill that's required than exactly what info is provided. Prompts that have additional context should still be considered original task.

jason-fries:
Updated all the prompts that reflect the original task.

jason-fries:
This PR now contains cleaned up prompts for:
- freebase_qa
- craigslist_bargins
- health_fact
- hyperpartisan_news_detection (byarticle)
- hyperpartisan_news_detection (bypublisher)

stephenbach (approved):
Thank you!

PR #663, opened 2021-11-25
add MeDAL dataset templates

WojciechKusa:
Adding templates for [MeDAL](https://huggingface.co/datasets/medal) dataset  (biomedical group).


stephenbach:
Thanks so much! A couple of things related to answers to get started:
- Can you use the answer choices field where possible? Please see the [contributing guide](https://github.com/bigscience-workshop/promptsource/blob/main/CONTRIBUTING.md#options) if you need a primer.
- Can you use natural language everywhere, e.g., `{{"Yes"}} | {{"No"}}` -> `{{"Yes"}} or {{"No"}}`?
- Also looks like at least one prompt isn't rendering correctly based on the tests.

Thanks!

WojciechKusa:
> * Can you use the answer choices field where possible? Please see the [contributing guide](https://github.com/bigscience-workshop/promptsource/blob/main/CONTRIBUTING.md#options) if you need a primer.
> * Can you use natural language everywhere, e.g., `{{"Yes"}} | {{"No"}}` -> `{{"Yes"}} or {{"No"}}`?

Sure I will change that!

> * Also looks like at least one prompt isn't rendering correctly based on the tests

At least for one [test](https://github.com/bigscience-workshop/promptsource/runs/4335999311?check_suite_focus=true), it looks like the issue is just with the dataset size/disk space.

WojciechKusa:
@stephenbach could you take a look now?

stephenbach:
Thanks @WojciechKusa! Looking good!
A few thoughts
- What metric were you thinking for "Other?" Would accuracy not be appropriate?
- I'm a little skeptical about the yes/no questions. It seems like the no candidates are so random that it's not useful as a training/test task. Perhaps they would be better replaced with other variations of the original task. What do you think?

Sorry, some of these comments would have been more useful earlier, but I just finally downloaded this big dataset.

WojciechKusa:
Hi @stephenbach, sorry for the late reply. 
I updated the metric to accuracy; you are right.

I also added two more prompts that reverse the original task and correspond to how the original MeDAL dataset was created (reverse substitution).



>  - I'm a little skeptical about the yes/no questions. It seems like the no candidates are so random that it's not useful as a training/test task. Perhaps they would be better replaced with other variations of the original task. What do you think?

I agree, the negative candidates might be too random. Ideally, we would need to add different examples having the same abbreviation, but this information is not available in this dataset. I have no idea that this would be possible from the Jinja template or even without altering the dataset. So I can remove this three yes/no templates and keep only the four original ones (unless you believe there would be gain from paraphrasing the original task more). What do you think?

stephenbach:
No worries, thanks for following up! Yes, agree that just removing the yes/no templates is best. Maybe to get a fifth original-task template you could situate the prompt in a scenario. Something like, "I am a doctor and I need to..."

WojciechKusa:
I removed yes/no templates and added two more templates based on doctor/layperson scenario.

stephenbach:
Looks great! Finally got around to double checking these. And `test_templates.py` passes locally. This should be fixed after merging with main. Thanks!

PR #665, opened 2021-11-25
Clean e2e nlg

shanyas10:
Please review the metrics for *_qa style prompts. Was not sure what would be an appropriate metric for them


stephenbach (approved):
Thanks! I think accuracy is fine for these qa prompts. Interesting to see that T0++ already gets a bunch right.

Also changed "dinery" to "restaurant." Wikitionary says it's a rare word: https://en.wiktionary.org/wiki/dinery

PR #666, opened 2021-11-25
tab_fact_fix-name, metric fixing

srulikbd:


stephenbach (approved):
Thanks!

(I deleted a couple of prompts that violated the guidelines like negated questions and non natural language. Also tweaked the wording a bit to be clearer in my opinion.)

PR #667, opened 2021-11-26
sst-fix

srulikbd:
-adding prompts
-edit the prompt of parse trees (not sure of any metric for it..)


stephenbach (approved):
Thanks! I replaced the parsing prompt because the targets are not expressed in natural language.

PR #675, opened 2021-12-01
Clean up cord19

HanNight:
This PR cleans up `cord19` by:
- Rename the templates
- Add metric


VictorSanh:
thanks for working on this @HanNight !
I fixed a few targets (only the changing part in the target, no template)
merging now

PR #677, opened 2021-12-02
Add templates for remaining scan subsets

abheesht17:
addprim_jump, addprim_turn_left, filler_num0, length, simple, template_around_right, template_jump_around_right, template_opposite_right, template_right


stephenbach:
Ah sorry, I messed your last PR up because the copy script doesn't delete templates that are in the target but not the source. So the only correct subset is scan/filler_num1. Can you delete all the other ones (just delete their sub directories in the repo), _then_ run the copy from filler_num1 to everything else. Sorry about that.

stephenbach (approved):
Thank you!!

PR #681, opened 2021-12-04
clean codah

yongzx:
- [X] Add metrics
- [X] Ensure 5+ original tasks
- [X] Rewrite some prompts so all the original task prompts are about choosing from answer choices.
- [X] populate all the prompts across different subsets (from codah subset to fold* subsets)


VictorSanh:
thanks for working on this @yongzx !
merging now

PR #699, opened 2021-12-20
fix answer_choices (1,2,3,4 vs a,b,c,d) in ai2 arc

VictorSanh:
In somre rare cases, the indexing used is 1,2,3,4 instead of A,b,c,d.
This fixes the discrepancy so that the target is always in `answer_choices`


stephenbach (approved):
LGTM! Merging now.

PR #700, opened 2021-12-20
Updated the existing prompts for onestop_english and

dragomirradev:
scientific_papers/pubmed

Add new prompts for both of these datasets.


dragomirradev:
reopening

VictorSanh:
thanks for working on this @dragomirradev !
merging now

PR #701, opened 2021-12-21
add sports prompt in agnews

infinite-Joy:
Since this is the first PR w.r.t this task, please review if the PR makes sense and is correct.

### DIFF
--- a/promptsource/templates/ag_news/templates.yaml
+++ b/promptsource/templates/ag_news/templates.yaml
@@ -12,6 +12,22 @@ templates:
       original_task: true
     name: classify_question_first
     reference: ''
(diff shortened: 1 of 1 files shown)


VictorSanh:
Hello @infinite-Joy 
Thanks for your contribution!

A prompt is a small function that transforms instances into a text to text format that makes the query (or task) explicit in natural language. Prompts should be applicable to all examples in the dataset/subset, but your prompt look specific to one instance... Could you look at a the existing prompts for ag news first? you can scroll through the resulting instances in the dataset viewer mode

stephenbach:
I'm going to close this because it's stale. Feel free to reopen if needed!

PR #704, opened 2021-12-22
Updated the existing prompts

tangxiangru:
Updated the existing prompts for billsum, pubmed_qa, and scientific_papers(arxiv).


stephenbach:
Thanks @tangxiangru. Can you look into why the tests are failing? It looks like at least one template doesn't parse.

VictorSanh:
It would be also worth adding the metrics in the same pass!

VictorSanh:
thanks for working on this @tangxiangru!
i pushed a few minor changes
the test `show_new_templates` is failing because the combination of 3 datasets is fairly big so there is not enough space on the CI device, but I ran the test locally and they work fine
merging now.

PR #707, opened 2021-12-30
finish qed

tae898:
* Add more prompts
* Fix typos
* Add metrics
* tick boxes
* Add sanity checks


stephenbach (approved):
Great, thanks! I deleted a couple prompts:
- The "does it contain the answer?" one, because it seems the answer is always or almost always yes.
- The "how many sentences" one because it requires counting.

Otherwise a few minor tweaks.

tae898:
@stephenbach thanks as always!

PR #711, opened 2022-01-05
Use requirements.txt inside setup.py

Guitaricet:
Enhancement proposal.

It would be great if `pip install git+https://github.com/bigscience-workshop/promptsource` would install the package dependencies. I faced this inconvenience when installing T0 repository in a clean environment. T0 only specifies `promptsource` as a dependency and because the `requirements.txt` in promptsource are not propagated into its `setup.py`, the package is installed without `sentencepiese` and `jinja2`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,9 @@ from setuptools import setup, find_packages
 with open('README.md') as readme_file:
     readme = readme_file.read()
 
+with open('requirements.txt') as requirements_file:
+    requirements = requirements_file.read().split('\n')
+
 setup(
     name='promptsource',
     version='0.1.0',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -3,13 +3,17 @@
 with open('README.md') as readme_file:
     readme = readme_file.read()
 
+with open('requirements.txt') as requirements_file:
+    requirements = requirements_file.read().split('\n')
```
VictorSanh, line 7: i tried it and it works as if, but wanted to double check with you that we don't need to strip out the commented lines in `requirements.txt`
Currently, the list is:
```python
['black', 'datasets>=1.7.0', 'flake8', 'isort==5.8.0', 'pytest', 'pyyaml>=5', 'streamlit==0.82', 'jinja2', 'plotly', 'requests', 'pandas', '##############################################################', '# Dependencies in this section are added for specific datasets', '##############################################################', 'py7zr', '##############################################################', '# End of dataset-specific dependencies', '##############################################################', '']
```

VictorSanh (approved):
thanks @Guitaricet !

thomasw21:
Is there any reasons to still use `requirements.txt`? Like couldn't one copy everything into setup.py?

PR #715, opened 2022-01-18
add 5 prompts to lama dataset

KhalidAlt:


VictorSanh:
thanks for contributing these @KhalidAlt 
one metric was missing. added "other" for the negations too as there is not a lot of diversity in the negations so BLEU likely won't capture much
merging now

PR #717, opened 2022-01-21
Some fixes that slipped through

VictorSanh:
Missing metrics and original_task metadata


VictorSanh:
`show_new_templates` CI fail because jigsaw is dataset that require manual download, wiki_hop currently fails (dataset download error).
I am bypassing the tests and merging.

Issue #694, opened 2021-12-13
Jinja sometimes splits in the `Prompted dataset viewer`


VictorSanh:
Figure out why this happens and open a PR to fix it

Some examples:
<img width="969" alt="Capture d’écran 2021-12-07 à 4 41 30 PM" src="https://user-images.githubusercontent.com/16107619/145869838-b0f7408d-45b3-4720-9b4a-85ca45269dbe.png">
<img width="1425" alt="Capture d’écran 2021-12-08 à 5 22 16 PM" src="https://user-images.githubusercontent.com/16107619/145869846-5d8d1716-406e-4455-9a42-e6e8393ffbf6.png">

zaidalyafeai:
Seems to be caused by multiple new lines.

VictorSanh:
yes, that sounds like the right diagnostic
would you like to propose a fix for that @zaidalyafeai ?

zaidalyafeai:
Using `re.sub(r'\n+', '\n', t)` inside `show_jinja` function seems to resolve the issue. Not sure why `{% for candidate in
candidate_answers -%}` is not in one line, though. Did we have this problem before?

![image](https://user-images.githubusercontent.com/15667714/146334849-397f1889-1063-49ec-92b4-c97aca04a656.png)

zaidalyafeai:
@VictorSanh, If that is ok, I can submit a PR today. Won't be able to work on it later, because I am traveling.

VictorSanh:
I'd like the rendering to be as faithful as the template (i.e. if there are 3 \n then we should display 3 \n). Would doing a `join` before calling https://github.com/bigscience-workshop/promptsource/blob/415a49333f792a52b682ece4ef107082b015a6ea/promptsource/app.py#L91 solve the issue?

PR #719, opened 2022-01-21
fix double linebreak issue in jinja rendering #694 

VictorSanh:
Addresses #694 

The issue was coming from `st.write` which does not handle well double linebreaks. I explicitly replaced `\n\n` with their HTML equivalent `<br/><br/>` to address that problem.

**Before:**
Capture d’écran 2022-01-21 à 5 55 51 PM
Capture d’écran 2022-01-21 à 5 56 21 PM

**Now:**
Capture d’écran 2022-01-21 à 5 56 02 PM
Capture d’écran 2022-01-21 à 5 56 11 PM

### DIFF
--- a/promptsource/app.py
+++ b/promptsource/app.py
@@ -82,12 +82,19 @@ st.markdown(
     "<style>" + HtmlFormatter(style="friendly").get_style_defs(".highlight") + "</style>", unsafe_allow_html=True
 )
 
-WIDTH = 80
+WIDTH = 140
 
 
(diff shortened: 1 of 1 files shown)


zaidalyafeai:
my internet is bad so can't verify, won't it be better if we replace every new line with `</b>` ?

VictorSanh:
> my internet is bad so can't verify, won't it be better if we replace every new line with `</b>` ?

indeed I can do that too. even more simple

PR #723, opened 2022-01-28
add templates for riddle_sense dataset

JonnodT:
Use almost identical templates as commonsense_qa because the two datasets are very similar.

@yuchenlin


VictorSanh:
thanks for contributing these @JonnodT !

I ran the tests locally and they passed (I pushed an update on the tests earlier cause they were failing automatically)

I moved the conditional statement on `answerKey` to the target

yuchenlin:
@VictorSanh  Thank you!

Issue #730, opened 2022-03-23
module 'datasets.load' has no attribute 'prepare_module' - fresh install


chris-aeviator:
I'm trying to bring up promptsource from a fresh environment. When acessing the `Dataset Viewer` or `Sourcing` section, streamlit shows this error.

```
AttributeError: module 'datasets.load' has no attribute 'prepare_module'

Streamlit encountered an error while caching the body of get_dataset_confs(). This is likely due to a bug in /promptsource/promptsource/utils.py near line 80:

module_path = datasets.load.prepare_module(path, dataset=True)
# Get dataset builder class from the processing script
builder_cls = datasets.load.import_main_class(module_path[0], dataset=True)

Please modify the code above to address this.

If you think this is actually a Streamlit bug, you may file a bug report here.
Traceback:

File "/promptsource/promptsource/app.py", line 264, in <module>
    configs = get_dataset_confs(dataset_key)
```

## Steps to reproduce
```
git clone https://github.com/bigscience-workshop/promptsource /opt/software-test/promptsource

# clean python 3.7 environment via docker
docker run -it --name python37-promptsource -p 8501:8501 -v /opt/software-test/promptsource:/promptsource python:3.7-bullseye /bin/bash

# setup and run promptsource inside the container
cd /promptsource/
pip install -e .
streamlit run promptsource/app.py
```

### Edit
seems `prepare_module` has been deprecated and removed 21 days ago - all new installs will likely see this. https://github.com/huggingface/datasets/pull/3803 

>         `prepare_module` was deprecated in version 1.13 and will be removed in the next major version.
 >       For backward compatibility, please use :func:`dataset_module_factory` or :func:`metric_module_factory` instead

stephenbach:
Thanks for raising this! Looking into it now.

VictorSanh:
adressed with PR #735 :)

PR #732, opened 2022-03-27
Pin datasets to 1.17.0 fixes module_helper issue on fresh install.

jordiclive:
#730 , Pinning datasets to 1.17.0 fixes module_helper issue on fresh install.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ with open("README.md", "r", encoding="utf-8") as readme_file:
 
 requirements = [
     "black<=21.12b0",
-    "datasets>=1.7.0",
+    "datasets==1.17.0",
     "flake8",
     "isort==5.8.0",
     "pytest",


stephenbach:
Thanks, I'm trying to see if we can get promptsource to run with datasets 2.0, but at the very least should we just say datasets<2 rather than pinning on 1.17.0? Any particular reason to not go higher?

nihalnayak:
I verified that `datasets==1.18.4` works fine for me.

VictorSanh:
@lhoestq suggested more simply on slack that we do:
```python
from datasets import load_dataset_builder

builder_instance = load_dataset_builder(dataset_path)
```
which will be compatible for datasets 1.x and 2.x.

If anyone wants to take a stab, feel free, otherwise I might come back to that end of the week :)

Issue #588, opened 2021-10-29
AttributeError: 'NoneType' object has no attribute 'session_id'


manandey:
Hi,

Is anyone facing this issue while running `streamlit run promptsource/app.py` ?

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\I355109\Anaconda3\envs\python37\lib\multiprocessing\spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "C:\Users\I355109\Anaconda3\envs\python37\lib\multiprocessing\spawn.py", line 114, in _main
    prepare(preparation_data)
  File "C:\Users\I355109\Anaconda3\envs\python37\lib\multiprocessing\spawn.py", line 225, in prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "C:\Users\I355109\Anaconda3\envs\python37\lib\multiprocessing\spawn.py", line 277, in _fixup_main_from_path
    run_name="__mp_main__")
  File "C:\Users\I355109\Anaconda3\envs\python37\lib\runpy.py", line 263, in run_path
    pkg_name=pkg_name, script_name=fname)
  File "C:\Users\I355109\Anaconda3\envs\python37\lib\runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "C:\Users\I355109\Anaconda3\envs\python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\I355109\promptsource\promptsource\app.py", line 59, in <module>
    state = _get_state()
  File "c:\users\i355109\promptsource\promptsource\session.py", line 84, in _get_state
    session = _get_session()
  File "c:\users\i355109\promptsource\promptsource\session.py", line 74, in _get_session
    session_id = get_report_ctx().session_id
AttributeError: 'NoneType' object has no attribute 'session_id'
```

srulikbd:
yeah I run again into the same issue.
windows 11, python 3.7 
I tried the updated promptsource package..

srulikbd:
I tried installing it on ubuntu 20.04 (WSL) but encountered the problem of : "streamlit: command not found"

MagedSaeed:
@srulikbd 

> I tried installing it on ubuntu 20.04 (WSL) but encountered the problem of : "streamlit: command not found"

[...]

Issue #584, opened 2021-10-28
❓ Support for python 3.8


pietrolesci:
Hi there,

Is there any plan to support python 3.8?

Best,
Pietro

VictorSanh:
see #733 :)

stephenbach:
Closed via #733.

Issue #572, opened 2021-10-27
Issue with Multiprocessing on MacOS


yongzx:
I run into the issue of `Multiprocessing causes Python to crash and gives an error may have been in progress in another thread when fork() was called`, which I fix (by referring to the [thread](https://stackoverflow.com/questions/50168647/multiprocessing-causes-python-to-crash-and-gives-an-error-may-have-been-in-progr)) by adding `export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` to my .zshrc

However, now I encounter another error of `The process has forked and you cannot use this CoreFoundation functionality safely`, of which one of the [solutions](https://stackoverflow.com/questions/30669659/multiproccesing-and-error-the-process-has-forked-and-you-cannot-use-this-corefou) requires modifying the script app.py.

Wondering if there's a quick fix to this issue, or should I modify the app.py and do a PR?

stephenbach:
I started hitting this issue after upgrading OSX to 12.3. When I include the suggested fix, it works in that I can navigate off the helicopter view and use the rest of promptsoruce normally, but the helicopter view itself fails with this error
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/bach/anaconda3/envs/promptsource/lib/python3.7/multiprocessing/spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "/Users/bach/anaconda3/envs/promptsource/lib/python3.7/multiprocessing/spawn.py", line 114, in _main
    prepare(preparation_data)
  File "/Users/bach/anaconda3/envs/promptsource/lib/python3.7/multiprocessing/spawn.py", line 225, in prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "/Users/bach/anaconda3/envs/promptsource/lib/python3.7/multiprocessing/spawn.py", line 277, in _fixup_main_from_path
    run_name="__mp_main__")
  File "/Users/bach/anaconda3/envs/promptsource/lib/python3.7/runpy.py", line 263, in run_path
    pkg_name=pkg_name, script_name=fname)
  File "/Users/bach/anaconda3/envs/promptsource/lib/python3.7/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/Users/bach/anaconda3/envs/promptsource/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
[...]

Issue #416, opened 2021-09-13
Interface doesn't load correctly


srulikbd:
I'm running on windows 10 with python 3.9
the web interface used to worked until I tried cloning the repo again..
this is the error:

```streamlit run promptsource/pr
omptsource.py

  You can now view your Streamlit app in your browser.

  Local URL: http://localhost:8501
  Network URL: http://10.0.0.101:8501

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "c:\users\user\appdata\local\programs\python\python39\lib\multiprocessing\spawn.py", line 116, i
n spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "c:\users\user\appdata\local\programs\python\python39\lib\multiprocessing\spawn.py", line 125, i
n _main
    prepare(preparation_data)
  File "c:\users\user\appdata\local\programs\python\python39\lib\multiprocessing\spawn.py", line 236, i
n prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "c:\users\user\appdata\local\programs\python\python39\lib\multiprocessing\spawn.py", line 287, i
n _fixup_main_from_path
    main_content = runpy.run_path(main_path,
  File "c:\users\user\appdata\local\programs\python\python39\lib\runpy.py", line 268, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "c:\users\user\appdata\local\programs\python\python39\lib\runpy.py", line 97, in _run_module_cod
e
    _run_code(code, mod_globals, init_globals,
  File "c:\users\user\appdata\local\programs\python\python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\User\Desktop\Srulik\Work\HuggingFace\BigScience\promptsource\promptsource\promptsource
.py", line 34, in <module>
    state = _get_state()
  File "C:\Users\User\Desktop\Srulik\Work\HuggingFace\BigScience\promptsource\promptsource\session.py",
 line 84, in _get_state
    session = _get_session()
  File "C:\Users\User\Desktop\Srulik\Work\HuggingFace\BigScience\promptsource\promptsource\session.py",
 line 74, in _get_session
    session_id = get_report_ctx().session_id
AttributeError: 'NoneType' object has no attribute 'session_id'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "c:\users\user\appdata\local\programs\python\python39\lib\multiprocessing\spawn.py", line 116, i
[...]

PR #733, opened 2022-03-27
Fix all the things by fixing multiprocessing

stephenbach:
Mutliprocessing in the helicopter view is the source of a lot of inconsistent bugs across OSs and Python versions. This PR fixes promptsource when using the spawn mode (default in python 3.8+ and required to run smoothly on latest OS X). It also changes it to the use spawn always so that hopefully there is more consistent behavior.

**Before merging, we should test on a range of OSs and Python versions.**

The main changes are actually pretty small, but the diff is huge because most of app.py had to be wrapped in a function. (This is so that this code is not executed in each subprocess, which was our problem when spawning.) The main changes besides wrapping the code is
- pass `all_infos` to `get_infos` as an argument
- Force start mode to spawn

### DIFF
--- a/promptsource/app.py
+++ b/promptsource/app.py
@@ -1,4 +1,6 @@
 import argparse
+import functools
+import multiprocessing
 import textwrap
 from multiprocessing import Manager, Pool
 
@@ -21,6 +23,22 @@ from promptsource.utils import (
     renameDatasetColumn,
     render_features,
 )
+
+
+# Python 3.8 switched the default start method from fork to spawn. OS X also has
+# some issues related to fork, eee, e.g., https://github.com/bigscience-workshop/promptsource/issues/572
+# so we make sure we always use spawn for consistency
+multiprocessing.set_start_method("spawn", force=True)
+
+
+def get_infos(all_infos, d_name):
+    """
+    Wrapper for mutliprocess-loading of dataset infos
+
+    :param all_infos: multiprocess-safe dictionary
+    :param d_name: dataset name
+    """
+    all_infos[d_name] = get_dataset_infos(d_name)
 
 
 # add an argument for read-only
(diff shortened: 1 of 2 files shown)


stephenbach:
Fixes #572
Fixes #588 
Fixes #416
Fixes #584

VictorSanh:
@stephenbach LGTM!

I have tested on MacOs Catalina with Python 3.7, 3.8 and 3.9, and everything looks good.
I also tested on Unix with python 3.7 for sanity check, same.

NB: I had to change L184 to `split_sizes = {k: v.num_examples for k, v in subset_infos.splits.items() if hasattr(v, "num_examples")}` because downloaded infos seemed to have change for the zaid/coqa_expanded and zaid/quac_expanded. We can also remove L186-L188's comments since they don't apply anymore.

Maybe keep the `if infos"` condition as sanity

setup.py
```diff
@@ -30,7 +30,7 @@
     url='https://github.com/bigscience-workshop/promptsource.git',
     author='BigScience - Prompt Engineering Working Group',
     author_email='sbach@cs.brown.edu,victor@huggingface.co',
-    python_requires='>=3.7, <3.8',
+    python_requires='>=3.7',
```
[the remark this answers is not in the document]
tianjianjiang: A python-version-related suggestion:

```python
requirements = [
    "streamlit==0.82",
    "datasets>=1.7.0",  # For this PR alone we need "<2" and yet we can just rebased onto `main` for #735.
    "pandas<1.4; python_version<'3.8'",  # Set pandas version for Python 3.7.
    "pandas>=1.4; python_version>='3.8'",
    "plotly",
    ##############################################################
    # Dependencies in this section are added for specific datasets
    ##############################################################
    "py7zr",
    ##############################################################
    # End of dataset-specific dependencies
    ##############################################################
    # TODO: We may consider moving this to a `dev` of `extras` and then changing GH Actions with `pip install .[dev]`.
    "black<=21.12b0",
    "flake8",
    "isort==5.8.0",
    "pytest",
]
```

Because `streamlit`, `datasets`, `plotly`, and `py7zr` will cover everything else, except that `pandas` needs a switch for Python 3.7.

tianjianjiang (approved):
I've tested using `"datasets==1.18.4"` without #735.
It works with `"python>=3.7,<3.10"` for macOS BigSur 11.6.5 on Intel, too.
I will also test Monterey and re-test Catalina soon, but they probably won't cause any issue.

promptsource/app.py
```diff
@@ -301,324 +145,496 @@ def get_infos(d_name):
...
+        #
+        # Metrics per dataset/subset
+        #
+        # Download dataset infos (multiprocessing download)
+        manager = Manager()
+        all_infos = manager.dict()
+        all_datasets = list(set([t[0] for t in template_collection.keys]))
+
+        pool = Pool(processes=multiprocessing.cpu_count())
+        pool.map(functools.partial(get_infos, all_infos), all_datasets)
```
tianjianjiang, line 166: Just a side note: IIRC, even after using `multiprocessing.set_start_method("spawn", force=True)`, a newer `streamlit` itself may still cause `fork` to happen. So #411 switches to threads. Also, multithreading is faster than multiprocessing in this case.

```suggestion
        pool = ThreadPool(processes=len(all_datasets))
        pool.starmap(get_infos, zip(repeat(all_infos), all_datasets))
```

stephenbach: Thanks, yeah, this was why I did something different from #411. I was still getting some errors occasionally with forking. If we want to upgrade streamlit, we could always force this pool to use spawning rather than set the global default.

Tagging in #725 .

Thanks for all your help with this!

PR #735, opened 2022-03-31
hopefully fixing load_module bug and compatible

jordiclive:
[732](https://github.com/bigscience-workshop/promptsource/pull/732), tested on datasets 1.17 and 2.0.
Hopefully using methods that are cross-compatible.

Ran into problems with `builder_instance = load_dataset_builder(dataset_path)`


VictorSanh (approved):
thank you @jordiclive! tested on my side, and it solves the issue!

PR #739, opened 2022-04-26
Add GEM/totto templates

manandey:


sebastianGehrmann:
So generally the ToTTo task is only the prediction of the describing sentence since some cells in the table are highlighted. Our data loader has a linearized_input field which can be used to represent the table. 

I am a bit worried grouping the URL and site name into the same task since they are considerably different. I would thus suggest to have consistent task here and only predict one of the tree. We could add additional templates in separate tasks that predict other properties.

manandey:
> I am a bit worried grouping the URL and site name into the same task since they are considerably different. I would thus suggest to have consistent task here and only predict one of the tree. We could add additional templates in separate tasks that predict other properties.

Thanks a lot for having a look! :) I have added this URL prediction template as a non-original task. For the original task, you have mentioned, the template I had written is in Line: 123. So, should I replace the URL template with a different one?

jzf2101:
@manandey I can't load the prompts when I use `promotsource viewer` when I try to load this commit.  

Could you try to call `streamlit run promptsource/app.py` and see if you can make changes to get the prompts to be viewed?



Screen Shot 2022-04-29 at 1 37 24 AM

manandey:
Hi @jzf2101, it would have been great if you could do a review of this PR sometime according to your convenience. I can start with the evaluation then. Thanks! :)

jzf2101:
@manandey @awebson I looked over the [original paper](https://arxiv.org/abs/2004.14373) again, it seems like the original task was:

> a controlled generation task: given a Wikipedia table and a set of highlighted table cells, produce a one-sentence description

In my naive opinion, I think we would need to render the actual table structure for these tasks to make reasonable prompts? E.g. 

Given the table below *question language of some kind*:

### Title

| col a    | col b |
| ----------- | ----------- |
| text a1      | text b1       |
| text a2   | text b2       |

I don't know how to do this in Jinja do either of you?

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,166 @@
...
+
+      {{sentence_annotations[0]["original_sentence"]}},
+
+      Guess the URL from which this table has been extracted!
+
+      |||
+
+      {{table_webpage_url}}
+
+      {% endif %}
```
jzf2101, line 47: This formatting does not seem naturalistic to me or representative of what the original task looked like- perhaps we should render the table?

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,166 @@
...
+      Table Page Title: {{table_page_title}},
+
+      {% if table_section_text != "" %}
+
+      Table Section Text: {{table_section_text}},
+
+      {% endif %},
+
+      and the following original text describing the highlighted cell of the table:
+
```
jzf2101, line 75: Similar concerns with formatting as opposed to representing the actual table

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,166 @@
...
+      {{sentence_annotations[0]["original_sentence"]}},
+
+      Can you guess the table''s section title?
```
jzf2101, line 78: Rereading the original paper I'm not sure correctly describes the original task? It appears that the original task was descriptions of table given the table as formatted?

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,166 @@
...
+
+      Table Page Title: {{table_page_title}},
+
+      Table Section Title: {{table_section_title}},
+
+      and the following original text describing the highlighted cell of the table:
+
+      {{sentence_annotations[0]["original_sentence"]}},
+
+      Can you guess the table''s section text?
```
jzf2101, line 110: Again I don't think this is what the target task was.

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,166 @@
...
+      and the following original text describing the highlighted cell of the table:
+
+      {{sentence_annotations[0]["original_sentence"]}},
+
+      Can you generate a final text taking into account the context from the details
+      given above?
+
+      |||
+
+      {{sentence_annotations[0]["final_sentence"]}}
```
jzf2101, line 153: This prompt reflects the original task best but again I think requires formatting more like the original paper

jzf2101 (changes requested):
Overall, I think we need to orient the prompts around the original task as closely as possible by presenting the table in it original formatting as opposed to text descriptions.  We also I think should orient the prompts around generating the final sentence, since this seems to be the main objective of the original task.

manandey:
Hi @jzf2101, a new PR #767 has been created addressing the comments in this PR. Feel free to close this PR once that is reviewed. Thanks!

shanyas10:
Hi @jzf2101 , we have tried to create 5 more prompts for the original task [here](https://github.com/bigscience-workshop/promptsource/pull/767). Can you please have a look? Also, there's one extra template (table_to_text) that's close to what you mention in [this comment](https://github.com/bigscience-workshop/promptsource/pull/739#issuecomment-1119696315) but does this look like an input that the model should be fed?

awebson:
@jzf2101 is right that in principal we should match the original task as much as possible, although @sebastianGehrmann in your paper, how is table-to-text formatted in the input sequence to your baseline models? My guess is not as a table drawn in ASCII art?

sebastianGehrmann:
The ToTTo data loader has a `linearized_input` field that has the recommended string representation from the paper.

PR #740, opened 2022-04-26
Adding GEM Simplification prompts. Includes challenge sets, and add…

jordiclive:
…itional prompts for Turk & ASSET.

May need to download dataset from [HF datasets](https://huggingface.co/datasets/GEM/wiki_auto_asset_turk/tree/main) manually.


sebastianGehrmann:
What's up with all the commented out things here?

jordiclive:
> What's up with all the commented out things here?

These were prompts that were based on those already in promptsource.  But not for the original task. e.g. "Make the text more verbose". Or "choose which is the simplified sentence A or B".

PR #745, opened 2022-04-26
Add GEM/xsum prompts

kkawamu1:
Add 10 prompts for GEM/xsum.
All the prompts are taken from the prompts already implemented for https://huggingface.co/datasets/xsum in the host version promptsource.

### DIFF
--- a/promptsource/templates.py
+++ b/promptsource/templates.py
@@ -26,7 +26,7 @@ env.globals.update(zip=zip)
 
 # These are users whose datasets should be included in the results returned by
 # filter_english_datasets (regardless of their metadata)
-INCLUDED_USERS = {"Zaid", "craffel", "GEM"}
+INCLUDED_USERS = {"Zaid", "craffel"}
 
 
 def highlight(input):
--- a/promptsource/templates.py
+++ b/promptsource/templates.py
@@ -26,7 +26,7 @@ env.globals.update(zip=zip)
 
 # These are users whose datasets should be included in the results returned by
 # filter_english_datasets (regardless of their metadata)
-INCLUDED_USERS = {"Zaid", "craffel"}
+INCLUDED_USERS = {"Zaid", "craffel", "GEM"}
 
 
 def highlight(input):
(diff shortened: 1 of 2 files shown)


jzf2101:
@awebson from an engineering perspective do we have to copy these directly or could we pull from  https://huggingface.co/datasets/xsum as is?

kkawamu1:
> @awebson from an engineering perspective do we have to copy these directly or could we pull from https://huggingface.co/datasets/xsum as is?

There is actually a slight difference in that https://huggingface.co/datasets/xsum  uses 'summary' as a key for the summary text,  while GEM/xsum uses 'target' as a key. So the prompts written for GEM/xsum reflect that difference. 

e.g. 

`{{document}}
      This boils down to the simple idea that ||| {{target}}`

instead of 

`{{document}}
      This boils down to the simple idea that ||| {{summary}}`

promptsource/templates.py
```diff
@@ -26,7 +26,7 @@
 
 # These are users whose datasets should be included in the results returned by
 # filter_english_datasets (regardless of their metadata)
-INCLUDED_USERS = {"Zaid", "craffel"}
+INCLUDED_USERS = {"Zaid", "craffel", "GEM"}
```
awebson, line 29:
```suggestion
INCLUDED_USERS = {"Zaid", "craffel"}
```

awebson: do not commit this hack

awebson (approved):
Yup what Ken did is correct. We need to copy the old xsum prompts to gem/xsum and make necessary changes. Merging now since the xsum prompts were already approved. Thanks!

promptsource/templates.py
awebson:
```suggestion
INCLUDED_USERS = {"Zaid", "craffel", "GEM"}
```

awebson:
@kkawamu1 Looks like taking out the hack broke automatic tests :( And I somehow I don't have permission to commit a fix from my end. Can you re-add `GEM` to `INCLUDED_USERS = {"Zaid", "craffel", "GEM"}` and push again?

kkawamu1:
@awebson  Should be updated! i.e. GEM is added to INCLUDED_USERS now!

PR #746, opened 2022-04-26
Add prompts for LinCE dataset Sentiment Analysis Task

RosenZhang:
Added 5 prompts to SA task from LinCE. The templates are close to ones of imdb dataset. The LinCE dataset is in Huggingface Datasets but it's unavailable from the promptsource interface. The `filter_english_datasets` method in utils.py is hence modified to add LinCE in the list before filtering. If there's more correct way to do so or other problems with the prompts, please comment below. Thanks!


awebson (changes requested):
Thanks Ruochen! 

1. Does this same prompt apply to all other subsets of LinCE? Or are we only asked to evaluate the `sa_spaeng` subset? 

2. Some prompts are missing the field of Answer Choices: positive ||| negative ||| neutral

3. This prompt's wording could be more natural
> The following post expresses what sentiment?

What sentiment does the following post express? Positive, negative, or neutral?

(in that case, you should also mark the "Choices in template" flag. That is, models are explicitly told the choices "Positive, negative, or neutral?" in the input.)

4. We're looking for at least 5 original task prompts. You're missing one.

RosenZhang:
> Thanks Ruochen!
> 
> 1. Does this same prompt apply to all other subsets of LinCE? Or are we only asked to evaluate the `sa_spaeng` subset?
> 2. Some prompts are missing the field of Answer Choices: positive ||| negative ||| neutral
> 3. This prompt's wording could be more natural
> 
> > The following post expresses what sentiment?
> 
> What sentiment does the following post express? Positive, negative, or neutral?
> 
> (in that case, you should also mark the "Choices in template" flag. That is, models are explicitly told the choices "Positive, negative, or neutral?" in the input.)
> 
> 4. We're looking for at least 5 original task prompts. You're missing one.

Hi Albert, thanks so much for reviewing this!
Re comments:
1. The templates in this PR would only apply to the `sa_spaeng` subset, other subsets are for different tasks, namely, language identification, POS, NER. I'm currently about to finish templates on NER tasks and probably will update the templates for it later today. I'm curious if we should develop prompts for all different tasks in this PR or go ahead to eval harness for testing the whole pipeline first? I think a bit more time is needed for developing prompts for language identification and POS tasks as there are less examples. (Welcome any pointers to similar tasks! Thanks! )
2. For Answer Choices, apologize for the inconsistency. I was wondering how the answer choices are passed into the model? In this dataset, the target is given in string format directly like ` 'sa':'positive'` and therefore referred directly as `sa` in the template, unlike other NER task where we may refer to the answer choices and write target as `choices[label]`. Would answer choices still be required in this case? 
3. Noted, I'll improve the wording of the prompts!
4. Also noted, I'll try to re-phrase and add an extra one! (If I understand correctly, negation wouldn't consider as original task right?)

RosenZhang:
Close this PR due to issues when rebasing to newly updated eval-hackathon branch.

PR #747, opened 2022-04-27
Support for multiple targets.

stephenbach:
THIS BREAKS BACKWARDS COMPATIBILITY WITH v0.2.x. Now, Template.apply now returns `Tuple[str, List[str]]` instead of `Tuple[str, str]`. The list of strings is the valid targets.

Users can add additional `|||` to their templates to denote additional valid completions.

Example of the interface:
Screen Shot 2022-04-26 at 8 37 27 PM


awebson:
Merging this now per discussion in our Monday (May 2) meeting. Thanks everyone!

PR #754, opened 2022-04-27
Adding prompts for POS tagging on Universal Dependencies dataset

aakanksha19:
Six prompts have been created to add part-of-speech tagging on the Universal Dependencies dataset to PromptSource. These prompts have been created from scratch since we could not find any POS tagging task references. We are using the following output format: the model must produce a sequence of word-tag pairs (e.g., the DET black ADJ sheep NOUN). Since computing model accuracy will require parsing out the POS tags from this string, we will need to implement a custom metric, and so the metric variable is currently set to "Other". 

In addition to these prompts, the universal dependencies dataset has also been added to Huggingface (under aakanksha/udpos) in a prompting-friendly format. To allow this dataset to be visible in promptsource, I have added my user name (aakanksha) to the list of included_users. If there is a better way to do this, please let me know and I can change this.

If there are any issues with the PR, please comment below and I will work on addressing them. Thanks a lot!

### DIFF
--- a/promptsource/templates.py
+++ b/promptsource/templates.py
@@ -26,7 +26,7 @@ env.globals.update(zip=zip)
 
 # These are users whose datasets should be included in the results returned by
 # filter_english_datasets (regardless of their metadata)
-INCLUDED_USERS = {"Zaid", "craffel"}
+INCLUDED_USERS = {"Zaid", "craffel", "aakanksha"}
 
 
 def highlight(input):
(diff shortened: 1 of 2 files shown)


awebson:
Thanks for the PR! Your prompts do not indicate no gold target. Please use `|||` in your jinja template to separate the input from the target. 

Additionally, we will discuss the use of non-natural language prompts in our Eastern Time standup meeting tomorrow (Thursday). I don't think non-expert humans can perform the task of generating every word and its POS tag as your prompts described. Please join our meeting and discuss if you can!

jzf2101:
@aakanksha19 - there are merge conflicts, could you please resolve?

PR #755, opened 2022-04-27
Allow nonenglish datasets.

stephenbach:
Relaxes dataset filter to no longer filter on datasets containing English data.


awebson:
Thank you so much! Will test this more extensively tomorrow, but merging now to allow people in other timezones to get things going.

PR #756, opened 2022-04-28
Add prompts for schema_guided_dstc8 response generation

kasnerz:
Added 5 prompts for chatbot response generation


stephenbach:
Thanks! Looks good! Just a couple of quick questions and thoughts:
- Should these be marked as "original task?"
- There are some new lines at the end of the targets that are added in some the prompt templates. These won't matter for methods that don't tokenize newlines separately, but probably best just to remove them. I assume you don't want to score methods based on that?

kasnerz:
- No, I do not think the prompts should be marked as "original task". The dataset was developed with dialogue state tracking for the [DSTC8 challenge](https://arxiv.org/pdf/2002.01359.pdf) in mind, which is quite different from the (stateless) response generation.
- The newlines are not intentional, it is probably just an artifact from the web interface. I tried to remove them in the YAML file, please check if I did it correctly. I also noticed the same pattern is present in other prompts (e.g. MetaWOZ which I was using as an inspiration).

kasnerz:
@stephenbach Done

promptsource/app.py
```diff
@@ -640,4 +640,4 @@ def show_text(t, width=WIDTH, with_markdown=False):
 
 
 if __name__ == "__main__":
-    run_app()
+    run_app()
```
stephenbach, line 643:
```suggestion
    run_app()

```

PR #757, opened 2022-04-28
Add LinCE sentiment analysis prompts

RosenZhang:
This new PR contains new prompt edits addressing previous PR comments:
> Does this same prompt apply to all other subsets of LinCE? Or are we only asked to evaluate the sa_spaeng subset?
> Some prompts are missing the field of Answer Choices: positive ||| negative ||| neutral
> This prompt's wording could be more natural
> The following post expresses what sentiment?
> 
> What sentiment does the following post express? Positive, negative, or neutral?
> 
> (in that case, you should also mark the "Choices in template" flag. That is, models are explicitly told the choices "Positive, negative, or neutral?" in the input.)
> 
> We're looking for at least 5 original task prompts. You're missing one.

Thanks a lot for the comments!

1. The templates in this PR would only apply to the sa_spaeng subset, other subsets are for linguistic structure tasks, namely, language identification, POS, NER. After standup, more work requires to be done on these tasks, More prompts for them are coming soon. I was thinking starting another PR for those since they're quite different from sentiment analysis task (requiring a label per word). 

2. For Answer Choices, apologize for the inconsistency. Would answer choices still be required if not referred in the target template?

3 & 4. The updated prompts in this PR improve wording and add an extra original task prompt.

### DIFF
--- a/promptsource/templates/lince/sa_spaeng/templates.yaml
+++ b/promptsource/templates/lince/sa_spaeng/templates.yaml
@@ -4,7 +4,7 @@ templates:
   24eba864-8859-4a15-a7a9-0fdf42d9f6cf: !Template
     answer_choices: positive ||| negative ||| neutral
     id: 24eba864-8859-4a15-a7a9-0fdf42d9f6cf
-    jinja: '{{words | join(" ") }} What is the sentiment expressed by the original
+    jinja: '{{words | join(" ") }}. What is the sentiment expressed by the original
       poster? ||| {{ sa }}'
     metadata: !TemplateMetadata
       choices_in_prompt: false
@@ -16,7 +16,7 @@ templates:
   29756194-c221-48b8-8d25-db1e681d9eeb: !Template
     answer_choices: positive ||| negative ||| neutral
     id: 29756194-c221-48b8-8d25-db1e681d9eeb
-    jinja: '{{words | join(" ") }} What sentiment is this post trying to express?
+    jinja: '{{words | join(" ") }}. What sentiment is this post trying to express?
       ||| {{ sa }}'
     metadata: !TemplateMetadata
       choices_in_prompt: false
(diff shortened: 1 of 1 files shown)


awebson (changes requested):
Thanks! Almost there!

> Would answer choices still be required if not referred in the target template?

Did you mean if choices are not mentioned in the *input* template? If so, yes. Choices should be given to models for all classification datasets. Models will score which answer choice is the most probable target sequence condition on the input sequence. Currently, the negation template is still missing choices.

> If I understand correctly, negation wouldn't consider as original task right?

Correct.

Can you put a period after the end of the main sentence for "express sentiment 2" and others? 

If you're interested, you can vary your wording even more. Currently they all basically say "sentiment" and "express". You can say for example "The author seems positive, neutral, or negative?", "The previous sentence has a good, bad, or neutral feeling?", etc.

RosenZhang:
> Thanks! Almost there!
> 
> > Would answer choices still be required if not referred in the target template?
> 
> Did you mean if choices are not mentioned in the _input_ template? If so, yes. Choices should be given to models for all classification datasets. Models will score which answer choice is the most probable target sequence condition on the input sequence. Currently, the negation template is still missing choices.
> 
> > If I understand correctly, negation wouldn't consider as original task right?
> 
> Correct.
> 
> Can you put a period after the end of the main sentence for "express sentiment 2" and others?
> 
> If you're interested, you can vary your wording even more. Currently they all basically say "sentiment" and "express". You can say for example "The author seems positive, neutral, or negative?", "The previous sentence has a good, bad, or neutral feeling?", etc.

Thanks again for the super helpful comments on writing prompts! The fixes are reflected in the new commit and add new prompts based on your suggestion!

awebson (approved):
Looks good now! Thanks so much!

promptsource/templates/lince/sa_spaeng/templates.yaml
```diff
@@ -0,0 +1,100 @@
...
+    jinja: "{{ words | join(\" \") }}. This is definitely \n||| \n{% if sa == \"negative\"\
+      \ %} \nnot a positive post. \n{% elif sa == \"positive\" %} \nnot a negative\
+      \ post. \n{% else %} \na neutral post.\n{% endif %}"
+    metadata: !TemplateMetadata
+      choices_in_prompt: false
+      metrics:
+      - Accuracy
+      original_task: false
+    name: negation template
+    reference: imdb
```
jon-tow, line 40: Hello, @RosenZhang ! The `answer_choices` for `negation template` do not align with the designed targets. E.g. the `positive` in `answer_choices` should actually be `not a negative`. Is there any chance we could get a fix for this? Thanks!

RosenZhang: Thanks for the catch, just open a new PR to fix this! Let me know if that works!

PR #761, opened 2022-05-04
adding templates for hi_shades

shanyas10:


stephenbach:
Hi! I can't read the prompts themselves, but a few questions:
- It seems like many of the prompts have targets that are 0 or 1. Should they be natural language responses?
- Please use `choice` instead of `random`.
- Is it problematic to have a prompt where the task is to add a stereotype to a passage?

jzf2101:
@shanyas10 there are merge conflicts could you please resolve?

jzf2101 (approved):

PR #766, opened 2022-05-07
add id prompt for tydiqa

afaji:
Indonesian prompt for Tydiqa_secondary.

Since the languages are not separated in tydiqa, I append the prompt with the original English prompt with an additional if-case.
Please let me know if this is an acceptable design. 

Failing check looks like because of non-English and non-Indonesian samples are not handled yet.


afaji:
@gentaiscool yes, the reason is that the prompt for tydiqa has an if-case where it only generates a prompt if the language is English. This PR adds Indonesian support as well. However, we have other unsupported languages, therefore producing empty prompt.

promptsource/templates/tydiqa/secondary_task/templates.yaml
```diff
@@ -98,6 +140,19 @@ templates:
 
       {{answers.text | choice}}
 
+      {% elif lang == "indonesian" %}
+
+      Aku lagi cari tau jawaban dari pertanyaan  "{{question}}"
```
gentaiscool, line 145: I found that we use both "Aku" and "Saya". I think we should use only one of them to make it consistent.

gentaiscool: I can help updating the yaml if you don't mind @afaji.

afaji: This is by design, since both Aku and Saya is naturally used by native speakers, so I try to cover all possibilities.

VictorSanh:
Hi,
I am unable to review indonesian prompts since I don't read/speak Indonesian myself... :/

However, would it be crazy to have the Indonesian prompts as different prompts than the english ones? i.e. instead of having an elif condition == Indonesian, just having the same thing under another prompt with just `if lg == indonesian`.

I ran the tests locally, they passed locally (your branch just does not seem up to date with the `eval-hackaton` branch). The reason why the tests fail is that there is a strange condition (https://github.com/bigscience-workshop/promptsource/blob/1b35aa0a749c72ee3ce5a22b6ef226a3c370e0d2/promptsource/templates.py#L194) that I need to double-check in regards to the blank results (like in tydiqa).

afaji:
Hi @VictorSanh, I tried updating my code with the current eval-hackathon branch but still have this blank result error.

I think this error does not exist in the main branch. That case did not exist there:
https://github.com/bigscience-workshop/promptsource/blob/main/promptsource/templates.py#L192

This is the reason for me adding `elif == indonesian`, since later if we support all languages, this error will go away. However, if we can ignore this error completely, I'm okay with your idea of another prompt with just `if == Indonesian`. I think this will be cleaner as well.

VictorSanh (approved):
@afaji 
I fixed things on the eval-hackathon main branch, and merge these changes back to your branch.
The tests are now passing.

Could you separate the prompts per language (i.e. having a different prompt for english and Indonesian)? it will be easier in particular once we introduce these language tags cf #771

gentaiscool:
@VictorSanh should we create a new file or new folder? do you have any suggestion?

VictorSanh:
> @VictorSanh should we create a new file or new folder? do you have any suggestion?

given promptsource structure, same file, just different prompts will work! (new folders or new files will not be correctly detected)

afaji:
Hi @VictorSanh 
Thanks for your suggestion. I've separated the prompt, but now got canceled on the show_template test. I believe this is because tydiqa is a huge template?

afaji:
I think #787 got the same issue

VictorSanh:
> I think #787 got the same issue

I increased the timeout on the test workflow (and merged bach the changes from eval-hackathon)
Will wait for the tests to run, and merge if they pass!

PR #767, opened 2022-05-07
Prompting totto

shanyas10:


promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,273 @@
...
+      metrics:
+      - BLEU
+      - ROUGE
+      original_task: false
+    name: guess the table section text
+    reference: ''
+  b25d135f-7c3a-4123-af40-2083a28a7add: !Template
+    answer_choices: null
+    id: b25d135f-7c3a-4123-af40-2083a28a7add
+    jinja: 'Convert the given linearized representation of a table to a coherent sentence:
```
jzf2101, line 218: This language is a bit technical- @awebson has suggested that we should avoid language that is too technical to describe a prompt.  Could we perhaps say "Describe the given table in one sentence"?

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,273 @@
...
+
+
+      {{linearized_input}}
```
jzf2101, line 221: For some reason I'm not getting the full table with this line

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,273 @@
...
+
+      Table Section Text: {{table_section_text}},
+
+      {% endif %}
+
+      and the following original text describing the highlighted cell of the table:
+
+      {{sentence_annotations[0]["original_sentence"]}},
+
+      Can you generate a final text taking into account the context from the details
```
jzf2101, line 255: I think `final_text` again is a technical term.  Could we replace "final text" with "one sentence description"?

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,273 @@
...
+      {% endif %}
+
+
+      '
+    metadata: !TemplateMetadata
+      choices_in_prompt: false
+      metrics:
+      - BLEU
+      - ROUGE
+      original_task: true
```
jzf2101, line 271: I still think the table should be rendered to be considered `original_task`

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,273 @@
...
+    jinja: '{% if target != "" %}
+
+      A table named "{{table_page_title}}" with section "{{table_section_title}}"
+      has the following highlighted cells:
+
+      {% for cell in highlighted_cells: %}
+
+      {% if table[0]|length >= cell[1] and table[0]|length >= cell[0] %}
+
+      {{table[0][cell[1]]["value"]}}: {{table[cell[0]][cell[1]]["value"]}}{% endif
```
jzf2101, line 159: Again, I think rendering the table would be necessary for this task to be considered an `original_task`

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,273 @@
...
+      original_task: true
+    name: best_describe_highlighted_sentence
+    reference: ''
+  4ef9e76e-7c03-4869-abdc-40ace4ce4c1c: !Template
+    answer_choices: null
+    id: 4ef9e76e-7c03-4869-abdc-40ace4ce4c1c
+    jinja: "{% if target != \"\" %}\nGiven the following information about a table\
+      \ and summarize the information from the highlighted cell:\nTable Page Title:\
+      \ {{table_page_title}} \nTable Section : {{table_section_title}}\"\nValues from\
+      \ Highlighted Cells: \n{% for cell in highlighted_cells: %}\n{% if table[0]|length\
```
jzf2101, line 99: This would still need the table.

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,273 @@
...
+
+      {% endif %}
+
+      '
+    metadata: !TemplateMetadata
+      choices_in_prompt: false
+      metrics:
+      - BLEU
+      - ROUGE
+      original_task: false
```
jzf2101, line 144: since we are currently not evaluating on `original_task: false`, for the sake of simplicity, could we exclude these non-original tasks?

jzf2101 (changes requested):
Only one of the tasks renders the table.  This would not be in line with the original task.  Could we try to render the table in all the tasks.  Also for simplicity, I'd like to remove all non-original tasks for now since they are currently out of scope.

jzf2101:
Please close #739 if you want this PR to replace it.

promptsource/templates/GEM/totto/templates.yaml
```diff
@@ -0,0 +1,273 @@
...
+    name: guess the table webpage url
+    reference: ''
+  484004fa-baf7-4757-a796-aba9f8c887a8: !Template
+    answer_choices: null
+    id: 484004fa-baf7-4757-a796-aba9f8c887a8
+    jinja: "{% if target != \"\" %}\nTable Page Title: {{table_page_title}} \nTable\
+      \ Section : {{table_section_title}}\n\nHow can you best use the above information\
+      \ and convert the  below highlighted cells of a table to a sentence?\n\nHighlighted\
+      \ Cells:\n{% for cell in highlighted_cells: %}\n{% if table[0]|length >= cell[1]\
```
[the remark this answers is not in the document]
shanyas10: Hi @jzf2101 the original actually consists of converting the table to text for the highlighted cells given the other informations like page title, section title etc. Which is why I believe the prompt here covers the information as the original task.

Can you please let me know the example id(s) for which you aren't able to render the highlighted cells? I'll check the scenario and modify the prompt if required

promptsource/templates/GEM/totto/templates.yaml
shanyas10: The linearized input only consists of table title, section title and highlighted cell information. Is that not being rendered?

shanyas10:
Hi @jzf2101 I have removed the prompts which don't contribute to the original task and made some minor changes to prompts to make them sound less technical. Regarding rendering of table, I've tried to create prompts in format other than `linearized_input` as well. Do you suggest that we should use `linearized_input` only to represent the table? However this again doesn't contain full table information (all rows and columns) but was used as input sequence to the baseline models as mentioned by Sebastian (author of the paper) in #739

stephenbach:
I still think there's an issue with the prompts based on these examples:
Screen Shot 2022-05-18 at 11 31 57 AM
Screen Shot 2022-05-18 at 11 32 31 AM

The prompt doesn't seem to provide the right info to generate that response.

I also don't think the `table_to_text` prompt is a useful way to do things. It took me a minute, but I think you're trying to indicate the highlighted cell by row and column number? I would not expect any current language model to be able to do that kind of counting. Perhaps there's a better way to do it?

shanyas10:
Thanks for reviewing the prompts @stephenbach . I have fixed the bug in the templates where the values were not rendering correctly. 
image
image


> I also don't think the `table_to_text` prompt is a useful way to do things. It took me a minute, but I think you're trying to indicate the highlighted cell by row and column number? I would not expect any current language model to be able to do that kind of counting. Perhaps there's a better way to do it?

Right! This was a temporary prompt that I had created to discuss the rendering of the table that @jzf2101 mentioned. I have removed that in the recent commit

stephenbach:
Sorry for the delay here. I think `best_describe_highlighted_sentence`, `final_text_describing_table`, and `summarize_the_highlighted_cell` are good. 

- I don't really get the `given_highlighted_cells_give_target_sentence` task. What is the goal? The target usually seems nearly identical to the "reference."
- `convert_linearized_input_to_sent` also seems a little off because the prompt suggests the task is to summarize the entire table, but it's really summarizing the highlighted cells.

PR #768, opened 2022-05-07
Add null prompt for `gsarti/flores_101` 

cjlovering:
[`gsarti/flores_101`](https://huggingface.co/datasets/gsarti/flores_101) is a dataset with 102 languages used for language modeling.
* Added the 102 language subsets
* The prompt is empty -- just the sentence itself -- as this dataset will be used for language modeling (LM).
* The metric is set to Other; downstream applications (eval harness) will select the LM metric.
* It was necessary to add `gsarti` to the user list because the dataset is listed under that user in hugging face datasets.

### DIFF
--- a/promptsource/templates.py
+++ b/promptsource/templates.py
@@ -26,7 +26,7 @@ env.globals.update(zip=zip)
 
 # These are users whose datasets should be included in the results returned by
 # filter_english_datasets (regardless of their metadata)
-INCLUDED_USERS = {"Zaid", "craffel", "GEM"}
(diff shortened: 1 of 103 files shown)


cjlovering:
Given that this is a null prompt and we're currently using it as as a language modeling dataset, we can skip this for now. Using `promptsource` isn't necessary.

stephenbach:
Can this PR be closed if we're not prompting it?

rbawden:
Hi there! For information, I have just prompted this dataset (automatically for MT). Linking the PR here: #779

PR #769, opened 2022-05-10
Prompts for crd3

shanyas10:


stephenbach:
Cool dataset! Looks great! Two minor things:
- Should `conversation_to_summary` be original task like the others?
- Some of the dialogs are really long, like 100+ turns, so they won't fit in the context window of most models. Not sure if there's anything you want to do to explicitly handle that, or at least put the instruction first in all cases so that it doesn't get truncated?

shanyas10:
Hi @stephenbach! Thanks for revie
> * Should `conversation_to_summary` be original task like the others?
Sorry about missing this. Changed this to original task


> * Some of the dialogs are really long, like 100+ turns, so they won't fit in the context window of most models. Not sure if there's anything you want to do to explicitly handle that, or at least put the instruction first in all cases so that it doesn't get truncated?
I have now put the all the instructions at the top to keep them from getting truncated.

stephenbach (approved):
LGTM!

PR #771, opened 2022-05-12
Language tags

stephenbach:
This PR adds a language tagging feature, so that users can annotate prompts with the language(s) used in the prompt.

Even though this is motivated by the eval hackathon, **this PR targets main** because it affects all prompts. All existing prompts in main are tagged with English. After merging into main, another PR should merge main into eval hackathon. This will require somewhat careful coordination because new prompts on that branch will need to have their metadata updated in the .yaml before they will work in the UI.

Regarding the tags themselves, the eval group requested using the subtags in [this list](https://en.wikipedia.org/wiki/IETF_language_tag#List_of_major_primary_language_subtags). I took the liberty of changing how the tags are displayed in the UI, appending the English names in parens, but this is changeable.

Screen Shot 2022-05-12 at 1 48 37 PM


awebson (approved):
Looks great to me! Thanks so much Steve!

I think we agreed to emphasize somewhere either in the UI or in the contribution guide that the language tag should be about the languages of each prompt, not the languages of the dataset examples (which should be already documented by the datasets themselves)?

VictorSanh (approved):
Looks good to me, thank you for adding this @stephenbach!

Regarding the language codes, the HF ecosystem (including datasets) is using ISO 639 for language codes (see https://huggingface.co/languages). Could we use the same here?
It will make running analysis on languages of prompts and datasets actually possible (or at least it will be smoother).

PR #772, opened 2022-05-13
Wikiann Prompts

jzf2101:


promptsource/templates/wikiann/en/templates.yaml
```diff
@@ -0,0 +1,189 @@
+dataset: wikiann
+subset: en
+templates:
+  41b109ce-7504-4263-b455-315fa7a7f3c2: !Template
+    answer_choices: Yes ||| No
+    id: 41b109ce-7504-4263-b455-315fa7a7f3c2
+    jinja: 'Given, "{{''  ''.join(tokens)}}", is {{spans[0].split('':'')[1]}} a location?
```
awebson, line 7: Why is the `''  ''.join(tokens)` here joining using multiple spaces?

awebson (changes requested):
Thanks Jess! NER is just hard to be adapted to a prompt form. Recall that we talked about this with @yongzx on Slack:

> Unless we are talking about entity classification (where we already know what is the entity and we want to figure out its type). NER on the other hand is entity extraction + entity classification.   

Your prompt gives the model the gold span as opposed to let models extract the span. So that's already non-original task. Further, several of your prompts ask for a binary classification, not multiple choice among location, person, organization. So those are also non-original task prompts. Right?

Lastly, are you sure the metrics are "AUC, Accuracy, COQA F1, and Other"?

awebson:
Also, the targets of your *_bool prompts should be, e.g., "False" as a single target sequence. Right now it's saying this example has "F", "a", "l", "s", "e" as five separate correct answers.

スクリーンショット 2022-05-29 20 34 22

promptsource/templates/wikiann/en/templates.yaml
```diff
@@ -0,0 +1,189 @@
...
+      - Accuracy
+      - COQA F1
+      - Other
+      original_task: true
+    name: is_organization
+    reference: ''
+  53bd1751-f0e6-4ca8-84d4-80c1c09a230d: !Template
+    answer_choices: True ||| False
+    id: 53bd1751-f0e6-4ca8-84d4-80c1c09a230d
+    jinja: 'Given the following information:
```
awebson, line 61: As opposed to "Given", or "Given the following information:", maybe it's more natural to say "in the following sentence…"?

yongzx:
I am working on this right now, but I ran into some weird GH issue (been Googling the whole day to debug it) where there's wikiann en templates in the folder, but streamlit doesn't load it.
Screen Shot 2022-06-01 at 3 42 26 PM
image

PR #775, opened 2022-05-20
Fix negation prompt answer choices in lince-sa task

RosenZhang:


RosenZhang:
Close this as solved by #780

PR #779, opened 2022-05-22
Flores 101 prompts

rbawden:
Automatically created prompts for MT using the Flores-101 dataset.

Contains prompts for all language directions using 31 BigScience languages (1 English prompt per language direction = 930 templates).

Question: is this format ok or should I separate out those that are for inclusion in the upcoming evaluation (i.e. only into and from English). This depends on how the notion of subtask is going defined and whether there is a possibility of selecting only certain templates.


rbawden:
@jzf2101

VictorSanh (approved):
looks good to me, thank you @rbawden !
(nice usage of anchors in yaml btw! :) )

I don't have a strong opinion on the format so will defer to the evaluation folks!

rbawden:
(Actually python seems to have dealt with that automatically since I created the templates automatically, so I won't take too much credit)!

PR #780, opened 2022-05-24
Fix `lince` negation template

jon-tow:
Fixes the issue in which `answer_choices` for the `negation template` were not aligned with the designed targets. E.g. the `positive` in `answer_choices` should actually be `not a negative post`.


jon-tow:
Bump. @jzf2101 Any chance I could get a quick skim of this? Thanks!

VictorSanh (approved):
thank you @jon-tow ! mergning now

PR #785, opened 2022-06-13
add id prompt for xquad

gentaiscool:
Indonesian translation for xquad
cc: @haileyschoelkopf @afaji


gentaiscool:
Just found that xquad does not support indonesian. I will close this PR.

PR #786, opened 2022-06-13
add Chinese wiki lingua on Wikilingua

gentaiscool:
Chinese translation for wikilingua
cc: @haileyschoelkopf @afaji


VictorSanh (approved):
thank you! looks good to me (I had a quick look at the Chinese though it's possible I made mistake since my Chinese reading is very rusty)
merging now

PR #788, opened 2022-06-25
Fix blank results

VictorSanh:
This is fixing is a bug in how we handle blank results (for instance `tydiqa` has a bunch of them).
+ standardize how we detect these blank results

Example: PR https://github.com/bigscience-workshop/promptsource/pull/766

### DIFF
--- a/promptsource/templates.py
+++ b/promptsource/templates.py
@@ -191,6 +191,10 @@ class Template(yaml.YAMLObject):
         # Splits on the separator, and then replaces back any occurrences of the
         # separator in the original example
         parts = [self._unescape_pipe(part).strip() for part in rendered_example.split("|||")]
+        if parts == [""]:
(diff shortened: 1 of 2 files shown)


VictorSanh:
In order to unblock other PRs, i am going to merge things first @stephenbach, but I am still eager to have your check!
We can add another fix if necessary

stephenbach:
Reviving this conversation to double check some stuff. The contributing guide says
```Filtering prompts. If a prompt is applied to an example and produces an empty string, that prompt/example pair will be skipped. You can therefore create prompts that only apply to a subset of the examples by wrapping them in Jinja if statements. ```
I could have sworn it was either a blank input/output/or totally blank, as supported by the way the tests are written. Did we change this at some point? Perhaps we can run a check on existing prompts or change the code further to allow all options?

PR #796, opened 2022-07-16
xcopa ID prompt translation

afaji:


haileyschoelkopf:
Thanks for translating these! @Muennighoff could we include these in XCOPA for mP3?

Muennighoff:
I think we want to do 2 datasets: 1) English prompts only 2) As many multilingual prompts as possible

So I think it's fine to merge this & i'll just make sure that we use the english prompts only for the first dataset iteration.

PR #798, opened 2022-07-17
Add xwinogrande

Muennighoff:
I just duplicated the files including the ID for fr & pt as I didn't run into any problems after that.
If it's not ok I can also change their IDs, but leave the duplicate prompts.

@haileyschoelkopf @thomasw21 @lintangsutawika


haileyschoelkopf:
LGTM! 

As long as the duplicated template IDs don't actually cause any bugs at runtime, beside the CI complaining on this PR.

Maybe @VictorSanh would know if duplicated template IDs will be a problem.

VictorSanh:
hmm locally i can't pass the test/test_templates.py test, but on the online CI, it does not complain... looking into that.
there is a test to prevent duplicate uuids

VictorSanh:
could you bring the latest changes from `eval-hackathon` into this branch and relaunch the tests?

PR #803, opened 2022-07-21
Tidy tydiqa

Muennighoff:


haileyschoelkopf:
Just saw this! Unfortunately, I’ll be away from my computer today and tomorrow.

Also just a note: we should not merge this into the 
`eval-hackathon` branch,

because the eval WG is using this dataset with the prompts + lang-specific if stmts in their experiments.

Muennighoff:
> Just saw this! Unfortunately, I’ll be away from my computer today and tomorrow.
> 
> Also just a note: we should not merge this into the `eval-hackathon` branch,
> 
> because the eval WG is using this dataset with the prompts + lang-specific if stmts in their experiments.

I think it's fine to merge this as it's a new dataset.
We can just leave the other tydiqa in the repo so there are both options.

PR #816, opened 2022-08-14
Xlsum

Muennighoff:
cc @lintangsutawika feel free to review / suggest diff prompts


lintangsutawika:
Yeah, might need more prompts where the prompt is at the beginning just incase samples are long and we end up truncating them.

Muennighoff:
> Yeah, might need more prompts where the prompt is at the beginning just incase samples are long and we end up truncating them.

Since our code currently skips too long samples, I just went with truncation directly in the templates 😁
Might actually be beneficial as it introduces some difference to the outputs, i.e. the articles for the different summarization prompts aren't always the same

Issue #840, opened 2022-11-15
Possibly incorrect prompt


xuanqing94:
https://github.com/bigscience-workshop/promptsource/blob/40ae287a96e1675ea07d390e64182df86b9e11ea/promptsource/templates/race/high/templates.yaml#L13

options[answer] is always a correct answer. I guess you mean options[candidate] (candidate is defined a few lines before)?

VictorSanh:
yes, good catch!
Want to open a PR to fix this?

PR #841, opened 2022-11-15
fix a prompt of RACE dataset

xuanqing94:
A fix for issue #840


VictorSanh (approved):
thank you @xuanqing94 !

PR #851, opened 2023-07-06
clarification on set up options

LadyHermitage:
In Setup direction, it was confusing to have primary direction for "If you don't intend to modify prompts" when the description of PromptSource is "a toolkit for creating, sharing...". To reduce confusion, I recommend "If you do not intend to create new prompts"


stephenbach:
Sure, I think that's helpful. Thanks!

PR #855, opened 2023-10-15
Apply black formatting.

shermansiu:
The current code fails to pass the style checks.


shermansiu:
@stephenbach @VictorSanh

stephenbach:
Thanks! I think we were skipping this check at some point because of other failures and never fixed the style.

